PR #1, opened 2016-10-13
Fix missing </complexType> tag on 'taskListener' definition.

roberthafner:


tijsrademakers:
Thanks for providing the first PR to the Flowable project Rob ;-)

PR #18, opened 2016-10-23
Add interface CandidateManager for getGroupsForCandidateUser

marcus-nl:
Most projects will have their own user/group/role management, so will not need the IdmIdentityManager. This PR greatly simplifies the interface to implement for getGroupsForCandidateUser support in such cases.


tijsrademakers:
Thanks

PR #20, opened 2016-10-24
slf4j-log4j12 and junit should be in test scope + fix indentation

marcus-nl:
flowable-idm-engine pom:
- maven dependency slf4j-log4j12 should be in test scope. this prevents users from inheriting it. junit should also be test scope.
- use consistent indentation.


jbarrez:
Thanks!

PR #23, opened 2016-10-24
Remaining optimizations for 'or queries' with multiple process variable values

m4ng0:
I addressed the same problem of Activiti/Activiti#988 (resolved with Activiti/Activiti#989) for Process Instance, Task and Historic Task queries.

I created a similar pull request for Activiti (Activiti/Activiti#1053), but this one is specific for flowable and can be merged directly.


jbarrez:
Thanks! This is indeed a nice optimisation!

PR #28, opened 2016-10-26
Update CallActivityBehavior to send VARIABLE_UPDATED events.

roberthafner:
This change makes CallActivityBehavior consistent with the variable event behavior for starting a process via the API.
- Modeled DataObjects - send VARIABLE_CREATED event
- Variables passed in via API call or CallActivity send a VARIABLE_UPDATED event if the variable was modeled as a dataobject otherwise a VARIABLE_CREATED event is sent.


roberthafner:
Tijs/Joram, Any idea when you'll get a chance to review this PR? Thanks!

tijsrademakers:
Hi Rob, sorry for the delay. We'll make sure to respond tomorrow.

PR #29, opened 2016-10-26
Incorrect event sent for external subprocess canceled by boundary…

roberthafner:
message event on callActivity.

Do not merge this pull request as it only includes a test to identify a bug and does not include a fix.

Tijs/Joram, We are seeing a PROCESS_COMPLETED event sent when an external subprocess is canceled via a message boundary event on a callActivity. My colleague and I spent sometime trying to correct the behavior but couldn't find a satisfactory fix.

Our initial thought was that line 152-153 of BoundaryEventActivityBehavior.java should pass in true for cancel parameter and line 456 in ExecutionEntityManagerImpl.java should inspect the cancel parameter to determine whether to send a PROCESS_COMPLETED event or a PROCESS_CANCELLED.  This change resolves the failure in the attached test case. However, it results in  test "testActivityTimeOutEventInCallActivity" in org.activiti.engine.test.api.event.ActivityEventsTest failing due to it receiving multiple ACTIVITY_CANCELLED events for the same users tasks. We did not find an easy way (maybe we are overlooking something) to prevent the multiple canceled events from being sent so we thought it was best to share our finding and hope you can provide some insight.

### DIFF
--- a/modules/flowable-engine/src/test/java/org/activiti/engine/test/api/event/CancelCallActivityTest.java
+++ b/modules/flowable-engine/src/test/java/org/activiti/engine/test/api/event/CancelCallActivityTest.java
@@ -1,0 +1,272 @@
+/* Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


roberthafner:
Opened a new PR which includes a potential fix for this issue.
https://github.com/flowable/flowable-engine/pull/30

PR #30, opened 2016-10-27
Incorrect event sent for external subprocess canceled by boundary mes…

roberthafner:
…sage event on callActivity.

Tijs/Joram, We are seeing a PROCESS_COMPLETED event sent when an external subprocess is canceled via a message boundary event on a callActivity. 

Our initial thought was that line 152-153 of BoundaryEventActivityBehavior.java should pass in true for cancel parameter and line 456 in ExecutionEntityManagerImpl.java should inspect the cancel parameter to determine whether to send a PROCESS_COMPLETED event or a PROCESS_CANCELLED. This change resolves the failure in the attached test case. However, it results in test "testActivityTimeOutEventInCallActivity" in org.activiti.engine.test.api.event.ActivityEventsTest failing due to it receiving multiple ACTIVITY_CANCELLED events for the same users tasks.  To resolve the duplicate TASK_CANCELLED event I added a new field (canceled) to the TASK_ENTITY class and set it to true when a TASK_CANCELLED event is sent and then check the value in ExecutionEntityManagerImpl to prevent sending a duplicate canceled event. The change results in the behavior we expect in our tests. However, I'm not sure if this is the optimal way to implement this change.

### DIFF
--- a/modules/flowable-engine/src/main/java/org/activiti/engine/impl/agenda/DestroyScopeOperation.java
+++ b/modules/flowable-engine/src/main/java/org/activiti/engine/impl/agenda/DestroyScopeOperation.java
@@ -60,7 +60,7 @@ public class DestroyScopeOperation extends AbstractOperation {
     TaskEntityManager taskEntityManager = commandContext.getTaskEntityManager();
     Collection<TaskEntity> tasksForExecution = taskEntityManager.findTasksByExecutionId(scopeExecution.getId());
     for (TaskEntity taskEntity : tasksForExecution) {
-      taskEntityManager.deleteTask(taskEntity, execution.getDeleteReason(), false, false);
+      taskEntityManager.deleteTask(taskEntity, execution.getDeleteReason(), false);
     }
 
     // Delete all scope jobs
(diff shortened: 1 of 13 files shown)


roberthafner:
Open a new PR which includes a potential fix for this issue.
https://github.com/flowable/flowable-engine/pull/30

tijsrademakers:
Thanks Rob, I've taken your PR changes and applied them manually with some changes. Maybe you can check if you think this commit is ok?

https://github.com/flowable/flowable-engine/commit/99cba7639c4c8b9a0c88ec8228ddae926c1228c6

roberthafner:
Hi Tijs,

The commit you merged looks fine and our tests are passing with the updated flowable jars. Thanks!

I noticed you left the increased timeout value in the waitForJobExecutorToProcessAllJobs. I assume that was oversight. 

.../flowable-engine/src/test/java/org/activiti/engine/test/api/event/ActivityEventsTest.java
-    waitForJobExecutorToProcessAllJobs(5000, 500);
 +    waitForJobExecutorToProcessAllJobs(50000000, 500);

tijsrademakers:
Hi Rob,

Oops thanks ;-) Used it for debugging.

PR #31, opened 2016-10-28
form engine REST API + admin app multiple endpoint type support

yvoswillens:
- added form engine REST API
- added form engine management info resource
- added multiple endpoint type support in admin app
 - separate endpoint config for process / dmn / form apis
- added DMN engine management info resource


yvoswillens:
implemented review remarks

PR #38, opened 2016-11-11
Fixed search variables by activityInstanceID in HistoricDetailQuery

dbmalkovsky:
This fixes a bug in the treatment of HistoricDetail (VariableUpdate type). It is occurs when variables stored in a UserTask with "processEngine.getTaskService().setVariable()" without completing the UserTask.

There is a unit test to validate the change.

### DIFF
--- a/modules/flowable-engine/src/main/java/org/activiti/engine/impl/persistence/entity/VariableScopeImpl.java
+++ b/modules/flowable-engine/src/main/java/org/activiti/engine/impl/persistence/entity/VariableScopeImpl.java
@@ -690,8 +690,16 @@ public abstract class VariableScopeImpl extends AbstractEntity implements Serial
         } else {
 
           VariableScopeImpl parent = getParentVariableScope();
+          //if (parent != null) {
+          //  parent.setVariable(variableName, value, sourceExecution, fetchAllVariables);
+          //  return;
+          //}
           if (parent != null) {
-            parent.setVariable(variableName, value, sourceExecution, fetchAllVariables);
+            if (sourceExecution == null) {
+               parent.setVariable(variableName, value, fetchAllVariables);
+            } else {
+               parent.setVariable(variableName, value, sourceExecution, fetchAllVariables);
+            }
             return;
           }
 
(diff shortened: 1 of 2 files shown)


tijsrademakers:
Hi David,

I've added the following unit test:

https://github.com/flowable/flowable-engine/commit/55b38134720c5caca5f5abf4bde5439414ea52c1

I think the change in VariableScopeImpl is not correct, because this creates 3 historic variable updates for the variable1 value1 update. What do you think?

dbmalkovsky:
I think you are correct; let's just close this pull request until the correct fix can be made.

Thank you for the investigation.

motorina0:
Hi,
  I've re-tested the fix and it creates exactly 2 Historic Variable Updates for _variable1_:
act_hi_detail_variableupdate

  Activiti version: **5.22.0**
  
  @tijsrademakers what version have you used for testing? I will try to reproduce the "3 historic variable updates" on it.

tijsrademakers:
Hi,

It was tested on the Flowable master version (which is the version 6 engine). With the change in this PR 4 historic variable instances are created.

motorina0:
Hi Tisj,
 -  I've re-tested with a build made from master (`commit 649c82de7982c46cd3a8db42d6c62509e6fce4a6
`) and it works fine (only 2 entries present, both have ActivityInstanceID)
 - the [added unit-test](https://github.com/flowable/flowable-engine/commit/55b38134720c5caca5f5abf4bde5439414ea52c1) is not correct since it does not query for ActivityInstanceID, therefore it always succeeds (the bug was about activityInstanceID)

**Actual Query**:
```
      List<HistoricDetail> resultSet = historyService.createHistoricDetailQuery().variableUpdates()
         .orderByTime().asc().list();
```
**Expected Query**:
```
List<HistoricDetail> resultSet = historyService.createHistoricDetailQuery().variableUpdates()
			.activityInstanceId(historicActivitiInstance.getId()).orderByTime().asc().list();
```
- I've re-checked the DB, the correct version seems to be deployed:
act_ge_property_

- is it possible to be other causes for the issue you've noticed (3 or 4 entries), like DB not dropped between Unit Tests, or mix-up with other fixes/features?

tijsrademakers:
Hi,

Did you apply this PR or did you apply your own PR? Did you also test the query without activity instance id to have a count of 2 as well? Because that's where it failed for me.

motorina0:
- I've made the changes directly (since it was very simple)
- Yes, I've checked also without activity instance ID. There is a total of 2 Activity instances. I've checked the ACT_HI_DETAIL table directly also.

tijsrademakers:
Ok thanks, I think I've made a mistake while checking the PR. I've merged the changes in a separate commit now. Thanks for checking this, closing the PR

PR #44, opened 2016-11-22
AsyncExecutor improvements

marcus-nl:
- only attempt to acquire new jobs if the queue has capacity.
- only attempt to acquire at most the number of jobs that fit in the queue.
- if there are still rejected jobs, unacquire them.
- option to unacquire all jobs by lock owner at startup.


marcus-nl:
Note that this corresponds to points 3 and 6 in https://forums.activiti.org/content/proposed-changes-job-handling-mechanism.

marcus-nl:
Any questions/remarks/issues before this PR can be merged?

tijsrademakers:
Hi Marcus,

Could you check why the number of files changed for this PR is 393?
It could be related to a force push we had to do a week ago and some rebasing might be necessary in your branch.
About the changes you proposed, I think they are fine. It just requires some additional testing on our side, maybe we have some questions then.

Thanks,

Tijs

marcus-nl:
The rebase is done.

PR #46, opened 2016-11-23
remove no use merge diff file

jasl:
just reading source code of flowable, found a `.orig` file seems no use which added at [commit 908125e](https://github.com/flowable/flowable-engine/commit/908125e412b083b82563c286b58fe596a315cc8c)


tijsrademakers:
Thanks

PR #51, opened 2016-12-03
Refactor deprecated usages

jasl:
- `ObjectUtils.toString(obj, nullStr)` to `Objects.toString(o, nullDefault)`
- `jsonNode.put(fieldName, jsonNode)` to `jsonNode.set(fieldName, jsonNode)`
-  typo
- removing some not existing param in javadocs

### DIFF
--- a/modules/flowable-bpmn-model/src/main/java/org/activiti/bpmn/model/Process.java
+++ b/modules/flowable-bpmn-model/src/main/java/org/activiti/bpmn/model/Process.java
@@ -112,10 +112,10 @@ public class Process extends BaseElement implements FlowElementsContainer, HasEx
   }
 
   /**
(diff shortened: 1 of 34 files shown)


jasl:
@tijsrademakers I found there are many different indent styles in files, including tab, 2 spaces and 4 spaces, does Flowable have a coding style? I may help to reformat

tijsrademakers:
@jasl thanks for the PR and offering the help. Right now we use 2 spaces, but indeed there are some parts of code where the code style is different. Before we release 6.0.0 we want to change the code style for every Java class to 4 spaces. But because this is a huge commit, it's probably easiest if we perform the change for all files at once.

jasl:
@tijsrademakers I see, you're right, reformatting codes after 6.0.0 released and before newer version development starting is the easiest way.

PR #57, opened 2016-12-06
Remove unnecessary 'null' check before 'instanceof' expression.

dbmalkovsky:
Since the instanceof operator always returns false for null, there is no need to have an additional null check.

Here is an example of the change:
   ` if (x != null && x instanceof String) { ... }`
Becomes:
    `if (x instanceof String) { ... }`

Also made a small change to one Javadoc instance that defined a parameter that did not exist.


dbmalkovsky:
The conflicts for this request have been resolved.

PR #58, opened 2016-12-06
JavaDoc improvements

dbmalkovsky:
The changes are only to JavaDoc and the changes made are the low hanging fruit; that is the easy changes without digging too deeply into the recent refactoring.


dbmalkovsky:
The conflicts for this PR have been fixed.

I will make another pass through the code sometime in the near future.

tijsrademakers:
Hi David,

See a couple of issues in the PR:

* @see {}https://activiti.atlassian.net/browse/ACT-1847} -> {} at the begin instead of {
* @throw InterruptedException -> should be @throws I think

Could you have a look?

dbmalkovsky:
Fixed those two occurrences and found a few more that were wrong in master.

PR #63, opened 2016-12-12
Use character for single character string

dbmalkovsky:
Replace string literals of length one being used in concatenation with equivalent character literals, gaining some performance enhancement.


tijsrademakers:
Hi David,

I would propose to keep the double quote strings like they are right now. The performance gain is minimal and the code doesn't improve with this change to my opinion. But open for other opinions of course.

Best regards,

Tijs

dbmalkovsky:
Other than the _small_ performance gain it does make it more consistent with existing code that does you the single character in the current code base.

But that is just one man's opinion....

tijsrademakers:
Closing PR for now, feel free to reopen when there are more opinions that this should be changed.

PR #72, opened 2016-12-13
Add missing whitespace in messages

dbmalkovsky:
Minor change to add missing whitespace to message text.


modules/flowable-spring/src/test/java/org/flowable/spring/test/email/MockEmailTransport.java
```diff
@@ -28,7 +28,7 @@ public void sendMessage(Message message, Address[] addresses) throws MessagingEx
     try {
       logger.info(message.getContent().toString());
     } catch (IOException ex) {
-      logger.error("Error occured while sending email" + ex);
+      logger.error("Error occured while sending email: " + ex);
```
PascalSchumacher: I guess `logger.error("Error occurred while sending email", ex);` would be even better. What do you think?

PR #77, opened 2016-12-14
Add support for rescheduling a timer via a service task.

roberthafner:
Here is the initial prototype support for rescheduling a boundary or intermediate timer using a message catching event that flows into the rescheduling service task. 

FYI, I'm not sure how to make this functionality work for the case of an intermediate timer in a multi-instance subprocess. For that case, I believe it will reschedule the timers (defined with the save intermediate catching event) across all of the subprocess instance while the client may want to be able to reschedule a timer in a single subprocess instance. Let me know if you have any thoughts on how this PR could be improved to support that. 

Thanks!
Rob


roberthafner:
I'm closing this PR while we continue to discuss the best way to support this type of functionality in the forums.  

Rob

PR #87, opened 2016-12-16
Duplicate condition in if stmt or condition

dbmalkovsky:
In one case ".map" was included twice and in the other "png".

Question: In the skipAuthenticationCheck() method should "'tif", "tiff", and "jpeg" be included also?


tijsrademakers:
@dbmalkovsky Yes I think they should also be included in the skipAuthenticationCheck

PR #103, opened 2016-12-21
Modifiers should be specified in the canonical preferred order

dbmalkovsky:
Addresses issue #3 from the [standards forum post](http://forum.flowable.org/t/coding-style-and-standards/154).


tijsrademakers:
Hi David, in the REST resources @ResponseBody seems to have been moved in the wrong place, or was this on purpose?

dbmalkovsky:
Just following the [Spring docs](http://docs.spring.io/spring/docs/current/spring-framework-reference/html/mvc.html#mvc-ann-responsebody) where they put in on the method itself.  I can undo this if you don't want it.

tijsrademakers:
Right, see it as well. In the Spring docs it's placed under the @RequestMapping notation instead of above, maybe we should follow that pattern as well?

dbmalkovsky:
Can be (and will be) done; it will have to wait until tomorrow; sorry.

dbmalkovsky:
Tijs,
I moved the location of the @ResponseBody annotations and in the process made the commenting of return code values more consistent.

tijsrademakers:
thanks

PR #110, opened 2016-12-27
Update prototype lib to version 1.7.x

ddpardo:
* Hash internals are now protected and only interactions trough function calls are allowed. (more immutable)
* Update is required to upgrade angulars version to 1.6.x
* General readability improvements in the editor code.
* Added 1.6, 1.7 version of prototype. (+ update helper)


tijsrademakers:
Thanks a lot for the patch, great work. I've been testing the changes with a lot of different models and it looks like it's working really well, except for pools / lanes and sub process scopes. You can't move an activity from one lane to the other, and this is possible without applying the patch. The same is true for a sub process, you can't move an element in a sub process to the main process for example. So I think there's different behaviour somewhere in that code with the upgrade to the new prototype library version.

A small thing is that in bpmn2.0.js line 414 a check is needed if lanes.first().properties["oryx-name"] is undefined. When you drag a pool to the canvas it fails on this line.

On a different note, do you think it would be possible to get rid of the prototype library fully? Or would this be a big effort?

ddpardo:
Hi Tijs,

I will look into the lane / subprocess problems. Thanks for testing this.

I don't think its possible to throw away prototype.js without altering something in the process. 
Here is my assumption list of what must be done: (big effort tough)...

- The editor (ORYX component) should not fetch data (stencil, model, ...) trough prototype synchronous (!) AJAX calls. This should be moved to angular. We might have to update angular for easier chaining of httprequests....  .then promise syntax vs .success / .error callbackhell syntax. 3 request must be performed on startup i guess... the model json, the stencil set and the plugins.xml . when all data is received the editor-canvas can be initialized.
- A replacement must be found for prototype new Hash() a.k.a $H its used almost everywhere where there is a state who needs to be tracked. We could just make our own implementation tough ....
- Prototype syntax / functions $$ , $A , $H , each, member, should be replaced 
- Cleanup allot of code.

ddpardo:
The subprocess, lane problems should be fixed now.

tijsrademakers:
Thanks David, works well now

PR #111, opened 2016-12-30
Update to angular 1.6.1

ddpardo:
* Removed angular 1.2.13 javascript files.
* Added 1.6.1 angular folder
* updated angular translate to 2.13.1
* Updated source code to fix problems.

 Important changes are:
 * Controllers can not be global variables : https://github.com/angular/angular.js/blob/master/CHANGELOG.md#breaking-changes-40
 * Sanitize strategy level for angular translate is now set on sanitize. see https://angular-translate.github.io/docs/#/guide/19_security
 * IE8 is not supported anymore: https://github.com/angular/angular.js/blob/master/CHANGELOG.md#breaking-changes-52
 * removed the soyc configuration flag of the gwt-maven-plugin it has no impact.


tijsrademakers:
Thanks David. I think the minified angular 1.6.1. files are missing from the PR.

ddpardo:
Hi Tijs,

Added the missing files. 

Angular mocks & scenario don't have a .min.js & .min.js.map.

Regards,

David

ddpardo:
Added a minor fix duo to prototype upgrade.

PR #112, opened 2017-01-02
fix a never touched if branch

jasl:
Both `Transaction` and `AdhocSubProcess` are `SubProcess`'s subclasses, so if `subProcess instanceof Transaction == false` condition is `true`, the last condition `subProcess instanceof AdhocSubProcess` will never be touched, this may a potential issue.


jasl:
Addition: is `if (condition)` & `if (!condition)` better than `if (condition == true)` & `if (condition == false)` ?

tijsrademakers:
Thanks, in the Flowable code we use if (condition) mostly I think and there's a mix of if (!condition) and if (condition == false)

PR #124, opened 2017-01-04
Update swagger doc with class name change

dbmalkovsky:
The title says it all.


yvoswillens:
Hi David.
These are files that are generated by the swagger maven plugin and should not be in the repository.
I'm working on a change in my feature branch. (adding additional swagger docs and some changes on generating the spec files)
I'll make sure that these changes will be applied there.

dbmalkovsky:
OK, sorry about that; we tend to hand code a lot of our swagger doc :)

I'll just close this PR then and let you handle it.

Issue #123, opened 2017-01-04
CountingTask (similar to the CountingExecution)


motorina0:
**Purpose**:
 - provide performance improvements by reducing the number of calls to the DB

**Counting Execution approach:**
 - has been implemented in this issue: https://github.com/Activiti/Activiti/pull/971
 - keep a count on executions to avoid deletes
 - an **ExecutionEntityImpl** holds references to lists of:

```
List<EventSubscriptionEntity>
List<JobEntity>
List<TimerJobEntity>
List<TaskEntity>
List<IdentityLinkEntity>
List<VariableInstanceEntity>
```
 - for each of these lists a count flag is kept separately (the **List.size()** can not be used since the lists are not populated in every scenario):
```
int eventSubscriptionCount
int jobCount
int timerJobCount
int taskCount
int identityLinkCount
int variableCount
```
 - two extra counts (no corresponding lists) are kept for
```
int suspendedJobCount
int deadLetterJobCount
```
 - these 8 counts are persisted to the ACT_RU_EXECUTION table (which now has 30 columns!)
 - the counts are increased/decreased in the **insert()/delete()** methods of their corresponding **XXXXEntityManagerImpl**
 - the counts are ONLY checked in the **ExecutionEntityManagerImpl.deleteDataForExecution()** method. Basically the delete is called for an entity only when we are sure that there is something to delete:
```
if (...((CountingExecutionEntity) executionEntity).getXXXXCount() > 0)) {
   //call delete
```
 - the "counting" performance improvement can be enabled/disabled by setting the **enableExecutionRelationshipCounts** property:
```
 	<bean id="processEngineConfiguration" class="...">
		<property name="enableExecutionRelationshipCounts" value="true" />
		...
```

**Counting Tasks approach:**
 - will follow the Counting Execution style
 - an **TaskEntityImpl** hold references to lists of:
```
List<IdentityLinkEntity> getIdentityLinks() 
List<VariableInstanceEntity> getQueryVariables()
```
 - **TaskEntityImpl** also has a Set of candidates (not sure how to be handled):
`public Set<IdentityLink> getCandidates()`
 - A **TaskEntityImpl** can have sub-tasks.  Let ExecutionEntityManager handle the sub-tasks count (together with normal task count)?

**Implementation**
 - add **CountingTaskEntity** interface (implemented by **TaskEntityImpl**)
 - add the two count fields (identityLinkCount and variableCount) to **TaskEntityImpl**
[...]

PR #137, opened 2017-01-10
CountingTask (similar to the CountingExecution) 

motorina0:
For more details please check #123

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/ProcessEngine.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/ProcessEngine.java
@@ -46,7 +46,7 @@ import org.flowable.idm.api.IdmIdentityService;
 public interface ProcessEngine {
 
   /** the version of the flowable library */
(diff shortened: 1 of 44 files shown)


dbmalkovsky:
I think that the db upgrade scripts need to be reviewed as my quick scan noticed that for db2 and hsql tha the same column is added twice.

<pre>
 +alter table ACT_RU_TASK add column VAR_COUNT_ integer;
 +alter table ACT_RU_TASK add column VAR_COUNT_ integer;
</pre>
 
I think that the second one should be `ID_LINK_COUNT_ `

motorina0:
Sorry for that.  I will review them immediately.

motorina0:
Done. Committed & pushed.
I will review the rest of the SQL scripts again.

dbmalkovsky:
Thanks for the quick turnaround.

A question for the group:  is this something that should be committed at this time (being in Release Candidate stage) as it includes modifying the database schemas or should it wait for the next development track?

tijsrademakers:
I think we can still include it for the 6.0 release as it's based on the execution count change. But it needs some testing before it can be merged, so will take a few days.

Issue #133, opened 2017-01-06
Inconsistent Handling of IdentityLinks for tasks


motorina0:
**Scenario**
  When a user is involved in a task an "IdentityLink" is created and it points to the TaskEntity

  Identity Links are not handled consistently dough:
   - "candidate" identity link types are persisted in "ACT_RU_IDENTITYLINK" and point to a task "ACT_RU_TASK"
   - "assignee" and "owner" identity link types 
      - are persisted in "ACT_RU_IDENTITYLINK" but have **TASK_ID** null
      - have a "PROC_INSTANCE_ID" value
      - have the type "participant" (in fact they refer to a process participant rather than a task owner or assignee)
      - the actual link between a **TaskEntity** and a User/Group is kept in the "ACT_RU_TASK" table in column OWNER_, ASSIGNEE_
      - the **TaskService.getIdentityLinksForTask(String)** method executes **GetIdentityLinksForTaskCmd** command which adds the two links manually and appends them two whatever it found in the DB

**Problems**
   - the API is not consistent. **TaskService.getIdentityLinksForTask(String)** and **IdentityLinkEntityManager.findIdentityLinksByTaskId(String)** can return different results (depending if the task has an owner or assignee)
    - it affects #123  by making it difficult to keep track (count) of task identity links
    - the Historic Identity Links are incomplete. The updates regarding task assignee and task ownership are not present in ACT_HI_IDENTITYLINK. Only the entries regarding process participation are kept.

motorina0:
This is a problem I have encounter while trying to implement #123 
The current approach might be desired way of handling the Identity Links (for performance problems).
I need confirmation/infirmation on this.

There is a comment in the **GetIdentityLinksForTaskCmd** class regarding the issue:
```
    // assignee is not part of identity links in the db.
    // so if there is one, we add it here.
    // @Tom: we discussed this long on skype and you agreed ;-)
    // an assignee *is* an identityLink, and so must it be reflected in the
    // API
    //
    // Note: we cant move this code to the TaskEntity (which would be
    // cleaner),
    // since the task.delete cascaded to all associated identityLinks
    // and of course this leads to exception while trying to delete a
    // non-existing identityLink
```

motorina0:
I did implemented the fix. Will make a pull request with the changes.
The project owners will decide if they will merge them or not.

tijsrademakers:
Hi,

[...]

PR #144, opened 2017-01-11
Create HistoricIdentityLinks when task assignee and task owner chage

motorina0:
It only adds `HistoricIdentityLinks` (that is for History Tables).

The runtime  identity links for **assignee** and **owner** are kept implicit in the ACT_RU_TASK table.

For more details see #133


tijsrademakers:
Thanks for the PR. I see one issue with this approach. When you would change the assignee for example 3 times there would 3 history identity links with type assignee in the database. But you would not be able to determine which one is the current assignee and in which order the assignees were changed. Maybe it's good to add a create date to the ACT_HI_IDENTITYLINK table?

motorina0:
Currently, the only way to tell the order of the changes is by the _ID column, but this is not a good idea since this is a technical field and it can not be used easily from the current API (to sort).

Adding a  TIME_ column to ACT_HI_IDENTITYLINK I also think is the best approach. All the other History table have either a TIME_, END_TIME_ or LAST_UPDATED_TIME_ column.

tijsrademakers:
Yep agreed, and the id can also be a string (StrongUUIDGenerator), so that's not generally usable. Could you add the CREATE_TIME_ column to this PR? I can also work on it, just let me know.

motorina0:
I will add it and get back to you.

motorina0:
- added CREATE_TIME_ logic
 - updated sql scripts
 - updated unit test
 - did rebase on master - no merge conflicts

PR #152, opened 2017-01-12
Fix bug in groups and users UI.

dpratt:
This PR fixes a problem in the IDM Admin UI that prevents adding users to groups due to the view resource for the popup not being resolved properly.

### DIFF
--- a/modules/flowable-ui-idm/flowable-ui-idm-app/src/main/webapp/scripts/common/directives.js
+++ b/modules/flowable-ui-idm/flowable-ui-idm-app/src/main/webapp/scripts/common/directives.js
@@ -412,7 +412,7 @@ flowableModule
         return {
             restrict: 'AEC',
             link: function(scope, element, attrs) {
-                element.on('hidden.bs.dropdown	', function () {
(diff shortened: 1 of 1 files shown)


tijsrademakers:
Hi, thanks for submitting the PR. I haven't been able to reproduce the issue yet. I'm able to add users to group with the existing code as well. Could you share how you encountered the issue?

dpratt:
Running flowable-ui-idm-app 6.0.0-RC1 on Safari 10.0.2 causes the following in the javascript console when you attempt to hit the '+ add user' button on the group editing page.
```
[Error] Failed to load resource: the server responded with a status of 404 (Not Found) http://localhost:8080/flowable-idm//views/common/popover/select-people-popover.html?1484258746816
```

This happens intermittently, and I'm not sure of the origins of that - if you don't see it, try doing a hard reload on the same page and see if it pops up for you. I do know that removing the spurious slash from the resource base URL makes it go away completely.

tijsrademakers:
Fixed in separate commit, thanks

PR #153, opened 2017-01-12
Fix for setting message references in modeler.

dpratt:
The modeler's client-side code isn't properly parsing the value that populates the selection dropdown. This commit fixes that.

Fix typo.


tijsrademakers:
Fixed with different PR (due to prototype fixes)

PR #156, opened 2017-01-13
Fix cut-n-paste error in recently added test

dbmalkovsky:
Reporting and fixing my own error (egg on my face) that resulting from cut-n-paste of the `assert` tests. The second set of checks should be on the new deserialized version and not the original variable.


dbmalkovsky:
The original PR #136

PR #204, opened 2017-02-06
ISO8601 Zulu time updates

stephraleigh:
Problem
Flowable persists the repeat timer string to the timer jobs table.   Datetime strings are either formatted in ISO8601 local time or non-ISO8601 string.

Fix
Datetime strings returned in the API (new the timer API) should be in ISO8601 Zulu time.  Zulu time should be used (not localtime) in case you have multiple clients running in different timezones.

Details
If the repeat is specified as R/<duration>, flowable determines the start time using the duration and persists the repeat field as R/’start datetime string’/<duration>
TimerUtil.prepareRepeat() is the method that formats the string.  It was formatting it in ISO8601 local time. It has been updated it to use ISO 8601 Zulu time.

If the repeat is specified with either a startDate or endDate datetime variable such as, R/${startDate}/<duration>  or R/<duration>/${endDate}, the JUEL expression is evaluated and Date.toString() is used to format the datetime string.  This is a NON-ISO8601 format.  I have added support for a custom type converter so that Date objects are converted into ISO8601 Zulu time.

Question
We did not add validation to the management service. Question is, should we require (validate) the repeat string only accepts datetime strings in Zulu time?  For example, a client could invoke the repeat() API and pass in a timeCycle that contains datetime string in their own locale, which happens to be different from the Flowable engine's locale.

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/el/ExpressionManager.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/el/ExpressionManager.java
@@ -13,6 +13,7 @@
 package org.flowable.engine.impl.el;
 
 import java.util.Map;
+import java.util.Properties;
 
 import javax.el.ArrayELResolver;
 import javax.el.BeanELResolver;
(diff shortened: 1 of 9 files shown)


tijsrademakers:
Thanks, adding validation to the ManagementService is not necessary for now I think.

PR #207, opened 2017-02-07
Reschedule should throw job rescheduled event

stephraleigh:
Reschedule call should throw JOB_CANCELED event since existing timer is terminated and new timer created.

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/delegate/event/FlowableEngineEventType.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/delegate/event/FlowableEngineEventType.java
@@ -91,6 +91,11 @@ public enum FlowableEngineEventType implements FlowableEventType {
   JOB_RETRIES_DECREMENTED,
 
   /**
+   * The job has been rescheduled.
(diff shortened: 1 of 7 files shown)


tijsrademakers:
I was thinking should we not add a JOB_RESCHEDULED event with the old timer id and the new timer id instead of using the JOB_CANCELED event. WDYT?

stephraleigh:
I will refactor my changes to add a JOB_RESCHEDULED event

stephraleigh:
I pushed a new change to create a new JOB_RESCHEDULED event.  Currently the reschedule causes a JOB_RESCHEDULED event to be sent. It contains the newly created timer as well as the job id of the original (rescheduled) job.    The reschedule call will also send a TIMER_SCHEDULED event for the newly created timer.  Do you think the TIMER_SCHEDULED is necessary given that the JOB_RESCHEDULED event contains the new timer entity along with the original timerId?

tijsrademakers:
Thanks, I think it's good to keep the TIMER_SCHEDULED event, if only for backward compatibility reasons.

PR #271, opened 2017-02-27
Update DefaultClockImpl to support specified timezone

stephraleigh:
Update DeafultClockImpl to support specified TimeZone.  If not specified, it will continue to use the JVM's TimeZone.  The CURRENT_TIME variable changed to protected to make it easier to extend if needed.


stephraleigh:
just checking to see if this will be something you will merge in?

tijsrademakers:
Yes thanks, was taking a bit longer due to some other merges to master

PR #282, opened 2017-03-02
Use JDK7 try/closeable and unwind the code

dbmalkovsky:
The same change in the Bpmn and Dmn converters to use the JDK7 try syntax (eliminating the finally block) and explicitly made it clear the opening and re-opening of the stream.


tijsrademakers:
Hi David, thanks for the PR. Could you change the log message that statically mentions UTF8 encoding? The encoding can be different than UTF-8. In the DmnXMLConverter the log lines now say BPMN 2.0 XML instead of DMN XML.

dbmalkovsky:
Sorry for the delay; is this better?

PR #290, opened 2017-03-06
Add email LDAP attribute

jpotts:
This change will allow the value of the email attribute in LDAP to be set on the user object.


tijsrademakers:
Thanks

PR #297, opened 2017-03-10
Fix in comment user and filter in task

Belarmino:
Fix coment user and change user filter


tijsrademakers:
Thanks for the PR. What does this actually fix? When would a user not have a lastname? Can you share more information about the reasons for this PR?

Belarmino:
this service (/flowable-task/app/rest/process-instances/1111/comments?latestFirst=true) dosen't give user object but user id so in comment in user-picture show ?? and add by nothing.

PR #315, opened 2017-03-29
(fix): Process instance variables deleted 2 times

gro-mar:


gro-mar:
subVar1 in the test is deleted twice

PR #336, opened 2017-04-07
Test case to demonstrate regression introduced by 320/330.

stephraleigh:
Process defined with terminateEnd event.  When flow transitions to terminate end, we see an unexpected PROCESS_CANCELLED event. This regression was introduced by 330.  We now see both a PROCESS_CANCELLED and a PROCESS_COMPLETED_WITH_TERMINATE_END_EVENT.   Only
PROCESS_COMPLETED_WITH_TERMINATE_END_EVENT is expected.
This is just a test case to identify the regression.

### DIFF
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/event/ProcessInstanceEventsTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/event/ProcessInstanceEventsTest.java
@@ -646,6 +646,88 @@ public class ProcessInstanceEventsTest extends PluggableFlowableTestCase {
         assertTrue(subProcessFound);
     }
 
+
+    @Deployment(resources = "org/flowable/engine/test/api/runtime/multipleSubprocessTerminateEnd.bpmn20.xml")
+    public void testProcessInstanceWithMultipleSubprocessAndTerminateEnd2() throws Exception {
+        ProcessInstance pi = runtimeService.startProcessInstanceByKey("multiplesubProcessWithTerminateEndTest");
+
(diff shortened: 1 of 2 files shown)


martin-grofcik:
PROCESS_COMPLETED_WITH_TERMINATE_END_EVENT does not contain the cause of the termination.  That's the reason why I have proposed to fire PROCESS_CANCELLED (which contains cause of the cancellation) ehavevent too. Flowable 5.22 fires PROCESS_CANCELLED event on the terminate end event too. 
Another possibility is to add termination cause to PROCESS_COMPLETED_WITH_TERMINATE_END_EVENT and remove PROCESS_CANCELLED events.

roberthafner:
Tijs and I discussed the fact that Flowable was sending a PROCESS_CANCELLED event when a process flow reached a terminate end event and decided it should be changed to PROCESS_COMPLETED_WITH_TERMINATE_END_EVENT in version 6. See the following discussion for reference:

https://forum.flowable.org/t/why-is-process-cancelled-event-sent-from-a-terminate-end-event-instead-of-process-completed/87/3

Therefore I think adding the termination cause to the PROCESS_COMPLETED_WITH_TERMINATE_END_EVENT is probably the direction we'd prefer to go in. 

Tijs, can you confirm? 

thanks,
Rob

tijsrademakers:
I agree that the cause should be added to the PROCESS_COMPLETED_WITH_TERMINATE_END_EVENT event and that we can remove the PROCESS_CANCELLED events for these cases. @martin-grofcik do you want to implement this change or should I have a look at it?

martin-grofcik:
I can implement the change.

tijsrademakers:
Thanks Martin

PR #346, opened 2017-04-12
applying ExecuteAsyncRunnableFactory in default async executor

martin-grofcik:
I did not find a place where executeAsyncRunnableFactory is used.


tijsrademakers:
Merged manually

PR #367, opened 2017-04-28
making databaseTypeMappings static again

martin-grofcik:


tijsrademakers:
Hi Martin,

We ran into issues in the Flowable QA environment where we are building and shutting down multiple engine, so we don't want to make the make static again.

PR #370, opened 2017-05-02
Add new property maximum and minimum length

robsoncardosoti:
Add new property maximum and minimum length in the Advanced tab of the fields text, multi-line-text, integer and decimal


tijsrademakers:
Thanks for the PR, the maxLength feature is working nicely. But I don't see the minLength feature doing anything for text fields. I think it should be possible to complete a form when the minLength is not met, or do you have a different view?

robsoncardosoti:
You are right! I will try to insert a visual treatment in the border / contour of the field that was not informed the minimum size/length.
After creating the PR I have identified that the creation of a new parameter in FormField.java is not necessary because the params can be used.
Do you prefer to cancel this PR and then reopen with the changes finalized?

tijsrademakers:
In addition to the visual treatment do you want to disable to complete button as well? Otherwise you will still be able to complete the form, also if the minimum length condition has not been met.
Maybe it's best to cancel this PR and then reopen it when ready. Thanks!

robsoncardosoti:
I've already implemented the visual treatment and disabling the complete button for the required fields. Non-mandatory fields will only visually highlight when the minimum size is not reported.
Is this behavior acceptable?

tijsrademakers:
Yes that's certainly acceptable, thanks

robsoncardosoti:
Sorry,
I ended up creating two more PRs to implement the visual treatment. I'm still not very skilled with Pull Requests.

PR #373, opened 2017-05-07
Rest Activity Feature

h7kanna:
Hello team,

I finished an initial implementation of HTTP Task.

**Features**
1. Http Task in flowable-modeler 
2. Http Task specification and documentation ( feedback and comments would be beneficial for final version)
3. Http Client configuration in flowable.cfg.xml ( Only few parameters for now)
4.  Implementation based on Apache HTTP client and this can be customized. Not all the settings for the client are implemented.
5.  Test cases for basic exception handling and retry scenarios

**Todo list** 

1. Implement and write tests for all http methods
2. Implement and write tests for all http client configuration parameters
3. Implement and write tests for  custom https certificates configuration
4. Implement and write tests for authentication
5. Implement and write tests for response parsing/handling
6. Performance testing and optimizations, code cleanup/refactor
7. Any other features/changes based on review and comments

Submitting pull request to get feedback and reviews. 

Flowable forum discussion
[Rest Activity Feature](https://forum.flowable.org/t/rest-activity-feature/380)

Thanks,
Harsha Teja Kanna


tijsrademakers:
Hi Harsha,

Thanks for all your work on this.
Looks like a very nice addition to the Flowable project.
A couple of comments:

1. I think we don't need to set the field values to the execution as variables. The field extension values can always be retrieved when necessary. Or do you have a specific reason for doing this?

2. You can remove the following code from the static initialization part of the HttpActivityBehavior(Impl) class I think:

// TODO: Assuming Context is available in static scope
 +        HttpClientConfig config = Context.getProcessEngineConfiguration().getHttpClientConfig();
 +        if (config == null) {
 +            config = new HttpClientConfig();
 +        }

h7kanna:
Hi @tijsrademakers ,
1) I added the field values to variables, so that we can listen for them in event handlers, execution listeners. I can make it optional , with a flag.

2) So config is available in static block right, there is no need to get it from beans?

Thanks,
Harsha

tijsrademakers:
Hi @h7kanna,

1. You can still get the values in execution listeners by getting the expression value there. So I don't think these variables are necessary.

2. What I meant was, is it really necessary to build the defaultRequestConfig in the static block? You also set the config again when executing the http activity behavior, so isn't that enough?

h7kanna:
Hi tijs,
Got you. Will make the necessary changes. Thanks.

Do you have any other recommendations?

I'm planning to add more tests, but work on the features later on.
What is the timeline for 6.1.0 release, can we get this into it?

Also, I did not handle any compatibility in HttpActivityBehavior. So this would be only for 6.1.X series.

Thanks,
Harsha Teja Kanna

tijsrademakers:
More tests would be very useful indeed. The 6.1.0 release planning is about a month from now.
No other recommendations at this point.

tijsrademakers:
Hi Harsha,

Just wanted to ask if implementing the tests is progressing well. When do you expect this to be ready so we can merge the PR?

Thanks!

h7kanna:
HI @tijsrademakers ,
I am making progress on this.
Will make final commits along with tests in a week.
Does this work?

Thanks,
Harsha

tijsrademakers:
Hi Harsha,

Yes that works well!

Thanks,

Tijs

h7kanna:
Hi @tijsrademakers , 
Progressing at slow pace.
Will keep you posted by EOW.

Thanks,
Harsha

h7kanna:
Hi @tijsrademakers ,
Added more tests.
Please review.

Thanks,
Harsha

h7kanna:
Hi @tijsrademakers , 
Apache http client version is 4.3.5 . Can we update to 4.5.X series?
I see it being used in flowable-mule and flowable-ui-* modules. So using the same version. Thought would result in class loading issues with multiple versions.

Thanks,
Harsha

tijsrademakers:
Hi @h7kanna ,

Yes I think we can update to Apache Http Client 4.5.x. Will do that change today.

Regards,

Tijs

tijsrademakers:
Hi @h7kanna,

The Apache HttpClient library is upgraded now.
The PR is looking great, some last comments:

- In HttpActivityBehaviorImpl can you change the static initialization code to the constructor?
- In addition the Timer instance should not be a static class variable, because this will fail when multiple http tasks are executed by different threads
- Could you add a storeResponseParameters option to the http task configuration? The response body itself always needs to be stored of course, but the other values like the response code etc, might not always be needed

We can also help with doing these changes, so just let us know.

Thanks,

Tijs

h7kanna:
Hi @tijsrademakers ,
I can make those changes quickly and update you.

Thanks,
Harsha

tijsrademakers:
Hi @h7kanna ,

Thanks!

h7kanna:
Hi @tijsrademakers ,
Made required changes, Please review.

Thanks,
Harsha

tijsrademakers:
Thanks for the changes, will do some testing with it and let you know the results

h7kanna:
Hi, @tijsrademakers ,

Also, I will keep adding Http Client Configuration parameters when I get time. But may not be in this pull request.

Also thinking how to add generic Auth Handler, Response Handler to make the task more complete. Please give any suggestions.

Are you planning to merge this in 6.1.0?
The task do not have compatibility handling with 5.X flowable version. Please verify.

Thanks,
Harsha

tijsrademakers:
@h7kanna, thanks a lot for this great feature! I've added a couple of config options (response variable name, http request and response handler) as well. Need to add these options to the Modeler still, will try to do that this week.

h7kanna:
I can help add those in modeler if needed.
Thanks @tijsrademakers

PascalSchumacher:
Great feature. Thank you very much for adding this!

PR #375, opened 2017-05-07
Remove some duplicate HTML element attributes.

chrisgavin:
I spotted these duplicate HTML element attributes (the attribute values are the same in all cases) so I thought I would remove the redundant ones.

---

I found these issues through [lgtm.com](https://lgtm.com/projects/g/flowable/flowable-engine/alerts/), a service that analyzes open-source code to look for potential problems. (Full disclosure: I work for the company that operates it.)


tijsrademakers:
Thanks

PR #376, opened 2017-05-08
Fixed upload form-field to allow readOnly

paulstapleton:
Fixed upload form-field to allow the same field ID to be used in multiple forms and when used as a readOnly field.
This fixes how the content-id(s) of the attached files is handled in flowable-task app.

See the discussion at https://forum.flowable.org/t/error-when-you-reuse-the-form-field-of-the-attachment-upload-type/522


tijsrademakers:
Thanks

PR #380, opened 2017-05-12
Added ability to save forms in flowable-task and support for multiple…

paulstapleton:
… form instances on the same task

See in the forum - https://forum.flowable.org/t/ability-to-save-without-completing-forms-in-flowable-task/538/2

I have added the ability to save forms using the flowable-task app. I have made a number of comments in the commit as well to explain my decisions.

This has required a few changes in the form Engine. In particular I have changed so that its possible to save multiple formInstances for a task and that the latest submitted is treated as the last saved for ongoing tasks or completed values for completed tasks. Previously this caused an error when creating the formInstanceModel for a task.

I have not looked in depth at the form rest API but I Believe it should work with my solution.

As far as I can see none of the classes I have modified contain unit tests and I haven't had time to add these.
I suspect there will be improvements needed so please let me know your comments/feedback.

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/cmd/CompleteTaskWithFormCmd.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/cmd/CompleteTaskWithFormCmd.java
@@ -79,7 +79,7 @@ public class CompleteTaskWithFormCmd extends NeedsActiveTaskCmd<Void> {
             FormService formService = processEngineConfiguration.getFormEngineFormService();
             Map<String, Object> formVariables = formService.getVariablesFromFormSubmission(formModel, variables, outcome);
 
-            formService.createFormInstance(formVariables, formModel, task.getId(), task.getProcessInstanceId());
+            formService.saveFormInstance(formVariables, formModel, task.getId(), task.getProcessInstanceId());
 
             processUploadFieldsIfNeeded(formModel, task, commandContext);
 
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/cmd/GetTaskFormModelCmd.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/cmd/GetTaskFormModelCmd.java
@@ -87,7 +87,7 @@ public class GetTaskFormModelCmd implements Command<FormModel>, Serializable {
         } else {
             formModel = processEngineConfiguration.getFormEngineFormService()
                     .getFormModelWithVariablesByKeyAndParentDeploymentId(task.getFormKey(), parentDeploymentId,
-                            task.getProcessInstanceId(), variables, task.getTenantId());
+                            task.getProcessInstanceId(), taskId, variables, task.getTenantId());
         }
 
         // If form does not exists, we don't want to leak out this info to just anyone
--- a/modules/flowable-form-api/src/main/java/org/flowable/form/api/FormService.java
+++ b/modules/flowable-form-api/src/main/java/org/flowable/form/api/FormService.java
@@ -48,20 +48,39 @@ public interface FormService {
      */
     FormInstance createFormInstance(Map<String, Object> values, FormModel formModel, String taskId, String processInstanceId);
 
+    FormInstance saveFormInstance(Map<String, Object> values, FormModel formModel, String taskId, String processInstanceId);
+
+    FormInstance saveFormInstanceByFormModelId(Map<String, Object> variables, String formModelId, String taskId, String processInstanceId);
+
     FormModel getFormModelWithVariablesById(String formDefinitionId, String processInstanceId, Map<String, Object> variables);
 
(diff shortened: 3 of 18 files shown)


tijsrademakers:
Thanks for the PR Paul. Can you share the thoughts of removing the fillPreviousFormValues method? That method was used to fill the map with form field values that were submitted in earlier user task or start forms of the process instance. I'm also thinking if we shouldn't put the saved form instances in a different table. Or add a column to the form instance table to mark a row as a saved form instance type.

paulstapleton:
Hi,

I did make a comment about this. But I'll explain my reasoning again. 

I'm not really sure how the method `fillPreviousFormValues` was supposed to work, and I Think the logic is a bit flawed.
For starters I don't Think this function has ever actually worked as expected. There is the code here:

```
if (submittedNode == null || submittedNode.get("values") != null) {  
    continue;  
}  

JsonNode valuesNode = submittedNode.get("values");  
Iterator<String> fieldIdIterator = valuesNode.fieldNames(); 
```

The condition would almost always evaluate to true and if not we would get a `null` value for `valuesNode` causing a NullPointerException. So I Believe this method has Always returned an empty Map.

However even with this fix I still think the behaviour is strange, The form field value is looked up in all existing form instances in the same process instance and this overrides any values set using process variables.
This would be a problem if form A sets a field value, then a serviceTask Changes the value and finally form B uses the same value. This will display the value from form A and not the serviceTask. I suspect this is not what is wanted. Shouldn't we not use the process variable instead of previous usertask forms?

So because of this I Believe this function is not needed. But please correct me if I have missed something.


As far as the database goes, I made a design decision to reuse the formInstance table for saved forms, and that the latest submitted entry is the latest save for uncompleted usertasks or the submitted form for completed tasks.
I did not feel an extra column or extra table was needed as the same information exists by examining whether the task is completed or not.

I feel the design I chose has a low impact on other APIs and really the only change I have made is to allow multiple formInstances for a task, and that the latest formInstance is used when populating the FormModel/FormInstanceModel.

There is also a question of whether we want to store all saved forms or not. An alternative design could be to only have a single formInstance which is created on first save and constantly updated when saved or the form is completed.

A risk with my solution the formInstance table will become quite large and slow as the form-Engine is used. This could impact the response time from the Engine. If this is a concern possibly only storing one formInstance and updating it or a separate saved formInstance table could be a better design.

I'm open for making further Changes to the PR if you feel I made the wrong design decisions. Please let me know what you Think.

Kind regards,
Paul

tijsrademakers:
Hi Paul,

Thanks for the info and I fully agree with you, this has never worked correctly. I think the process variables already cover the case of values filled-in in a previous form so that should also work without this logic anyway.

For the design I would propose to use a single form instance. There's no real reason to store every saved form instance (or we should store that in a history table). Would it be ok for you to change the PR to a single form instance?

Thanks

paulstapleton:
Hi Tijs,

I will make the changes to a single formInstance as you suggest.

Just a question about how you generally like to work with PRs. 
Should I replace the existing commit with a new updated commit? or should I add an extra commit to the PR with the modifications?
I am happy to do it either way.

Regards,
Paul

tijsrademakers:
Hi Paul,

Thanks! Just choose what's the most convenient for you. We always look at the diff to master and if the changed files look ok that's the important part.

Regards,

Tijs

tijsrademakers:
Hi Paul,

Just wanted to check on the status of this. We are approaching the 6.1.0 release and maybe we can still add this to the release as well. Also if you want us to have a shot at the implementation, just let us know.

Thanks,

Tijs

paulstapleton:
Hi Tijs,

Sorry its been a bit quiet here. I have been slowly working on this and hopefully can have a new PR soon. As a new release is upcoming I'll try to get it done today.

Regards,
Paul

paulstapleton:
I have now rebased and updated my commit.

There is now only a single FormInstance and I added a method to Save (upsert) a FormInstance. This is used when both saving the form and completing a task with a form. I did not change the behvaiour of the existing create FormInstance method to maintain backwards compatibility.

I had a few problems when trying to update the FormInstances submitted values (JSON saved as byte Array). When I updated the bytes in ResourceRef.java using `ResourceRef#setBytes(byte[] bytes)` this did not automatically update the value in the database. I added an extra call to `Context.getCommandContext().getResourceEntityManager().update(entity)` to perform this update. I am not sure if this is the correct solution or if something else is wrong. In `org.flowable.engine.impl.persistence.entity.ByteArrayRef` there is no explict update either - but I have not tested that updating a ByteArrayRef works as expected or is also broken.

I also fixed a typo in the SQL Query name `selectFormInstancesCountByQueryCriteria` where it was sometimes written as `selectFormInstanceCountByQueryCriteria`

Please let me know if any further Changes are needed and I can make them.

Regards,
Paul

tijsrademakers:
Thanks!

wjansenw:
Hi,
I noticed the saving of forms does not seem to work for checkboxes or for dropdown menus. Maybe also for other form items.
regards,
Wim

PR #385, opened 2017-05-17
updated distro ant to include swagger spec files

yvoswillens:
fixed issue not using swagger profile for REST APP WAR build


yvoswillens:
will incorporate this in other PR

PR #393, opened 2017-05-19
Add New Form Field Hyperlink

robsoncardosoti:
Create new form field Hyperlink in Flowable Modeler.


robsoncardosoti:
Tijs,

The new Hyperlink form field is working correctly in Flowable Modeler and Task, but in Admin it is not appearing in Process Engine> Instances> Variables in Column Type and Value. What can it be? Which file do you have to adjust to display in Flowable Admin the value of the new field created?

tijsrademakers:
Hi,
The hyperlink form field is not transformed into a process variable when submitting the form. I think that's also the desired behavior, because the hyperlink is a readonly form value.

PR #394, opened 2017-05-20
Update en.json

robsoncardosoti:
Correction of the translation of the expression Maximum length and Minimum length


tijsrademakers:
Changed by PR #393 already

PR #395, opened 2017-05-23
Add fluent API to create BPMN files for testing

dbmalkovsky:
Add a new module, `flowable-bpmn-fluent` that provides a mechanism for building models in code using a Fluent API.  

See `modules/flowable-bpmn-fluent/FluentApi.adoc` for a description of the API.

See `modules/flowable-bpmn-engine/modules/flowable/engine/src/test/java/org/flowable/engine/test/api/v6/Flowable6Test.java` where a few of the tests were re-written to use the new API.  Instead of replacing the old tests the new version is directly below the original.

This implementation is *not* a complete implementation of the BPMN supported by Flowable nor does it include layout (DI) for the models.

The intended use case is to provide an easy way to generate BPMN code for testing purposes.

This code is a derivative work of a similar implementation from another Apache-2 licensed workflow project.


dbmalkovsky:
Closing this pull request so it can be reworked per offline comments.

PR #396, opened 2017-05-23
Add input mask in form fields

robsoncardosoti:
Add input mask and Regex Pattern in form fields text, multi-line-text, decimal and integer.


robsoncardosoti:
I am not allowed to create folders in the project, so the mask.min.js file has been added to the ../modules/flowable-ui-task/flowable-ui-task-app/src/main/webapp/libs folder. But it might be interesting to create the folder angular-ui-mask inside the ../modules/flowable-ui-task/flowable-ui-task-app/src/main/webapp/libs and add the file mask.min.js in this pasta.

PR #405, opened 2017-05-31
Using concatenation to build SQL

carlosalbertomjr:
The search of full name is a like on the first name or on the last name. That causes no results with typing full name at the search in people field form and others.

Ex.: 
First Name: Carlos Alberto
Last Name: Moraes Junior

Search:
"Carlos Alberto" - Ok
"Carlos Alberto " - No results
"Carlos Alberto M" - No results
...

There isn't no row with first name = "Carlos Alberto M" OR last name = "Carlos Alberto M"
 
This solution use concatenation of first name and last name to search.

Ex
Full Name: Carlos Alberto Moraes Junior

Search:
"Carlos Alberto" - Ok
"Carlos Alberto " - Ok
"Carlos Alberto M" - Ok
...

Forum related:
https://forum.flowable.org/t/problem-using-white-space-in-the-people-form-field/349


tijsrademakers:
Thanks

PR #406, opened 2017-06-02
Minor version changes

dbmalkovsky:
Made some minor version changes in notice.txt file.

It appears the file needs a further and more complete review, for example, where is carrotsearch used?

I would also help to label the columns in the table; the last two are obvious but it isn't always clear to me as to the meaning of the first two columns.   I would be happy to update this PR as directed.


tijsrademakers:
Yes this file needs some improvements. The full list of dependencies need to be reviewed, because some are missing and others are not used anymore. Labeling the columns in the table makes sense indeed. If you have time to do a proposal with improvements that would be great.

Issue #354, opened 2017-04-19
Password is stored in clear text in ACT_ID_USER table PWD column


mgrouch:
Password is stored in clear text in ACT_ID_USER.
It shouldn't even store encrypted password, it should only store its crypto hash.

Thanks

faizal-manan:
Hi,

Currently I'm working on this issue. Instead of implement password hashing, I decided to reuse existing library. So far I've added support for Spring Security, Apache Common Codec, Jasypt and jBrypt
Let me know if I miss out any other library

Cheers,
Faizal

#419

tijsrademakers:
Thansk Faizal, there's now an option to use password encoders.

wirob:
Where can these options be found when one is using Docker?

I can't seem to find any documentation about this for any of the Docker images.

Thanks

gwendew:
Hi !
Is there an update to this subject ? 
Where can these options be found when one is using Docker?
I can't seem to find any documentation about this for any of the Docker images.

yvoswillens:
Please use the forum for asking questions like this.
This issue was closed almost 5 years ago.

You can provide configuration properties when starting a container using environment variables or as runtime arguments.

F.e.

```
docker run -e flowable.idm.password-encoder=spring_delegating_bcrypt -p8080:8080 flowable/flowable-ui:6.7.2
```

or 
```
docker run -p8080:8080 flowable/flowable-ui:6.7.2 --flowable.idm.password-encoder=spring_delegating_bcrypt 
```

PR #419, opened 2017-06-10
Password is stored in clear text in ACT_ID_USER table PWD column

faizal-manan:
1. Support Spring Security, Apache Common Codec, Jasypt, jBrypt
2. IdentityTest.testAuthentication() and IdentityServiceTest.testCheckPasswordNullSafe() run without any error 
3. Unit test provided

#354


faizal-manan:
#354

faizal-manan:
#354

faizal-manan:
flowable/flowable-engine#354

dbmalkovsky:
It just might be my bias but outputting specific error information is better than catching generic exceptions and printing as generic stack trace:

<pre>
        } catch (Exception e) {
            e.printStackTrace();
        }
</pre>

faizal-manan:
thought abt that too ...
maybe throwing missing jar or incorrect jar version msg error is enough.

modules/flowable-idm-engine/src/main/java/org/flowable/idm/engine/impl/IdmIdentityServiceImpl.java
```diff
@@ -212,4 +221,46 @@ public PrivilegeQuery createPrivilegeQuery() {
...
+    @Override
+    public void setPasswordSalt(PasswordSalt passwordSalt) {
+        this.passwordSalt = passwordSalt;
+    }
+
+    private String getSalt() {
+        Object source = passwordSalt.getSource();
+        if (source instanceof String) return (String) source;
+        try {
+            Class<?> aClass = Class.forName("org.springframework.security.authentication.dao.SystemWideSaltSource");
```
tijsrademakers: Do you see a possibility to implement this in a different way. I think the rest of the classes are defined in a clean way. But I would like to prevent this reflection code if possible.

tijsrademakers:
Hi,

Thanks for the PR, could you have a look at the comment I've added?

Thanks!

modules/flowable-idm-engine/src/main/java/org/flowable/idm/engine/impl/IdmIdentityServiceImpl.java
faizal-manan: The only way I can think off is to stick with 1 provider, spring sec perhaps or creating a bridge to support others provider but I believe it's too much effort since IDM itself can be easily disable.

tijsrademakers: I see your point. I think we should go for 1 provider, probably commons-codec in the flowable-idm-engine package. Then we can add the Spring security option in the flowable-idm-spring module, so there's an alternative that can be configured in a Spring environment. The other ones, Jasypt and JBcypt could be used in the src/test code to show that alternative implementations are also possible, but they should not be part of the src/main code. Then you don't need reflection.

tijsrademakers: @faizal-manan is this something you want to do?

faizal-manan: Yes, still haven't complete yet

faizal-manan: @tijsrademakers code refactored, added new API to updateUserPassword, please review before I bubble up updateUserPassword to the rest of the modules

tijsrademakers:
@faizal-manan thanks for the updates, they look good. One comment, could you move the Spring security provider and encoder to the flowable-idm-spring module? Then we keep the flowable-idm-engine module clean of Spring dependencies and offer people a choice if they want to depend on Spring or not.

faizal-manan:
@tijsrademakers sure and I gonna bubble up IdmIdentityService.updateUserPassword to the rest of the flowable modules. (rest module etc)

faizal-manan:
refactored spring encode service
Just a simple searching, found out the following module will be affected with updateUserPassword method
flowable-idm-api
flowable-idm-engine
flowable-idm-spring
flowable-spring-boot
flowable-engine
flowable5-engine
flowable-ldap
flowable-app-rest
flowable-rest
flowable-ui-idm

PR #425, opened 2017-06-13
timer-test

martin-grofcik:


martin-grofcik:
https://forum.flowable.org/t/nullpointerexception-in-timerutil-java/640

tijsrademakers:
Fixed with slightly different implementation. Due date is mandatory for timer jobs, and an exception is thrown now if no due date could be determined.

Issue #428, opened 2017-06-16
Missing index on ACT_HI_VARINST


joaopsilva:
Hello,

In the `ACT_HI_VARINST` table, there is no index on the `EXECUTION_ID_` column, which results in a full table scan and slow query times when doing a historic process instance query with the `includeProcessVariables` option:

```
HistoricProcessInstance processInstance = historyService.createHistoricProcessInstanceQuery()
    .includeProcessVariables()
    .processInstanceId(processId)
    .singleResult();
```

This option, as defined in `HistoricProcessInstance.xml`, translates into a left outer join with the `ACT_HI_VARINST` table, which can grow rather large when using the `full` configuration for the history level:

```
<if test="includeProcessVariables">
  left outer join ${prefix}ACT_HI_VARINST VAR ON RES.PROC_INST_ID_ = VAR.EXECUTION_ID_ and VAR.TASK_ID_ is null
</if>
```

Would you consider adding the following index to solve this problem? A similar one already exists in the `ACT_RU_VARIABLE`.

```
create index ACT_IDX_HI_PROCVAR_EXE on ACT_HI_VARINST(EXECUTION_ID_);
```

Thank you.

tijsrademakers:
Hi Joao,

Good point, we'll make sure to add the index.

Best regards,

Tijs

tijsrademakers:
Merged your PR, thanks

PR #429, opened 2017-06-16
Create index on ACT_HI_VARINST(EXECUTION_ID_)

joaopsilva:
Index is required to speed-up historic process instance queries which include process variables.

#428


tijsrademakers:
Thanks for the speedy PR, after raising the issue ;-)

PR #434, opened 2017-06-19
Remove extra column in ACT_RU_SUSPENDED_JOB table

dbmalkovsky:
See this [forum post](https://forum.flowable.org/t/adv-handler-cfg-id--column/656l).


tijsrademakers:
Thanks

PR #437, opened 2017-06-22
Updates after latest series of changes

dbmalkovsky:
* unused imports
* use bulk methods instead of iteration
* remove used variables
* remove redundant class variable initialization
* remove toString() method on string variable 
* simplify junit assertion


tijsrademakers:
Hi David, can you remove the changes made to the JUEL code? This is one on one copy of existing JUEL code, and we don't want to make changes to this part of the code.

dbmalkovsky:
I undid the changes in the de.odysseus package.

PR #441, opened 2017-06-26
Flowable 440 - Multidiagram Subprocess Parser does not support artifacts

lsmall:
Need to include artifacts when refactoring subprocess diagrams.

### DIFF
--- a/modules/flowable-bpmn-converter/src/main/java/org/flowable/bpmn/converter/SubprocessXMLConverter.java
+++ b/modules/flowable-bpmn-converter/src/main/java/org/flowable/bpmn/converter/SubprocessXMLConverter.java
@@ -116,7 +116,9 @@ public class SubprocessXMLConverter extends BpmnXMLConverter {
         List<BpmnModel> subModels = new ArrayList<BpmnModel>();
 
         // find all subprocesses
-        Collection<FlowElement> flowElements = model.getMainProcess().getFlowElements();
+        Process main = model.getMainProcess();
+        Collection<FlowElement> flowElements = main.getFlowElements();
+        Collection<Artifact> artifacts = main.getArtifacts();
         Map<String, GraphicInfo> locations = new HashMap<String, GraphicInfo>();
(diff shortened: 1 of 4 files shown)


tijsrademakers:
Hi Lori,

Would it be possible to add a small unit test?

Thanks!

lsmall:
Actually I updated the existing test bpmn file and added annotations. Do you need something more thorough?

tijsrademakers:
Yes saw that, and that's fine. But it would be good to add some assertions for these data objects and text annotations you added in the BPMN file, to the SubProcessMultiDiagramConverterTest unit test. Because without the changes you made to the SubProcessXMConverter I would expect the test to fail then.

lsmall:
sure, will do

lsmall:
My apologies, but I am not git-savvy and cannot figure out how to merge my latest updates.

lsmall:
Opened new pull request https://github.com/flowable/flowable-engine/pull/460

When I tested the latest SubProcessConverterTest with annotations locally, it looks like they are all added to the main process (including any subprocess annotations). It appears that the subprocess isExpanded = false by default now and there are 2 diagrams generated which seems redundant with what I had submitted sometime ago.

I have not had time to explore using this updated parser, but felt there might be a similar issue there with artifacts. I added the equal() method to GraphicInfo and the BPMNDI parsing code to the AbstractConverterTest with the intention of testing this further. If I find it also requires an update, I will open a new issue.

Is the new collapsible/multi-diagram  feature supported for flowable 6 Designer yet?

https://forum.flowable.org/t/collapsed-subprocess-navigation-in-the-web-based-bpmn-modeler/138/22

lsmall:
Any updates on this pull request? Can you merge or do you prefer a different approach?

lsmall:
I added a specific check for the text annotations to ensure they are in the appropriate BPMN DI diagram.

    	// verify annotations are in correct diagram
    	assertTrue(multiMainShapeMap.containsKey("textannotation1"));
    	assertTrue(multiSubShapeMap.containsKey("textannotation2"));

PR #445, opened 2017-06-28
added updateUserPassword function

faizal-manan:
Update the rest of flowable module with new _updateUserPassword_ function. Current _saveUser_ function is not able to recognize either hashing is required or not

`User user = idmIdentityService.createUserQuery().userId("johndoe").singleResult();`
`user.setEmail("newemail@email.org");`
`idmIdentityService.saveUser(user);`
Above code will cause password encoder hashing existing hashed password 

Affected Modules:
flowable-engine, flowable-idm-api, flowable-idm-engine, flowable-rest, flowable-ui-idm, flowable5-engine


tijsrademakers:
Thanks

Issue #448, opened 2017-06-30
Recording of task events for user identity links is inconsistent


reva:
Given a user task and calling `taskService.claim('1', 'flowable')`, flowable will add a comment with action `AddUserLink`. Calling `taskService.unclaim('1')` should create a `DeleteUserlink`comment, but doesn't.

This leads to the recording of an incorrect task history.

I would suggest maybe removing the add*IdentityLinkComment() methods from the HistoryManager interface and moving all of that logic into the AbstractHistoryManager.

reva:
Actually, `claim`doesn't event add a `AddUserLink`, only `setAssignee` does, so even more inconsistency there.

tijsrademakers:
Fixed inconsistency

PR #453, opened 2017-07-03
Recording of task events for user identity links is inconsistent (#448)

reva:
Please have a look at the unit tests. As for the implementation:
- In `ClaimTaskCmd` I tried honoring the contract of `TaskSerive.claim` by doing just what is supposed to be the difference to `TaskService.setAssignee` and then calling `AddIdentityLinkCmd`. Still not that nice.
- I tried to refactor unused/superfluous methods in the history manager. There are still problems left: `org.flowable.engine.impl.history.HistoryManager#createIdentityLinkComment` `create` is always false, `org.flowable.engine.impl.history.HistoryManager#createGroupIdentityLinkComment` `create` is always true and `org.flowable.engine.impl.history.HistoryManager#createUserIdentityLinkComment` `create` and `forceNullUserId` depend on each other


tijsrademakers:
Fixed the user identity links inconsistency manually in the mentioned commit. Thanks for the unit tests!

PR #460, opened 2017-07-05
440 - Multidiagram Subprocess Parser does not support artifacts

lsmall:
Opening new pull request with detailed verification for graphic info and other elements.


lsmall:
Failed test with associations - will revisit and resubmit new pull request

PR #470, opened 2017-07-12
Inconsistency with ACTIVITY_CANCELLED events for multi-instance usertask

stephraleigh:
Different events are received depending on how the multi-instance user task is
cancelled.  We expect the events to be consistent no matter how
cancelled.  These tests reproduce the differences. One test cancels using
terminate end event, and one cancels the multi-instance user task by using
a message boundary event.


stephraleigh:
I updated the pull request. I now see consistency in the multi-instance cancelled events that are received. However I still see a slight ordering difference depending on how the tests are run (in eclipse vs via maven build).  You can see the ordering difference in the 'testMultiInstanceCancelledWhenFlowToTerminateEnd' test as it tries to handle the two scenarios I ran into.

PR #475, opened 2017-07-13
474 - Multi-instance collection syntax proposal

lsmall:
Added a new "collection" extension element of type string as an
alternative to the "collection" attribute so we can support complex
CDATA values.

### DIFF
--- a/modules/flowable-bpmn-converter/src/main/java/org/flowable/bpmn/converter/child/MultiInstanceParser.java
+++ b/modules/flowable-bpmn-converter/src/main/java/org/flowable/bpmn/converter/child/MultiInstanceParser.java
@@ -58,10 +58,19 @@ public class MultiInstanceParser extends BaseChildElementParser {
                     }
 
                 } else if (xtr.isStartElement() && ELEMENT_MULTIINSTANCE_CONDITION.equalsIgnoreCase(xtr.getLocalName())) {
-                    multiInstanceDef.setCompletionCondition(xtr.getElementText());
+                	multiInstanceDef.setCompletionCondition(xtr.getElementText());
 
+                } else if (xtr.isStartElement() && ELEMENT_EXTENSIONS.equalsIgnoreCase(xtr.getLocalName())) {
+                	// save all extension elements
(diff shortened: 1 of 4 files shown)


tijsrademakers:
Hi Lori,

So you want to hardcode such collections inside the BPMN XML? I'm thinking of support for more collection types as an alternative, where the collection is resolved as a process variable. So instead of hard coding it inside the BPMN XML you would set the JSON as a process variable. Would that work for your use case as well, or should it be included in the BPMN XML like you proposed?

lsmall:
We discussed this with Joram [here](https://forum.flowable.org/t/multi-instance-collection-syntax-proposal/615)

First of all, we have a hard requirement to model all variables as explicit data objects in our templates. In other words, we have a history of solutions overloading the template with data objects that have no relevance at runtime (i.e. not used in a service task or decision point). So, we do not allow solutions to create variables directly, but must model it at design time - thus, we need to include it in the BPMN XML.

For multi-instance tasks, we needed a way to represent a string rather than create an explicit data object. We tried using the [collection attribute](https://forum.flowable.org/t/multi-instance-collection-syntax-proposal/615), but had to escape the string to get it to work and it was cumbersome. We felt it would be more efficient to support a string as an element since we could use CDATA.

Finally, we do not support a collection type data object and string support would allow any syntax, not just JSON. I believe Joram agreed that [a collection element is more practical for complex string support than an attribute](https://forum.flowable.org/t/multi-instance-collection-syntax-proposal/615), but you already had the attribute syntax defined. If you accept this argument and prefer I add getters/setters for the new element, I can update my pull request. We could even open a separate issue to support this new model element at runtime as an alternative o the collection attribute (i.e. one must be specified if the elementVariable attribute is present.

Alternatively, it would be handy to support extensions in all your parsers/exporters. Do you agree? If so, we could just use our own extension. However, our preference is to include it as an official extension in the flowable namespace.

tijsrademakers:
Ok thanks for the additional info. I think the PR is fine and yes we should support extension elements in every place.

Issue #48, opened 2016-11-28
DynamicBpmnService only works when the enableProcessDefinitionInfoCache is enabled


ddpardo:
Hi,

While testing the functionality of the DynamicBpmnService I noticed that this does not work unless you enable the ProcessDefinitionInfoCache on the ProcessEngine.

```java
/**
 * Created by Pardo David on 28/11/2016.
 */
public class DynamicBpmnServiceTest {
	private ProcessEngine processEngine;

	@Before
	public void setup(){
		ProcessEngineConfiguration configuration = ProcessEngineConfiguration.createStandaloneInMemProcessEngineConfiguration();
		configuration.setEnableProcessDefinitionInfoCache(true);
		//configuration.setJdbcUrl("jdbc:h2:mem:DynamicBpmnServiceTest;DB_CLOSE_DELAY=1000");
		configuration.setJobExecutorActivate(false);

		processEngine = configuration.buildProcessEngine();
	}

	@After
	public void teardown(){
		processEngine.close();
	}

	@Test
	public void itShouldBePossibleToUpdateTheCandidateUser(){
		InputStream resourceAsStream = this.getClass().getResourceAsStream("/activiti/dynamic-bpmn-test-process.bpmn20.xml");
		RepositoryService repositoryService = processEngine.getRepositoryService();
		Deployment deploy = repositoryService.createDeployment()
				.addInputStream("dynamic-bpmn-test-process.bpmn20.xml", resourceAsStream)
				.deploy();


		RuntimeService runtimeService = processEngine.getRuntimeService();
		String processDefinitionId = repositoryService.createProcessDefinitionQuery().active().singleResult().getId();

		IdentityService identityService = processEngine.getIdentityService();
		User david = identityService.newUser("david");
		identityService.saveUser(david);

		runtimeService.startProcessInstanceById(processDefinitionId);
		TaskService taskService = processEngine.getTaskService();
		Task task = taskService.createTaskQuery().taskCandidateUser("david").singleResult();
		assertThat(task,is(not(nullValue())));

		DynamicBpmnService dynamicBpmnService = processEngine.getDynamicBpmnService();
		ObjectNode processInfo = dynamicBpmnService.changeUserTaskCandidateUser("sid-B94D5D22-E93E-4401-ADC5-C5C073E1EEB4", "bob", true);
		dynamicBpmnService.saveProcessDefinitionInfo(processDefinitionId,processInfo);

[...]

PR #483, opened 2017-07-18
Explicit type can be replaced with <>

dbmalkovsky:
This is based on the [coding style post last year](https://forum.flowable.org/t/coding-style-and-standards/154/4?u=dbmalkovsky).

The specific item is #48: Explicit type can be replaced with <>.

Which read: Correct all new expressions with type arguments which can be replaced with diamond type <>. That is no need to specify the type again.


tijsrademakers:
Thanks, that was one of the biggest PRs until now ;-)

PR #497, opened 2017-07-28
[IDM] Fix a regression about LDAP integration

cdeneux:
The commit148f4f2 introduced a regression in project flowable-ldap-configurator: unit tests fail, and perhaps the LDAP integration is no more working.


jbarrez:
I was actually looking into it too just now due to https://forum.flowable.org/t/ldapintegrationtest-test-case-fail/782, so thanks for figuring out the problem and fixing it :-)

I'll add some tweaks to it (like probably the db tables get created now and this shouldn't be the case when using ldap) and update the docs (if needed).

jbarrez:
Merged. Did some additional tweaks to the fix to make sure it's backwards compatible.

cdeneux:
Thanks Joram

PR #501, opened 2017-07-29
getDatabaseSchemaUpdate() should be "false" or "true",never be null.

teamlet:
getDatabaseSchemaUpdate() return String type value, that default is 'false' and should be 'true' or 'false'.


tijsrademakers:
databaseSchemaUpdate can also have a value of "drop-create" or "create-drop" so this change is not valid.

PR #506, opened 2017-08-01
loop collection as an expression

maxime066:
In a multi-instance tag, we want to use the field collection with an expression. Currently, it doesn't work with the Groovy Scripting engine because the code detect that the field is an expression based on a JUEL syntaxe attribute. This PR fix that.

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/MultiInstanceActivityBehavior.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/MultiInstanceActivityBehavior.java
@@ -267,7 +267,7 @@ public abstract class MultiInstanceActivityBehavior extends FlowNodeActivityBeha
     @SuppressWarnings("rawtypes")
     protected void executeOriginalBehavior(DelegateExecution execution, int loopCounter) {
         if (usesCollection() && collectionElementVariable != null) {
-            Collection collection = (Collection) resolveCollection(execution);
+            Collection collection = (Collection) resolveAndValidateCollection(execution);
 
             Object value = null;
             int index = 0;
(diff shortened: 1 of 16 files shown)


lsmall:
My understanding of the BPMN standard was that expression fields were JUEL-based only. Can one of the Flowable developers confirm?

maxime066:
We extended org.flowable.engine.impl.el.ExpressionManager to create a GroovyExpressionManager that we load in the processEngineConfiguration before the initialization (See org.flowable.engine.cfg.AbstractProcessEngineConfigurator.beforeInit(ProcessEngineConfigurationImpl processEngineConfiguration) ).

I'm working to make the tests pass.

jbarrez:
This would break backwards compatibility for people that assume it's going to be interpreted as a variable. 

I think there might be a way to fix it differently (cause that check on a curly brace is not nice), I'll comment here when it is pushed.

@lsmall : we use JUEL everywhere, but BPMN is pretty lenient when it comes to expressions, it doesn't mandate it. And the expression manager is indeed pluggable.

jbarrez:
@maxime066 / @lsmall : please take a look at https://github.com/flowable/flowable-engine/commit/859c7b801d37326bd80ee8948e83467f506b37fd, it applies the same fix as in the first commit here, but changes the way the collection is resolved: 
* if a collection comes back, all is good
* if it's a String (an expression without ${} will resolve to the String itself), the variable is fetched.

This does mean the collectionVariable is not used anymore, but I left it in and kept it in the resolveCollection(), for people that might have overriden the behavior class.

wdyt, Is this okay?

maxime066:
@jbarrez Seems good to me. This means that with a GroovyExpressionManager, the variable provided can be a collection or a String. If it's a String, your code will look for a variable with the name of the content of the String. This behavior seems good to me but I don't know if it's coherent with the other activity fields.

roberthafner:
Joram, I don't believe this resolves our issue completely. In general we have two problems to solve. The first is how to pass in a string value for the collection that is not  a variable name. The second is how to convert the (JSON) string value that is passed in via the collection to a java.util.Collection instance. We are hoping to be able to use standard flowable extensions for both cases rather than having to override more impl classes.  

Lori has proposed an option to solve the first problem via PR (https://github.com/flowable/flowable-engine/pull/513).  The solution is to make use of <flowable:string> element in the new <flowable:collection> element to distinguish between a string and an expression as well as introduce some new fields in the MultiInstanceActivityBehavior to store the value. As far as I understand the PR isn't flushed out completely but provides a starting point for further discussion.

lsmall:
My proposal uses a new field (collectionString), so these changes will not affect us. As Rob said, we would like to support a string value without assuming it is a variable name. I think this is more standard BPMN behavior - similar to Flowable field values.

In summary, our users will either have knowledge at design time for an explicit assignment (string value) or will set a variable at runtime (dynamic assignment). Since we force all variables to be associated with a data object, this will allow us to use the string value without creating a data object that the user doesn't need.

If you push these changes, I will rebase and update my pull request accordingly. Again, we understand that the parsing of the string (via a registered function?) is missing in our request, but we are overriding the relevant multi-instance behavior methods to work with our syntax for now.

maxime066:
I'm not sure to fully understand but your problems seems to be Juel problems. I'm not sure how's that translate in Groovy.

What's the point of passing a String as the collection variable when you can specify the collection object? Is it because it's easier for the user to write a String in Juel? In Juel the difference is "collectionName" (simpler) vs "${collectionObj}". In groovy it's "'collectionName'" vs "collectionObj" (simpler). So if you use Groovy instead of Juel, your problem disappear.

I'm not sure why you continue to use Juel, I think it's not easy to use (vs Groovy or Javascript), the user need to know if he should use Juel or String (a problem that can probably disapear with Groovy, as I said before) and the script task lets you work with Groovy or Javascript, so Juel is another syntax to learn.

lsmall:
We are not using an expression at all - just a String. The issue is we do not need a variable to "hold" the string value - we just want to process the string directly as part of the definition.

jbarrez:
Proposed changes are merged. Left the tests as-is, as changing them wasn't necessary anymore with the new logic.

@maxime066 : thanks for reporting and the initial fix!

PR #509, opened 2017-08-02
Activity cancelled testing updates

stephraleigh:
Added test case for user task cancellation scenario. Also added validation of execution ids for various cancelled event tests.


jbarrez:
Thanks for the good additions to those tests!

Issue #512, opened 2017-08-04
Multi-instance collection syntax proposal - revised


lsmall:
We have been trying to use the new syntax and would eventually like to include a pluggable function (similar to the expression functions) that would parse the string and return a collection. For now, we would like to update the BPMN syntax and include it in the validation so that it can be deployed. As discussed on the forum [here](https://forum.flowable.org/t/multi-instance-collection-syntax-proposal/615), using a collection element might have been more efficient since you could specify whether the value is a string or expression using a child element. The attributes were less verbose and will need to stay to maintain backwards-compatibility, but we would like to update the collection element syntax to support both types. We will submit a separate proposal for behavior to support pluggable passing functionality based on the child element type. Meanwhile, we will continue to use our taskListener to set the relevant task participants.

lsmall:
Revised syntax is as follows:

        <extensionElements>
          <flowable:collection>
             <flowable:string>
               <![CDATA[
                 [
                   {
                     "principalType" : "User",
                     "role" : "PotentialOwner",
                     "principal" : "wfuser1",
                     "version" : 1
                   },
                   {
                     "principalType" : "User",
                     "role" : "PotentialOwner",
                     "principal" : "wfuser2",
                     "version" : 1
                   }
                 ]
                ]]>
             </flowable:string>
          </flowable:collection>
        </extensionElements>

where the flowable:string child element can be replaced by a flowable:expression element if an expression is used.

jbarrez:
@lsmall : thanks for the PR. It's merged. 

You referenced this issue here: https://github.com/flowable/flowable-engine/blob/e44e182932a350cbb78f42cbae5cb7c1c6dbc3ee/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/MultiInstanceActivityBehavior.java#L317. Do you want to keep this issue open for resolving the static string value?

lsmall:
I will open a new one for the function support - thanks!

lsmall:
Opened issue [520](https://github.com/flowable/flowable-engine/issues/520) for adding function support

lsmall:
Noting revised PR [518](https://github.com/flowable/flowable-engine/pull/518) for this issue that has been merged and closed.

PR #513, opened 2017-08-04
512 Multi-instance collection syntax proposal - revised

lsmall:
Added child elements to support either string or expression value for
collection

Also added new syntax to process validation

issue[512](https://github.com/flowable/flowable-engine/issues/512)


lsmall:
My local build worked, I will look into this error.

jbarrez:
The proposed syntax is ok: it offers the same functionality as the attribute, and an alternative with a static String value (which can't be solved with an attribute due to being xml).

I'll apply the changes discussed in https://github.com/flowable/flowable-engine/pull/506 so you can finish this PR.

lsmall:
Need to rebase on latest code with [PR506](https://github.com/flowable/flowable-engine/pull/506)

PR #516, opened 2017-08-07
Filter process definition by potential starter in task app

paulstapleton:
Hi,

I've created this pull request after discussion in the forum https://forum.flowable.org/t/how-to-restrict-the-start-of-process/257

There are two separate commits here:

1) Fixed a bug in the JSON <-> BPMN model converter. The potential starters were not populated in the JSON output in the function `BpmnJsonConverter#convertToJson(BpmnModel model, Map<String, ModelInfo> formKeyMap, Map<String, ModelInfo> decisionTableKeyMap)`. This meant that exporting and importing an app in the flowable-modeler would remove the potential starter users/Groups when re-importing. I have updated a unit test to verify this behaviour.

2) Added a `CandidateManager ` which uses the `RemoteIdmService` and added logic to filter out process definitions that are not startable by the current user in the flowable-task app. If no starters are defined for a process definition then its assumed to be startable by everyone.

I decided to do this logic in the flowable-task app and not in the process Engine to maintain backwards compatibility. Today if we query process definitions by a start user (`ProcessDefinitionQueryImpl#startableByUser(String userId)`) and no potential starters are defined in the BPMN then the process definition is filtered out for all userIds. I'm not sure of the consequences of making a change there so I have restricted the fix to the flowable-task app.

Regards,
Paul


tijsrademakers:
Thanks, made some changes to prevent needing a remote candidate manager. Let me know if you see issues with these changes.

PR #523, opened 2017-08-10
520 - Add support for revised syntax collectionString parsing function

lsmall:
1st draft for collection string function parsing support. This is still a WIP, but I would like to request comments at this point.

The example (json:parseJson) function is not really reusable, so I would prefer to put it in the test branch, but I could not figure out how to get it added to the list of functional delegates prior to deploying the template. I was able to do this for the collection string in the test case itself because it is not required for deployment, only at runtime for collection resolution.

    @Deployment
    public void testParallelUserTasksCustomCollectionStringExtension() {
        // register function delegate
        processEngineConfiguration.getExpressionManager();
        List<FlowableFunctionDelegate> functionDelegates = expressionManager.getFunctionDelegates();
        functionDelegates.add(new FlowableCollectionFunctionDelegate());

    	checkParallelUserTasksCustomCollectionStringExtension("miParallelUserTasksCollection");
    }

But the 2 new expression test cases require the function at deployment. If you can help me figure out how to register it just for the test class, then I can move the example function and its delegate class to the test branch.

We are using a proprietary participant class and a task listener to do the assignment after collection resolution and task instantiation. This example is less complex and leverages the candidate user field. I was trying to keep it simple and use a single parameter function method.


lsmall:
Waiting for Rob and Flowable team to review. I suspect this PR will need to be revised and is not ready to merge yet.

jbarrez:
I'm wondering if an expression is the right approach here? I definitively see why you need the name+prefix to call the expression, but it feels inconsistent with other expression usages. Of course you can't do something like '${myExpression(collectionString)}, as this would need some hackery of the expression resolving only for this particular feature.

Is there a compelling reason to go for an expression? Wouldn't a custom interface (simply passing in the string value to its method) make more sense, similar to how it is currently possibl to have a custom http request handler for the http task behavior (see https://github.com/flowable/flowable-engine/blob/233489ecc443ec709bd931e34dc1214a22e394eb/modules/flowable-http/src/main/java/org/flowable/http/impl/HttpActivityBehaviorImpl.java#L143)?

lsmall:
Yes, we agree that handling it as a JUEL expression seems forced, but we were trying to reuse the standard string vs expression usage scenarios. We were thinking that a generic custom function of some kind rather than an expression function would be more appropriate, but did not see any obvious way to implement it. I will review your suggestion and see if we can refactor this PR to use a custom interface instead. Thanks for your feedback!

lsmall:
Creating revised implementation in new PR for review.

lsmall:
Opened new PR [527](https://github.com/flowable/flowable-engine/pull/527)

lsmall:
Created new PR [533](https://github.com/flowable/flowable-engine/pull/533)

PR #532, opened 2017-08-17
Explicit type can be replaced with <>

dbmalkovsky:
Part 2 of [PR 515](https://github.com/flowable/flowable-engine/pull/515) which read in part:

>This is based on the coding style post last year.
>
>The specific item is #48: Explicit type can be replaced with <>.
>
>Which read: Correct all new expressions with type arguments which can be replaced with diamond type <>. That is no need to specify the type again.


dbmalkovsky:
FWIW, did not get the failure when running in my fork: https://travis-ci.org/dbmalkovsky/flowable-engine

PR #533, opened 2017-08-17
520 revised - Add support for revised syntax collectionString parsing function

lsmall:
Updated handler syntax (attribute rather than element) and added class and delegate implementation support as delegate handler classes


tijsrademakers:
Thanks for all the work in the PR. I've made a couple of changes, but the main logic remained the same. Let me know if you see issues with these changes.

PR #536, opened 2017-08-18
Fix package name and sample BPMN 

dbmalkovsky:
Originally was just going to change the package name on the Task import but also changed product name in the sample BPMN file.


tijsrademakers:
Thanks!

PR #540, opened 2017-08-21
--Fixed query with groupIds

liaokevin:
when i use fllowing code,there is a problem with mybaits mapping;
`processEngine.getIdentityService() .createUserQuery().memberOfGroups(groupIds).list()`


tijsrademakers:
Thanks!

PR #542, opened 2017-08-22
Resolve bug in multi-instance parsing.

roberthafner:
Change-Id: I9a6d3b7612ea7b443151cd12d7f94dc8a75907c0

The MultiInstanceParser fails to include flow elements after a multi-instance user task that uses the new <flowable:collection> element. This failure can be reproduced by reordering the flow elements in the bpmn file. The changeset includes a reordered bpmn file which can be used to reproduce the error as well as a proposed fix.


tijsrademakers:
Thanks, merged manually

PR #554, opened 2017-09-06
Fix upgrade scripts for ACT_HI_VARINST

dpratt:
The 6.1.2 -> 6.2.0 upgrade scripts for h2, hsql and mysql had the wrong table names in the history upgrade scripts.


jbarrez:
Thanks, good catch!

PR #572, opened 2017-09-18
Centralise Camel dependency and upgrade to latest

zwets:
This PR adds a `camel.version` property to the root pom and adds the Camel dependencies to `dependencyManagement`. Second commit upgrades Camel 2.17.1 to 2.19.3.

### DIFF
--- a/modules/flowable-camel/pom.xml
+++ b/modules/flowable-camel/pom.xml
@@ -32,8 +32,8 @@
       <groupId>org.springframework</groupId>
       <artifactId>spring-test</artifactId>
       <scope>provided</scope>
-	 </dependency>
-   <dependency>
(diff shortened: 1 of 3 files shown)


PascalSchumacher:
Camel 2.18+ requires java 8 and flowable only requires java 7, so I guess this is not possible.

zwets:
Oww, you're right. Now I remember why I kept it on that branch :-]

Sent from ProtonMail mobile

-------- Original Message --------

tijsrademakers:
Thanks, kept version 2.17.1 for now. We are thinking of dropping JDK 7 support starting from Flowable 6.3.x, but need to get some community feedback on this first.

PR #576, opened 2017-09-19
Correctly bind headers

### LINKED ISSUE #575: Unable to bind Headers to a Http task
Unable to bind the http headers to an Http task with keys that have a : as a value. When we look at the exception stack (detailed below) and check line 264 of the code, it has a check that prevents values of http headers from having a colon in it (Code snippet also included below).
HTTP_TASK_REQUEST_HEADERS_INVALID = requestHeaders are invalid

Stack:

org.flowable.engine.common.api.FlowableException: requestHeaders are invalid
at org.flowable.http.impl.HttpActivityBehaviorImpl.setHeaders(HttpActivityBehaviorImpl.java:264) ~[flowable-http-6.1.2.jar:6.1.2]

Code from HttpActivityBehaviorImpl

protected void setHeaders(final HttpMessage base, final String headers) throws IOException {
try (BufferedReader reader = new BufferedReader(new StringReader(headers))) {
String line = reader.readLine();
while (line != null) {
String[] header = line.split(":");
if (header.length == 2) {
base.addHeader(header[0], header[1]);
line = reader.readLine();
} else {
throw new FlowableException(HTTP_TASK_REQUEST_HEADERS_INVALID);
}
}
}
}

Sample Header that we are passing
host:localhost:7000
connection:keep-alive
content-length:48
postman-token:ffff44c7-e789-ba46-2f5a-8271b4dd7aba
cache-control:no-cache
origin:chrome-extension://fhbjgbiflinjbdggehcddcbncdddomop
user-agent:Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/60.0.3112.113 Safari/537.36
content-type:application/json
accept:/
accept-encoding:gzip, deflate, br
accept-language:en-US,en;q=0.8

filiphr:
Each header line is split only on first ':' everything before ':' is considered as name and everything after as value

Fixes #575 

I couldn't find a test that tests this behaviour. Is there something that actually tests this class?

### DIFF
--- a/modules/flowable-http/src/main/java/org/flowable/http/impl/HttpActivityBehaviorImpl.java
+++ b/modules/flowable-http/src/main/java/org/flowable/http/impl/HttpActivityBehaviorImpl.java
@@ -20,6 +20,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Timer;
 import java.util.TimerTask;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLSession;
@@ -77,6 +79,7 @@ public class HttpActivityBehaviorImpl extends HttpActivityBehavior {
 
     private static final long serialVersionUID = 1L;
     private static final Logger LOGGER = LoggerFactory.getLogger(HttpActivityBehaviorImpl.class);
+    private static final Pattern HEADER_PATTERN = Pattern.compile("(.+):(.+)");
     
     protected HttpServiceTask httpServiceTask;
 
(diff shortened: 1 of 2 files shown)


jbarrez:
The test for http task can be found here: https://github.com/flowable/flowable-engine/blob/master/modules/flowable-http/src/test/java/org/flowable/http/HttpServiceTaskTest.java

But I don't see a specific test for the headers.

filiphr:
Thanks a lot, I couldn't really find where and how the class is instantiated and it was a bit late :). I'll add a test for this. There are some calls that set the headers, I can set the custom one as well. I would improve the tests as well. And add something like expect to the server so we know that the stuff that were sent is what we would expect to send and not that it just passed by some accident :)

filiphr:
@jbarrez I've added something to the test. However, I am not satisfied with how well it is tested. The previous build passed, but it should have failed (as it should have looped without breaking). What does the Travis CI really do?

sachpal:
Thanks for the fix and update.

PR #579, opened 2017-09-21
578 Multi-instance export not working properly

lsmall:
The flowable prefix is missing from the collection element and the
collection expression is always set as the collection attribute.
Unfortunately, this causes any custom collection handler to get lost
upon export.

I have added a new unit test for these conditions and refactored the
handler export.


tijsrademakers:
Merged manually, thanks

PR #580, opened 2017-09-22
add candidateAssignedToOthers flag to TaskInfoQuery interface

mlippens:
For discussion see on flowable forum: https://forum.flowable.org/t/taskquery-api-questions-multiple-assignees-and-candidates-that-have-been-assigned/131/6

If needed I am of course willing to adapt this pull request.

git commit message:This extra flag has been implemented in order to allow querying without assignee logic. When setting the flag when querying, one can still request tasks or historic tasks regardless of the set assignee.The behaviour without this flag is that a task that is assigned can never be returned when searching for candidateUsers.

### DIFF
--- a/modules/flowable-task-service/src/main/resources/org/flowable/task/db/mapping/entity/Task.xml
+++ b/modules/flowable-task-service/src/main/resources/org/flowable/task/db/mapping/entity/Task.xml
@@ -686,7 +686,9 @@
         and RES.PARENT_TASK_ID_ IS NULL
       </if>
       <if test="!bothCandidateAndAssigned &amp;&amp; (candidateUser != null || candidateGroups != null)">
-        and RES.ASSIGNEE_ is null
+        <if test="!candidateAssignedToOthers">
+          and RES.ASSIGNEE_ is null
+        </if>
         and I.TYPE_ = 'candidate'
         and 
         ( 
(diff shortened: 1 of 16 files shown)


modules/flowable-task-service/src/main/resources/org/flowable/task/db/mapping/entity/Task.xml
```diff
@@ -802,23 +804,31 @@
           <!-- if dbIdentityUsed set true in process engine configuration -->
           <if test="userIdForCandidateAndAssignee != null">
             <if test="candidateGroups == null">
-              and (RES.ASSIGNEE_ = #{userIdForCandidateAndAssignee} or (RES.ASSIGNEE_ is null and (I.USER_ID_ = #{userIdForCandidateAndAssignee}
-              or I.GROUP_ID_ IN (select g.GROUP_ID_ from ${prefix}ACT_ID_MEMBERSHIP g where g.USER_ID_ = #{userIdForCandidateAndAssignee} ) ) ) )
+              and (RES.ASSIGNEE_ = #{userIdForCandidateAndAssignee}
+                or (
+                <if test="candidateAssignedToOthers">
```
tijsrademakers: Should this be <if test="!candidateAssignedToOthers"> instead?

tijsrademakers:
Thanks, added a comment about an if check, could you validate if my assumption is correct?

modules/flowable-task-service/src/main/resources/org/flowable/task/db/mapping/entity/Task.xml
mlippens: I assume you meant the missing negation? You are correct, good catch, I updated it.
It seems that that branch of the query is never tested (neither in the new tests nor in the original ones), hence why it slipped through.

tijsrademakers:
Merged the PR manually, thanks.
I've changed the boolean to ignoreAssigneeValue, hope that's ok for you.

mlippens:
@tijsrademakers that's quite alright.

Issue #11, opened 2016-10-19
Provide Transaction depentent EventListeners


wblancqu:
Just as with the new TransactionDependent Task/Execution Listener (as nicely demonstrated by Yvo in his blogpost: http://blog.mark-it-zero.com/transaction-dependent-listeners/), it might make sense to have the same kind of functionality for EventListeners.

e.g. I want to notify on a Kafka topic each time a process instance has started / ended, but do not want to clutter the BPMN xml with that. The current implementation (if I'm not mistaking) will notify the EventListeners even when the transaction has not yet been committed. I want to be notified only when I'm sure that the process instance really started / ended.

In the project I'm currently working on, we implemented this functionality on top of Flowable (/Activiti). Do you guys think it makes sense to move this closer to the core?

jbarrez:
Yes, it does make sense if the same transaction lifecycle semantics would be added for event listeners too. 

There are three ways of adding such event listeners (see http://www.flowable.org/docs/userguide/index.html#eventDispatcher):

1. Via process engine configuration
2. Via the API (RuntimeService)
3. in the bpmn 2.0 xml

Implementation-wise, I'm assuming a new more specialized event listener interface is need (extending the current event listener interface would not be backwards compatible). When the event is fired (here: https://github.com/flowable/flowable-engine/blob/master/modules/flowable-engine/src/main/java/org/activiti/engine/delegate/event/impl/ActivitiEventDispatcherImpl.java#L65), the event listener should be wrapper in a TransactionListener, like is done for the task/executionListener (see https://github.com/flowable/flowable-engine/blob/master/modules/flowable-engine/src/main/java/org/activiti/engine/impl/bpmn/listener/ListenerNotificationHelper.java#L185).

For 1 and 2, the current API's could be kept or a new methods are added where a transaction state is passed as extra argument. For 3, a new attribute is needed.

wblancqu:
I'll try working on this the coming next days, but quite a busy schedule. I'll reach out asap :-)

Grtz

mlippens:
@wblancqu any updates on this? Seems like an interesting feature to have.

wblancqu:
Sorry, no longer working actively on this. Feel free to pick this up!

Grtz

faizal-manan:
By adding this functionality for EventListener, does it'll affect FlowableEventListener?
or something totally new is required, like TransactionDependentEventListener & FlowableTransactionDependentEventListener

tijsrademakers:
[...]

PR #589, opened 2017-10-02
Issue #11 : Transaction dependent EventListener

faizal-manan:
as in issue #11 

Related unit test
[ActivityEventsTest](https://github.com/faizal-manan/flowable-engine/blob/transaction-event-listener/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/event/ActivityEventsTest.java)
[ProcessDefinitionScopedTransactionEventListenerTest(]https://github.com/faizal-manan/flowable-engine/blob/transaction-event-listener/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/event/ProcessDefinitionScopedTransactionEventListenerTest.java)
[ActivityEventTransactionDependentsTest](https://github.com/faizal-manan/flowable-engine/blob/transaction-event-listener/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/event/ActivityEventTransactionDependentsTest.java)
[TestTransactionFlowableEventListener](https://github.com/faizal-manan/flowable-engine/blob/transaction-event-listener/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/event/TestTransactionFlowableEventListener.java)
[ProcessDefinitionScopedTransactionEventListenerDefinitionTest](https://github.com/faizal-manan/flowable-engine/blob/transaction-event-listener/modules/flowable-engine/src/test/java/org/flowable/standalone/event/ProcessDefinitionScopedTransactionEventListenerDefinitionTest.java)

### DIFF
--- a/modules/flowable-bpmn-converter/src/main/resources/org/flowable/impl/bpmn/parser/flowable-bpmn-extensions-6.0.xsd
+++ b/modules/flowable-bpmn-converter/src/main/resources/org/flowable/impl/bpmn/parser/flowable-bpmn-extensions-6.0.xsd
@@ -654,7 +654,97 @@
     </complexType>
   </element>
 
-  <element name="in">
+    <element name="transactionEventListener">
+        <annotation>
+            <documentation>
+                Extension element for defining transaction-dependent-event-listeners on a
+                process-definition.
+            </documentation>
+        </annotation>
+        <complexType>
+            <attribute name="events">
+                <annotation>
+                    <documentation>
+                        Comma-separated list of event-types an event-listener is configured to be
+                        notified of. Can also be a single event-type.
+                    </documentation>
+                </annotation>
+            </attribute>
+            <attribute name="onTransaction" use="required">
+                <annotation>
+                    <documentation>
+                        Execute executionListener after a successful commit or after a rollback.
+                    </documentation>
+                </annotation>
+                <simpleType>
+                    <restriction base="string">
+                        <enumeration value="before-commit"/>
+                        <enumeration value="committed"/>
+                        <enumeration value="rolled-back"/>
+                    </restriction>
+                </simpleType>
+            </attribute>
+            <attribute name="entityType">
+                <annotation>
+                    <documentation>
+                        Type of entity that should be targeted by events for which the
+                        event-listener should be notified.
+                    </documentation>
+                </annotation>
+                <simpleType>
+                    <restriction base="string">
+                        <enumeration value="attachment"/>
+                        <enumeration value="comment"/>
+                        <enumeration value="execution"/>
+                        <enumeration value="identity-link"/>
+                        <enumeration value="job"/>
+                        <enumeration value="process-definition"/>
+                        <enumeration value="process-instance"/>
+                        <enumeration value="task"/>
+                    </restriction>
+                </simpleType>
+            </attribute>
+            <attribute name="throwEvent">
+                <annotation>
+                    <documentation>
+                        Type of event to be throw as a response to a matching activiti-event being
+                        dispatched.
+                    </documentation>
+                </annotation>
+                <simpleType>
+                    <restriction base="string">
+                        <enumeration value="signal"/>
+                        <enumeration value="globalSignal"/>
+                        <enumeration value="message"/>
+                        <enumeration value="error"/>
+                    </restriction>
+                </simpleType>
+            </attribute>
+            <attribute name="signalName">
+                <annotation>
+                    <documentation>
+                        Name of a signal.
+                    </documentation>
+                </annotation>
+            </attribute>
+            <attribute name="messageName">
+                <annotation>
+                    <documentation>
+                        Name of a message.
+                    </documentation>
+                </annotation>
+            </attribute>
+            <attribute name="errorCode">
+                <annotation>
+                    <documentation>
+                        Error-code of an error event.
+                    </documentation>
+                </annotation>
+            </attribute>
+        </complexType>
+    </element>
+
+    <element name="in">
     <annotation>
       <documentation>
         Element to specify Data Input in Flowable Shortcuts
--- a/modules/flowable-bpmn-model/src/main/java/org/flowable/bpmn/model/BpmnModel.java
+++ b/modules/flowable-bpmn-model/src/main/java/org/flowable/bpmn/model/BpmnModel.java
@@ -12,15 +12,15 @@
  */
 package org.flowable.bpmn.model;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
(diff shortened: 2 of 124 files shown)


tijsrademakers:
Hi, 

Thanks for the PR. In case transaction dependent event listeners are used, I think this replaces the default event listener. Right now there's a lot of additional code that adds dispatchTransactionEvent methods in addition to the existing dispatchEvent methods. But would it not work better if you can enable transaction dependent event listeners in the process engine configuration and that the existing event dispatcher code is used in that case but now with the transaction logic? Then throughout the code only the dispatchEvent methods are needed, and the transaction check is implemented there.

faizal-manan:
Hi Tj,
The changes has been made based my I understanding on the issue. Since new Listener should be used, I do assume everything need to had it own, including eventDispatcher.
So, how does design should be ...
FlowableEventListener & TransactionFlowableEventListener should shared the same eventDispatcher? I think I've done this somewhere on Jul, before I made this massive changes. Give me sometime to revert to early version.

cheers

faizal-manan:
The earlier version had it own transactionEventDispatcher [https://github.com/faizal-manan/flowable-engine/commit/8800aa027f218e1ad8dd8b41f7b766ded6f724e2#diff-fa533f46d338cdabe8ed6f04c80a87c3R79](https://github.com/faizal-manan/flowable-engine/commit/8800aa027f218e1ad8dd8b41f7b766ded6f724e2#diff-fa533f46d338cdabe8ed6f04c80a87c3R79)

I do create it own transactionEventDispatcher & transactionEventSupport to avoid messing up with current eventDispatcher. Are you ok with this, or transactionEventDispatcher & transactionEventSupport should be dropped too.

tijsrademakers:
Ok understood, I think it's ok to use the same eventDispatcher, so no transactionEventDispatcher is needed. It needs basically a switch between the current event listener and a transaction event listener in the same dispatcher.

faizal-manan:
changes has been made,
To avoid adding more _addListener_ function in FlowableEventDispatcherImpl and FlowableEventSupport, I had to change TransactionFlowableEventListener to extend FlowableEventListener, as in following
[FlowableEventSupport.java](https://github.com/faizal-manan/flowable-engine/blob/373805a4f7cf490206e297760138226d75118b04/modules/flowable-engine-common/src/main/java/org/flowable/engine/common/impl/event/FlowableEventSupport.java#L110)
[FlowableEventDispatcherImpl.java](https://github.com/faizal-manan/flowable-engine/blob/373805a4f7cf490206e297760138226d75118b04/modules/flowable-engine-common/src/main/java/org/flowable/engine/common/impl/event/FlowableEventDispatcherImpl.java#L89)
[TransactionFlowableEventListener.java](https://github.com/faizal-manan/flowable-engine/blob/373805a4f7cf490206e297760138226d75118b04/modules/flowable-engine-common-api/src/main/java/org/flowable/engine/common/api/delegate/event/TransactionFlowableEventListener.java#L9)

next,

1. [BpmnModelEventDispatchAction](https://github.com/faizal-manan/flowable-engine/blob/373805a4f7cf490206e297760138226d75118b04/modules/flowable-engine/src/main/java/org/flowable/engine/delegate/event/impl/BpmnModelEventDispatchAction.java#L44) should trigger one dispatchEvent as in FlowableEventDispatcherImpl, right?
2. Should I remove transactionEventListener element in [flowable-bpmn-extensions-6.0.xsd](https://github.com/faizal-manan/flowable-engine/blob/transaction-event-listener/modules/flowable-bpmn-converter/src/main/resources/org/flowable/impl/bpmn/parser/flowable-bpmn-extensions-6.0.xsd) and add onTransaction support to eventListener listener element? Personally I think should not remove it, otherwise [/DefaultListenerFactory.java](https://github.com/faizal-manan/flowable-engine/blob/transaction-event-listener/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/parser/factory/DefaultListenerFactory.java) required to determine what kind of instance it should return.

cheers

tijsrademakers:
@faizal-manan sorry for the big delay, let's try to get this merged quickly now :-)
There are a lot of commented pieces of code, is there some cleanup needed?
Also saw this:

public class FlowableEventListenerNotificationHelper {
 +
 +    public void dispatchEvent(FlowableEvent flowableEvent) {
 +
 +        if (flowableEvent instanceof FlowableEventImpl) {
 +
 +        }
 +    }

Should this be cleaned up as well?

Thanks

faizal-manan:
@tijsrademakers I think I've clean-up, let me double check again

Thnkas

faizal-manan:
Hi Tijs,

I've complete removed unused class and comments.

Thanks

modules/flowable-bpmn-model/src/main/java/org/flowable/bpmn/model/BpmnModel.java
```diff
@@ -564,4 +565,13 @@ public Object getEventSupport() {
     public void setEventSupport(Object eventSupport) {
         this.eventSupport = eventSupport;
     }
+
```
tijsrademakers: Can you remove these lines?

modules/flowable-bpmn-model/src/main/java/org/flowable/bpmn/model/Process.java
```diff
@@ -291,6 +292,14 @@ public void setEventListeners(List<EventListener> eventListeners) {
         this.eventListeners = eventListeners;
     }
 
+//    public List<TransactionEventListener> getTransactionEventListeners() {
+//        return transactionEventListeners;
```
tijsrademakers: Can you remove these lines?

modules/flowable-bpmn-converter/src/main/resources/org/flowable/impl/bpmn/parser/flowable-bpmn-extensions-6.0.xsd
```diff
@@ -654,7 +668,97 @@
     </complexType>
   </element>
 
-  <element name="in">
+    <!--<element name="transactionEventListener">-->
```
tijsrademakers: This can be removed right?

tijsrademakers:
Hi @faizal-manan,

Thanks for the cleanup, but I still see lines commented out here https://github.com/flowable/flowable-engine/pull/589/files.

I've added a number of comments in the PR, but there are a lot more cases like that. Could you have a look at it?

Thanks

faizal-manan:
Hi @tijsrademakers 

I've done clean up and removed all unrelated commit. Most of it is due to IDE code formatting.

jbarrez:
PR merged, thanks! 

Moved the onTransaction to the FlowableEventListener, so there is no need anymore for a new separate transaction dependent event listener (see latest referenced commit).

Issue #592, opened 2017-10-03
Add Data Object support for Modeler


lsmall:
https://forum.flowable.org/t/how-to-model-data-objects-in-the-web-designer/948/4

Some of the underlying support was there from the original implementation.

lsmall:
Merged and tested

PR #593, opened 2017-10-03
592 Add Data Object support for Modeler

lsmall:
Added view and translation elements

Also fixed some conversion errors in model setValue()

https://github.com/flowable/flowable-engine/issues/592

### DIFF
--- a/modules/flowable-bpmn-model/src/main/java/org/flowable/bpmn/model/BooleanDataObject.java
+++ b/modules/flowable-bpmn-model/src/main/java/org/flowable/bpmn/model/BooleanDataObject.java
@@ -12,6 +12,8 @@
  */
 package org.flowable.bpmn.model;
 
+import org.apache.commons.lang3.StringUtils;
+
 /**
  * @author Lori Small
  */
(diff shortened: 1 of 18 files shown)


PascalSchumacher:
Great. 👍 

Just a few days ago I was telling my colleagues, that data object support for modeler would be really nice.

tijsrademakers:
Hi Lori,

I've merged the PR with some minor changes. I've reverted the changes to the JSON property naming because there could be issues with backward compatibility. Could it be that some parts are missing in this PR? Because there's no DataObject element added to the Modeler palette, so it's not appearing in the Modeler to use in the BPMN process design.

lsmall:
The data object is a property on the process or subprocess, not a visual element. I considered leaving the "items" element in the JSON property naming, but was trying to emulate the form properties syntax. I will verify it with the reverted syntax and open a new issue if it has problems. Thanks!

tijsrademakers:
Ah right, now I understand. I've made a couple of changes locally to make it work. Will push them a bit later.

lsmall:
Let me know when you are done and I will run a quick regression test.

tijsrademakers:
Done with the changes

lsmall:
No issues found in regression test. The exported BPMN syntax looks good and the unit test passed.

PR #601, opened 2017-10-10
Issue 600 - process validation fails for multi-instance extensions

lsmall:
Deployed processes are order-sensitive and the extension elements must
be placed before any additional child elements. Moved the extension code
up in the MultiInstanceExport class. Also added a non-extension element
to the unit test template.


lsmall:
Can we get this merged? I think it is fairly straightforward and we have tested it locally with our deployment and are no longer getting a Flowable error for process validation.

tijsrademakers:
Merged it manually, thanks

PR #602, opened 2017-10-13
Move REST URL creation

stijndepestel:
I've created this pull request to simplify the use of the modeler code in other applications with, for example, another REST service. Hence I have created the APP_URL namespace to hold functions that return the REST URLs. (Note, I have not used the URL namespace as to not clash with the editor, of course, they can always be combined or renamed).

I have also created a forum post (https://forum.flowable.org/t/bringing-all-rest-urls-together/1060) to include everyone in the discussion of whether or not this is useful.


tijsrademakers:
Thanks

PR #604, opened 2017-10-13
default process - DMN engine expression manager configuration

martin-grofcik:


tijsrademakers:
The SpringDmnEngineConfigurator is a convenience class that can be used to easily use the DMN engine in a Spring context. For this use case the expression manager is always set to the SpringDmnExpressionManager. If you implement your own expression manager then it's better to use a modified version of this convenience class.

PR #606, opened 2017-10-16
Update API Swagger Documentation

jmpascal:
Update Swagger plugin version
Update Swagger annotation & Fix doc issues (Http Response...)
Improve documentation coverage


PascalSchumacher:
I guess it would also be helpful to move the swagger versions to the dependency section of the root pom. What do you think?

PR #608, opened 2017-10-18
Use specific annotations from spring mvc instead of RequestMapping

filiphr:
This is nothing major just using the Spring 4 meta annotations. I didn't do all of the endpoints as there a quite a few and wanted to see if this is something that you might want to have in the project.

### DIFF
--- a/modules/flowable-rest/src/main/java/org/flowable/rest/service/api/form/FormDataResource.java
+++ b/modules/flowable-rest/src/main/java/org/flowable/rest/service/api/form/FormDataResource.java
@@ -29,9 +29,9 @@ import org.flowable.rest.service.api.RestResponseFactory;
 import org.flowable.rest.service.api.runtime.process.ProcessInstanceResponse;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestBody;
-import org.springframework.web.bind.annotation.RequestMapping;
-import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RequestParam;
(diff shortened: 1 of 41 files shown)


filiphr:
By the way, I am also looking into using `ReponseEntity` instead of directly working with the `HttpRequest / Reponse`

jmpascal:
@tijsrademakers After investigation @GetMapping @PostMapping @PutMapping @DeleteMapping are not supported by swagger-maven-plugin see https://github.com/kongchen/swagger-maven-plugin/issues/451

It means it breaks the swagger documentation generation.

2 solutions: 
+ Fork the swagger-maven plugin to support it (& implement this solution https://github.com/kongchen/swagger-maven-plugin/pull/491/commits/6267734835111f0e4c76007659be7534c3221cf0 )
+ Revert GetMapping to the old annotation

filiphr:
Ops, didn't know that. Sorry.

We can also use jitpack.io to have a specific build from github (not sure though if it works for plugins).

Reverting is also fine for me if you prefer that.

And maybe most important. we should have some test for this (maybe have the file in source control)

tijsrademakers:
Forking the Swagger Maven plugin feels like the best solution for me. Would this be something you could do @jmpascal? Maybe we can just add it to this Flowable Github repo for now?

jmpascal:
After morning investigation with a fresh mind it appears the issue is only visible if we use the version 3.1.4 (which is the one currently used in Flowable master).

With PR https://github.com/flowable/flowable-engine/pull/606 this issue is solved as it bump up the version to 3.1.5 and fix the issue.

It's another good reason to accept this PR ;) @tijsrademakers

filiphr:
The commit you linked is not in 3.1.5 from the plugin. However, it works because 3.1.5 uses spring version 4.3.7.RELEASE and `AnnotatedElementUtils#findMergedAnnotation` which does the magic for getting the information out of the more specific annotations :).

Actually the commit you linked is not even on master (it is an old commit which is no where)

PR #614, opened 2017-10-20
update cglib to version 3.2.5 (which is java 9 compatible)

PascalSchumacher:


PascalSchumacher:
No really sure if this can cause conflicts with any transitive dependency, so let's close this for now.

PR #618, opened 2017-10-22
Generify DataResponse and PaginateList

filiphr:
Use generics and type safety for the `DataResponse` and `XXXPaginateList`.

I am looking into adding spring rest docs so documentation is automatically generated in asciidoc, some improvements into the mappings and this is a step into that direction. This would also make it easier not to make a mistake when implementing some new stuff :)


filiphr:
Wow you are fast @tijsrademakers :)

Issue #622, opened 2017-10-30
Optimize service task execution


cdeneux:
Definition maps (imports, item, message and operations ) of a service task are loaded for each execution. They should be loaded when creating the service task or on the first execution.

Moreover, this change can fix potential NPE occurring on service task when managing definition maps concurrently.

cdeneux:
I have proposed PR 625

tijsrademakers:
Thanks, merged the PR

PR #625, opened 2017-10-30
Optimize service task execution

cdeneux:
See issue #622

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/WebServiceActivityBehavior.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/WebServiceActivityBehavior.java
@@ -77,12 +77,24 @@ public class WebServiceActivityBehavior extends AbstractBpmnActivityBehavior {
     protected Map<String, ItemDefinition> itemDefinitionMap = new HashMap<String, ItemDefinition>();
     protected Map<String, MessageDefinition> messageDefinitionMap = new HashMap<String, MessageDefinition>();
 
+    private boolean definitionMapsFilled = false;
+
+    private final Object definitionMapsFillingLock = new Object();
+
     public WebServiceActivityBehavior() {
         itemDefinitionMap.put("http://www.w3.org/2001/XMLSchema:string", new ItemDefinition("http://www.w3.org/2001/XMLSchema:string", new ClassStructureDefinition(String.class)));
     }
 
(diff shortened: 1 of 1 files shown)


tijsrademakers:
Hi,

I understand the idea behind this PR, but adding a synchronized if check is not something that would be nice to have in a web service task. What could be an alternative solution is to cache the definition maps in a cache like it's done for the BpmnModels.

cdeneux:
Hi Tijs,
Can you give me more pointers or entry points about the cache usefor BpmnModels ?
Thanks

cdeneux:
Tijs,

Is it possible to add a new parameter to the method `ActivityBehaviorFactory.createWebServiceActivityBehavior(ServiceTask)` to transmit the current BPMN model at the creation of `WebServiceActivityBehavior`? And so, no more synchronization bloc is needed to load definition maps.

tijsrademakers:
Hi,

Yes we can pass the BPMN model at the creation of the WebServiceActivityBehavior.
But that only removes the need for the BpmnModel bpmnModel = ProcessDefinitionUtil.getBpmnModel(execution.getProcessDefinitionId()); line right?
So that's the line outside the synchronized block, right?

cdeneux:
The definition map loading requires only the BPMN model. And so, if pass the BPMN model at the creation of the `WebServiceActivityBehavior`, we can load definition maps through `fillDefinitionMaps(BpmnModel)` in the constructor. And so the synchronisation bloc is no more needed. The BPMN model is always needed by IO Specification for the execution.
```
   public WebServiceActivityBehavior(BpmnModel bpmnModel) {
        itemDefinitionMap.put("http://www.w3.org/2001/XMLSchema:string", ...);
        fillDefinitionMaps(bpmnModel);
    }

    public void execute(DelegateExecution execution) {

        BpmnModel bpmnModel = ProcessDefinitionUtil.getBpmnModel(execution.getProcessDefinitionId());
        FlowElement flowElement = execution.getCurrentFlowElement();

        ...
```

tijsrademakers:
Ah ok, understood. Let me check if we can easily add the BpmnModel in the WebServiceActivityBehavior constructor and then I'll let you know.

cdeneux:
Ok Tijs, I'll wait your feedback.

tijsrademakers:
Hi, made the change to pass the BpmnModel

cdeneux:
New PR #627 submitted including remarks of @tijsrademakers

PR #642, opened 2017-11-09
Fix vertical and horizontal alignment tools in modeler

dram:
Behavior and icon are mismatch for vertical and horizontal alignment tools in BPMN editor toolbar.


tijsrademakers:
Thanks!

PR #651, opened 2017-11-10
Decision table without a parent deployment, but with the same key, are used as fallback.

stijndepestel:
This addition of code allows deploying DMN decision tables without a parent deployment id, which are still found during the execution of a decision task.

A post on the forum will be created shortly where the benefits and drawbacks of this approach can be discussed.

### DIFF
--- a/modules/flowable-dmn-engine/src/main/java/org/flowable/dmn/engine/impl/cmd/AbstractExecuteDecisionCmd.java
+++ b/modules/flowable-dmn-engine/src/main/java/org/flowable/dmn/engine/impl/cmd/AbstractExecuteDecisionCmd.java
@@ -32,7 +32,7 @@ public abstract class AbstractExecuteDecisionCmd implements Serializable {
     private static final long serialVersionUID = 1L;
 
     protected ExecuteDecisionInfo executeDecisionInfo = new ExecuteDecisionInfo();
-    
+
     public AbstractExecuteDecisionCmd(ExecuteDecisionBuilderImpl decisionBuilder) {
         executeDecisionInfo.setDecisionKey(decisionBuilder.getDecisionKey());
         executeDecisionInfo.setParentDeploymentId(decisionBuilder.getParentDeploymentId());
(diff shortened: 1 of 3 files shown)


stijndepestel:
The forum discussion can be found on the following link: https://forum.flowable.org/t/global-dmn-decisiontable-lookup-at-runtime/1187

tijsrademakers:
Hi Stijn,

I understand the issue with the current parent deployment id approach and I agree there should be a fallback. But this change will not make the execution of a decision task resolve a decision table without a parent deployment id, see this from DmnActivityBehavior:

List<Map<String, Object>> executionResult = ruleService.createExecuteDecisionBuilder()
                        .decisionKey(finaldecisionTableKeyValue)
                        .parentDeploymentId(processDefinition.getDeploymentId())
                        .instanceId(execution.getProcessInstanceId())
                        .executionId(execution.getId())
                        .activityId(task.getId())
                        .variables(execution.getVariables())
                        .tenantId(execution.getTenantId())
                        .execute();

Do you agree more changes are needed?

stijndepestel:
Hi Tijs,

I was indeed mistaken. Still not completely familiar with the code base. I have reverted the first commit and added another one, which should do the trick (at least according to the exploratory testing I performed).

I have added fallbacks in two places in the AbstractExecuteDecisionCmd. They will check for the existence of a decision table if there is no decision table found which was linked to the parent deployment. However, right now, there is no check to make sure that any decision tables found in the fallback won't be linked to another deployment. Adding this, however, seems easy if this is preferred.

I have also removed/commented out some exceptions that were thrown by the DMN DeploymentManager. However, according to my checks, the exceptions defined in the AbstractExecuteDecisionCmd were completely redundant with those in the DeploymentManager and hence nothing should change for the callers of the AbstractExecuteDecisionCmds.
I have also checked if the adapted methods in the DeploymentManager were used elsewhere, but this did not seem to be the case.

Looking forward to some remarks!

Kind regards,

Stijn

tijsrademakers:
Hi Stijn,

Thanks, merged your PR. I enabled the exceptions again, because the methods are also used in other parts of the DMN Engine.

Best regards,

Tijs

Issue #636, opened 2017-11-04
Month always bumps by one in date picker


zwets:
The date picker on any form increases the month by one each time it is opened. 

To test, create a simple process with a start event and user task. Attach a form with just a date field to both the start event and user task. Start the process and enter a date, complete task. When the form opens in the user task, the month has increased by one.

This is on plain out-of-the-box 6.2.0.

tijsrademakers:
Merged PR, thanks!

PR #653, opened 2017-11-12
Fix month off by one in rendered date fields (#636).

zwets:
Found the bug causing #636: in `render-form.js` the `Date` object is populated from a parsed ISO date without taking into account that JavaScript Date has zero-based months.


tijsrademakers:
Thanks!

PR #655, opened 2017-11-13
Add sleep call to resolve ocassional failure in TaskIdentityLinksTest

roberthafner:
when building from maven.


roberthafner:
Closing in favor of another approach (a test clock) I am discussing with Joram.

PR #657, opened 2017-11-14
Throw a FlowableException when a decision task fails.

stijndepestel:
At the moment, when executing a decision task from a BPMN process, the result does not distinguish between an empty response (aka, no rules matched) and an error during the execution. This PR remedies this by throwing a FlowableException when the decision task failed to execute.

A colleague of mine opened up a discussion linked to this on the forum: https://forum.flowable.org/t/decision-task-option-to-throw-error-when-no-rule-matched/1194

Thanks!


tijsrademakers:
Thanks, added a comment to the forum post

stijndepestel:
As discussed on the forum, I have added a flag to a decision task that allows a user to indicate that they require an error if no rules were hit in the decision table.

tijsrademakers:
Thanks, made some small changes + added unit tests.

PR #669, opened 2017-11-26
Transaction not supported proposal

gro-mar:
since 78ba097cc3ac1f6711ce2e0c14ad1fda25c0d71a there is no possibility to run 2 flowable engines in parallel. This bug affects 6.2 release too.
This functionality is crucial for flowable-crystalball (run e.g. org.flowable.crystalball.examples.tutorial.step01.FirstSimulationRunTest).


tijsrademakers:
The schemaCommandConfig needs to be set to contextReusable = false (in case of the Flowable crystalball project), then it works just fine.

PR #677, opened 2017-11-29
DatabaseConfiguration: fix c3po "bad pool size config" warning by set…

PascalSchumacher:
…ting initialPoolSize to minPoolSize

Fixes ugly warn log messages `WARN  com.mchange.v2.resourcepool.BasicResourcePool  - Bad pool size config, start 3 < min 10. Using 10 as start.` when minPoolSize is greater than `3`, e.g. if the user does not configure `minPoolSize`.


PascalSchumacher:
Thanks for merging! 😃

PR #683, opened 2017-11-30
remove unused guava dependency

PascalSchumacher:
`git grep guava` does not show any usage of guava in the codebase.


PascalSchumacher:
Sorry forgot that guava package names do not include `guava`. `git grep google` shows usages of guava in the code base.

PR #686, opened 2017-12-03
travis: use check profile to compile and test more modules

PascalSchumacher:
To increase coverage.

I can replicate the `flowable-dmn-engine` `RuntimeTest#riskRating` test failure with java 7 locally. It is not related to travis.


tijsrademakers:
Thanks for the PR. Fixed the unit test (by changing the output order logic in the DMN engine) and enabled the check profile for Travis in a separate commit. So closing this PR

PR #689, opened 2017-12-03
flowable-http: Downgrade jetty test dependency to a java 7 compatible…

PascalSchumacher:
… version (9.2.14.v20151106) and adjust tests for different HeaderMap behavior in this jetty version.


PascalSchumacher:
Without this change the unit test of `flowable-http` can not be executed on java 7.

I'm not sure if it is correct to add `flowable-http` to the check profile. If that is not the case just discard the commit.

PR #692, opened 2017-12-04
Fallback DMN search

stijndepestel:
Hi Tijs,

After PR #651 you uncommented the exceptions in the deployment manager, which caused the fallback to stop working. Hence I have swallowed the exceptions thrown from the DeploymentManager.

Thanks,

Stijn


tijsrademakers:
Hi Stijn,

Thanks for the PR, Martin was fixing this at the same time :-)
So the fallback is working again now.

Thanks!

PR #702, opened 2017-12-08
Add MyBatis plugin to log sql statement execution time.

roberthafner:


tijsrademakers:
Hi Robert, would it be ok if we change the log level to debug instead of info?

tijsrademakers:
And Joram suggested to make it a configurable option in the engine configuration. Maybe that's even better than checking against the log level.

roberthafner:
Yes, changing the log level to debug is fine for us.

PR #704, opened 2017-12-10
flowable-dmn-engine: skip RuntimeTest#riskRating on java 7, because i…

PascalSchumacher:
…t always fails when this java version is used

Failure happens local and on travis.

Ideally this test should be fixed rather than ignored, but I do not know how to archive this. Imho it is better to ignore it than to make impossible to build flowable on java 7 without ignoring tests.


tijsrademakers:
Fixed the unit test issue with different output order logic in the DMN engine

PascalSchumacher:
Great! Much better than ignoreing the test!

PR #705, opened 2017-12-10
rerun failing tests (up to two times) to make build more stable

PascalSchumacher:
This could help to make the build more stable, as some asynchronous tests fail from time to time.

It only works for Junit 4 tests and only if the failed test does not affect state e.g. the running reused spring context, so I'm not sure how much it will help flowable. Enabling this feature made Apache Camel builds much more stable.


tijsrademakers:
Thanks, interesting option

PR #712, opened 2017-12-14
Draft tests for multiinstance script var test

lsmall:
https://forum.flowable.org/t/multi-instance-collection-syntax-proposal/615/10

My test files for the script task issue I encourntered


lsmall:
Is there a way to push a draft without the Travis Build or should I have just used @Ignore for the failing test?

PascalSchumacher:
As far as I know there is now way to skip the travis build, so using `@Ignore` is best option.

PR #723, opened 2017-12-20
Fix bug in parallel multi-instance behavior.

roberthafner:
The parallel multi-instance behavior sends an unexpected
ACTIVITY_CANCELLED event when a multi-instance user task completes
normally.


roberthafner:
I found an issue with this fix. The sequential multi-instance behavior also sends an ACTIVITY_CANCELLED event for the multi-instance task when it completes normally. I added another test to illustrate the issue. I tried updating the execution to be inactive similar to the parallel multi-instance behavior but the change breaks the recording of the end date for HistoricalActivityInstance.

roberthafner:
The latest update to this PR resolves the issue for both the sequential and parallel multi-instance user task as well as adds new tests to verify the behavior.

PR #724, opened 2017-12-21
Replacement of Json.org with Gson due to problematic license [Dev 6.2.0]

delchev:
Hi,

The Json.org classes used in the flowable-engine have a license, which is considered non-free by the major open source vendors and foundations as well as by the enterprises.

https://wiki.debian.org/qa.debian.org/jsonevil

To be able to use the flowable engine as the BPMN engine in Eclipse Dirigible we have to first fix the issue below:

https://dev.eclipse.org/ipzilla/show_bug.cgi?id=15021

The current pull request contains a variant for a replacement of these problematic classes with the Google's Gson framework.

Regards,
Nedelcho


tijsrademakers:
Thanks for the PR. We have removed the org.json classes and switched it to the Jackson framework. We already use Jackson a lot in the Flowable project so that's why we decided to go for that instead of the Google JSON library. But the PR was very helpful while doing the changes. We'll also remove it from the embedded V5 engine.

delchev:
Great!
Thanks a lot Tijs!

Issue #538, opened 2017-08-19
Add support for CamelBehavior in CDI environment


ruzkant:
Currently the CamelBehavior class only works with a SpringProcessEngineConfiguration. In order to make it work in a CDI environment a new Behavior implementation is needed. This has a knock on effect on the DefaultActivityBehaviorFactory which then needs to construct this class for a CDI processEngine.

I have created an implementation in a branch on my flowable fork:
https://github.com/flowable/flowable-engine/compare/master...ruzkant:camel-cdi

I wasn't sure about whether this should be in the flowable-camel module or in a module of its own, dependency wise. For now I have it in the flowable-camel module and introduced a dependency on flowable-cdi, but this forced me to put a version in as well, not sure if this has something to do with the maven profiles as I am not a maven expert (more used to gradle). Not sure if it should be marked with scope optional.

I am willing to make whatever updates the flowable team envisions necessary to be incorporated.

tijsrademakers:
Thanks for looking into this. I think it's fine to have it as part of the flowable-camel module but the dependency should be set to optional indeed. The main thing that's missing right now are unit tests to show it's working in a CDI environment. Would it be possible for you to add this?

ruzkant:
Yeah, I'll add some unit tests, it was quick for me to test manually with an existing setup, so wanted to first make sure I am barking up the right tree...

ruzkant:
@tijsrademakers It took a while but I finally had time to do some tests and added a pull request for this issue (https://github.com/flowable/flowable-engine/pull/728)

tijsrademakers:
Thanks for the PR @ruzkant !

ruzkant:
@tijsrademakers cool no prob... no PR without your hard work, thx for all you guys :)

PR #728, opened 2017-12-23
Add support for CamelBehavior in CDI environment (Issue #538)

ruzkant:
Note there are some tests that don't pass but it is unrelated to the contributed code. AsyncProcessRevisitedTest fails regularly and MultiInstanceTest has failed on rare occasion

There were 3 options:
1. Add to flowable-camel and add optional CDI dependencies.
2. Add to flowable-cdi and add optional camel dependencies.
3. Create a new flowable-camel-cdi module.

Option 1 was chosen as the test code more naturally fits with flowable-camel, and conceptually CDI is a broader module and best to not depend on more specific module. This has its downsides. Some CDI test infrastructure code was minimally duplicated to avoid dependency on flowable-cdi test jar. Note also that CDI tests were failing when executed in the same process as Spring tests, due to the way ProcessEngines are loaded and accessed on startup. The current solution is to use two separate executions for the maven surefire plugin, one for spring tests and one for CDI tests.


ruzkant:
Linking Issue: #538

ruzkant:
One more comment, I wasn't sure about serialVersionUID for the behaviour implementation classes. I upped them by one as it seems that it had been done before for the default one.

tijsrademakers:
@ruzkant thanks for the PR. A separate flowable-camel-cdi module would be a better fit for the current project modules to my opinion. Would it be possible for you to update the PR with a new module? The serialVersionUIDs don't need to be upped.

Thanks!

ruzkant:
@tijsrademakers I'll do that. The only downside for that option was some tests overlap with the camel module which led me to value proximity for changes to those. But none of the options had no downsides for me, and this is definitely conceptually better.

ruzkant:
@tijsrademakers no worries. I asked initially but as I progressed I thought I should maybe ask again and didn't :)

ruzkant:
@tijsrademakers I have completed the refactoring into a separate module and reverted the serialVersionUID changes

PR #735, opened 2018-01-05
Several minor improvements

marcus-nl:
- AbstractActivityBpmnParseHandler: extract method for creating MultiInstanceActivityBehavior, so that it can be overridden with custom behavior.
- SpringExpressionManager: extract method for creating the Spring EL Resolver, so that it can be overridden in a subclass.
- ExecutionEntityImpl: include name in toString and reduce code duplication.
- flowable-jms-spring-executor/pom.xml: minor fix to indentation.


tijsrademakers:
Thanks, merged the changes manually

Issue #740, opened 2018-01-09
Importing decision table from dmn resulted from exporting decision table that has == and a value with " " results in wrong import


AminaZoheir:
<img width="1381" alt="screen shot 2018-01-09 at 7 50 10 pm" src="https://user-images.githubusercontent.com/3508748/34735058-b2b39a2c-f576-11e7-9015-4439ea324b79.png">

This decision table  results in the following dmn after exporting:

<img width="1376" alt="screen shot 2018-01-09 at 7 56 58 pm" src="https://user-images.githubusercontent.com/3508748/34735224-626fa51e-f577-11e7-8b8d-342eb6a1c7a4.png">

When importing this resulted dmn the decision table looks as follows:

<img width="1381" alt="screen shot 2018-01-09 at 7 49 50 pm" src="https://user-images.githubusercontent.com/3508748/34735142-18741828-f577-11e7-9619-ed02cf878d4c.png">

tijsrademakers:
Merged your PR, thanks for raising the issue and providing the PR!

PR #741, opened 2018-01-09
Issue #740 : Remove check that doesn't add "==" operator to export

AminaZoheir:
Removed check on operator is "==" so that it gets added to the resulted dmn file when exporting.
And therefore fixed the issue with importing if the values contains a " "


tijsrademakers:
Thanks!

Issue #744, opened 2018-01-10
Flowable 6.2.1 OSGI dependency on junit.framework.


mperestrelo:
The org.flowable.idm.engine in 6.2.1 contains a dependency on junit.framework.

IN the manifest for org.flowable.idm.engine we find

Import-Package: junit.framework

This should not be necessary.

tijsrademakers:
Thanks for reporting the issue, changed the junit OSGi dependency to optional, like it was done for the BPMN and CMMN engine already

PR #745, opened 2018-01-10
Clean up junit dependency scope

PascalSchumacher:
Inspired by https://github.com/flowable/flowable-engine/issues/744:

Update to maven-bundle-plugin version 3.5.0 which supports optional dependencies

engine/spring/test modules: change JUnit dependency to compile scope and make it optional (let the maven-bundle-plugin generate the optional OSGI resolution)

non-engine/-spring/-test modules: change JUnit dependency to test scope (inherit scope defined in parent pom)

Not ready to merge yet: OSGI tests currently fail on my box (have to investigate).

### DIFF
--- a/modules/flowable-bpmn-converter/pom.xml
+++ b/modules/flowable-bpmn-converter/pom.xml
@@ -136,7 +136,6 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <scope>provided</scope>
     </dependency>
   </dependencies>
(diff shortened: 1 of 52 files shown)


PascalSchumacher:
Updating the bundle-plugin changes the generated OSGI manifest and that makes the OSGI tests fail, because the bundles do not seem to work. My knowledge of OSGI and the bundle plugin is minimal, so that I'm not able to fix this with reasonable effort.

Issue #719, opened 2017-12-18
flowable-modeler: changing business process model name and key does not change process name and identifier


PascalSchumacher:
When you create a business process model the name and the key are used as the process name and identifier.

If you later change the business procees model name and key this is not protagated to the process name and identifier.

This behavior is confusing our users, because they do not see their new models as deployed processes.

tijsrademakers:
Merged the PR and this should fix the issue. Thanks for the PR @seifeldeen92 !

PascalSchumacher:
@tijsrademakers I guess you did not push yet (I do not see the PR merged yet.)?

PascalSchumacher:
@tijsrademakers Sorry I missed https://github.com/flowable/flowable-engine/commit/2fa4b45440f8e1c0c6fa5c0012879686288a7520 somehow.

PR #746, opened 2018-01-11
Issue#719: change bpmn name and key does not change process name and identifier

seifeldeen92:
as in issue #719


tijsrademakers:
Thanks for PR, made a small change to use the key directly for the process_id attribute value

PascalSchumacher:
Thanks for fixing this! 👍

PR #755, opened 2018-01-16
fixing process model httpRequest timeout precedence before engineConf

martin-grofcik:
https://forum.flowable.org/t/configuring-httpclient-timeouts-like-socketimeout-in-flowable-rest-app/1447

### DIFF
--- a/modules/flowable-http/src/main/java/org/flowable/http/HttpActivityExecutor.java
+++ b/modules/flowable-http/src/main/java/org/flowable/http/HttpActivityExecutor.java
@@ -291,7 +291,7 @@ public class HttpActivityExecutor {
                 .setRedirectsEnabled(!requestInfo.isNoRedirects())
                 .setSocketTimeout(socketTimeout)
                 .setConnectTimeout(connectTimeout)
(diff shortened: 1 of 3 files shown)


tijsrademakers:
Merged manually, thanks

ruzkant:
@martin-grofcik @tijsrademakers 

The connectionRequestTimeout is the timeout httpclient uses for getting a connection when there is a connection pool. So it does not quite correspond to a timeout for the request processing.

Wouldn't setting socketimeout be more appropriate here? 

To be honest it is annoying to always specify two values, but I think the RequestTimeout field has no direct correspondence to the underlying model of TCP socket comms and therefore overriding connect and socket timeout separately may be even more appropriate than a single request timeout.

ruzkant:
Link to http client doc on connectionRequestTimeout:
[https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/client/config/RequestConfig.html#getConnectionRequestTimeout()](https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/client/config/RequestConfig.html#getConnectionRequestTimeout())

PR #758, opened 2018-01-17
Fix - DataObjects not being returned from nested subprocesses.

roberthafner:


tijsrademakers:
Thanks Robert, good catch

PR #762, opened 2018-01-18
Allow remember me cookie's domain to be set from application properties

KaneFreeman:
Resolves the issue described in https://forum.flowable.org/t/different-sub-domain-name-usage-per-ui-app/1308


tijsrademakers:
Thanks for the PR

liguobao:
please support env config on Docker

Issue #771, opened 2018-01-30
flowable-modeler: duplicating a business process with deferent name and key does not change the duplicated process name and identifier


seifeldeen92:
When you duplicate a business process model the name and the key are not used as the process name and identifier. instead, it uses the original process name and identifier

This behavior is confusing because users do not see their new model as deployed processes.

PascalSchumacher:
I guess this can be closed (as the pull request has been merged)?

tijsrademakers:
Merged the PR, thanks!

PR #773, opened 2018-01-30
Issue#771: Duplicating bpmn name and key does not change duplicated process name and identifier

seifeldeen92:
as in #771


tijsrademakers:
Thanks for the PR!

Issue #772, opened 2018-01-30
A strange question in Countersign 


jianlinz:
a Countersign have  ten numberOfInstances
I use a executionListener event="end"  to see nrOfCompletedInstances 
when countersign end in activiti6.* the nrOfCompletedInstances is 10
but flowable 6.2.1 the nrOfCompletedInstances  is null

tijsrademakers:
Hi, in Flowable there's a parent execution for a multi instance child execution that has the nrOfCompletedInstances variable with the value.

PR #775, opened 2018-01-31
https://github.com/flowable/flowable-engine/issues/772  unittest 

jianlinz:
A strange question in Countersign


tijsrademakers:
Thanks for providing the unit test, used it to reproduce the issue and fix it. The unit test (slightly modified) is now part of the MultiInstanceTest unit test class.

PR #790, opened 2018-02-12
Debugger 

martin-grofcik:


modules/flowable-ui-task/flowable-ui-task-rest/src/main/java/org/flowable/app/rest/runtime/RuntimeDisplayJsonClientResource.java
```diff
@@ -128,7 +128,7 @@ public JsonNode getModelJSON(@PathVariable String processInstanceId) {
 
         ProcessInstance processInstance = runtimeService.createProcessInstanceQuery().processInstanceId(processInstanceId).singleResult();
         if (processInstance == null) {
-            throw new BadRequestException("No process instance found with id " + processInstanceId);
+            return getModelHistoryJSON(processInstanceId);
```
tijsrademakers: Hi Martin, this change modifies the behaviour of the getModelJson method. If you need the history json then the history endpoint should be invoked.

modules/flowable-ui-task/flowable-ui-task-app/src/main/resources/META-INF/flowable-app/flowable-app.properties
```diff
@@ -96,7 +96,7 @@ idm.admin.password=test
 #app.redirect.url.on.authsuccess=http://localhost:8080/flowable-task
 #deployment.api.url=http://localhost:8080/flowable-task/process-api
 
-debugger.enabled=false
+debugger.enabled=true
```
tijsrademakers: Can you change this to false as the default?

martin-grofcik: Don't we want to allow debugging by default? as a feature to show.

modules/flowable-ui-task/flowable-ui-task-rest/src/main/java/org/flowable/app/rest/runtime/RuntimeDisplayJsonClientResource.java
martin-grofcik: Done

martin-grofcik:
@tijsrademakers: Please let me know the code review results.

tijsrademakers:
@martin-grofcik two comments:

1.  Can you revert the following change?

-    protected List<AsyncRunnableExecutionExceptionHandler> customAsyncRunnableExecutionExceptionHandlers;
+    protected List<AsyncRunnableExecutionExceptionHandler> customAsyncRunnableExecutionExceptionHandlers = Collections.EMPTY_LIST;

2. In the Flowable task app UI grid is already part of the included libraries. UI Grid is the sucessor / new version of NG grid, so can use UI grid instead and remove the addition of the NG grid libraries?

PR #792, opened 2018-02-12
Custom Authentication in IDM

ghonia:
The goal of this Pull Request is to allow users/forkers of flowable to customize the authentication process of the flowable-ui-*-apps. For example, at seal-software, we need our users to login to instances of Flowable apps (modeler/task/admin/rest/idm) using their Seal credentials without integrating with LDAP. To do so, we needed to override two thing in the default Flowable source code.
* IdmIdentityServiceImpl, and
* dbAuthenticationProvider

To allow any user to do the same without having to modify Flowable source code, this pull request did the following

Added a name to the IdmIdentitiService Bean created in org.flowable.app.conf.FlowableIdmEngineConfiguration called “defaultIdmIdentityService” and used it in the qualifier whenever IdmIdentityService was required in the IDM.

Added Autowire for IdmIdentityService with qualifier “customIdmIdentityService” in org.flowable.app.conf.FlowableIdmEngineConfiguration. FlowableIdmEngineConfiguration will return the custom Identity Service if it was provided on the class path. Otherwise, the default flowable Identity Service will be returned. Now anyone can override the default Identity service by adding their implementation on the CP.

Add Autowire for an AuthenticationProvider in org.flowable.app.conf.SecurityConfiguration. If not found, the configuration will fall back to the default dbAuthenticationProvider.

Finally, we added option to disable bootstrapping the IDM by setting property idm.bootstram.enabled=false in flowable-ui-app.properties. If not found, the value defaults to true

Example of custom authentication Beans provided in WEB-INF/lib of flowable-idm after building flowable-idm OOTB

package org.flowable.idm.extension.bean;
@Component("customIdmIdentityService")
public class SealIdentityServiceImpl extends IdmIdentityServiceImpl {
…

package org.flowable.idm.extension.bean;
@Component("customAuthenticationProvider")
public class SealAuthenticationProvider implements AuthenticationProvider {


tijsrademakers:
Hi,

Thanks for the PR, I think it's good to make the idm logic more pluggable.
About the implementation details. When there is a customIdentityService, then it would also be possible to set this as the IdmIdentityService implement on the IdmEngineConfiguration. Then the command executor is automatically set to this IdmIdentityService implementation. Would that work for you as well?

ghonia:
Hi, @tijsrademakers 
Thanks for the suggestion. Works even better!

PR #805, opened 2018-02-14
flowable-form-rest: remove test dependencies, because there are no tests

PascalSchumacher:


tijsrademakers:
Your are right for wanting to remove the test dependencies, but we should definitely have a number of unit tests in this module. Will make sure they are added.

PascalSchumacher:
That is an even better solution! 👍

Issue #800, opened 2018-02-14
Feature Request: Add the name of the variables updated by decision table in a process variable


AminaZoheir:
This could be used for building more generic processes that do not necessarily need to be updated with every decision table update.

PR #807, opened 2018-02-15
Issue #800 Add updatedVariablesList in decision table execution

AminaZoheir:


tijsrademakers:
Thanks for the PR. Having a list of updated variables as a process variable is something that can be handy for certain use cases, but we generally don't advise to store non primitive variables as process variables. We could however extend the DMN task with a DmnResponseHandler (or something similar) so that additional variables can be created when necessary. What do you think?

AminaZoheir:
@tijsrademakers  I agree that it should be done as part of the DMN task. Could you please elaborate more on the DmnResponseHandler idea?

AminaZoheir:
I'll close this pull request and will implement the DmnResponseHandler in a different branch

PR #813, opened 2018-02-17
Bring back commons-fileupload in apps that need it

filiphr:
In #801 @PascalSchumacher removed the `commons-fileupload` from all `ui` modules. However, we need that in some of the modules (the ones that use the `CommonsMultipartResolver`)

I am bringing back that dependency only in the modules that actually use that Multipart resolver

### DIFF
--- a/modules/flowable-ui-admin/flowable-ui-admin-conf/pom.xml
+++ b/modules/flowable-ui-admin/flowable-ui-admin-conf/pom.xml
@@ -82,6 +82,10 @@
         </dependency>
         <!-- COMMON dependencies -->
         <dependency>
+            <groupId>commons-fileupload</groupId>
+            <artifactId>commons-fileupload</artifactId>
+        </dependency>
+        <dependency>
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
         </dependency>
--- a/modules/flowable-ui-modeler/flowable-ui-modeler-conf/pom.xml
+++ b/modules/flowable-ui-modeler/flowable-ui-modeler-conf/pom.xml
@@ -82,6 +82,10 @@
         </dependency>
         <!-- COMMON dependencies -->
         <dependency>
+            <groupId>commons-fileupload</groupId>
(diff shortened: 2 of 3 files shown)


PascalSchumacher:
Hi Filip,

the only usages of `CommonsMultipartResolver` I see are by the
 `flowable-ui-*-rest` modules, which still have the `commons-fileupload` dependency. So everything should be fine. What am I missing?

filiphr:
The 3 modules are using `CommonsMultipartResolver` from spring which uses `commons-fileupload`. However, it is an optional dependency in spring so we must pull it in. See

* [ui-task Dispatcher Configuration](https://github.com/flowable/flowable-engine/blob/master/modules/flowable-ui-task/flowable-ui-task-conf/src/main/java/org/flowable/app/servlet/AppDispatcherServletConfiguration.java#L64)
* [ui-admin Dispatcher Configuration](https://github.com/flowable/flowable-engine/blob/master/modules/flowable-ui-admin/flowable-ui-admin-conf/src/main/java/org/flowable/admin/conf/DispatcherServletConfiguration.java#L105)
* [ui-modeler Dispatcher Configuration](https://github.com/flowable/flowable-engine/blob/master/modules/flowable-ui-modeler/flowable-ui-modeler-conf/src/main/java/org/flowable/app/servlet/AppDispatcherServletConfiguration.java#L64)

If the dependencies are not there there would be `ClassNotFoundException` (that's how I noticed it 😄).

@PascalSchumacher maybe you don't have all the modules activated (I had the same issue with something else 😄)

PascalSchumacher:
I did not know that. Thanks for explaining! 👍 

And sorry for breaking this!

filiphr:
No problem really 😄

tijsrademakers:
Before applying the PR from Pascal I also did a quick search and didn't find it being used. But it looks like org.springframework.web.multipart.commons.CommonsMultipartResolve needs commons-fileupload. Thanks for looking at this Filip!

Issue #825, opened 2018-02-22
why SecurityUtils is final


jianlinz:
Hi!
I have a question. why SecurityUtils is final? How can I cover getCurrentFlowableAppUser method.

i hope i can inject my impl. if is ok,i will submit a PR

tijsrademakers:
Hi,

There's no good reason, I think it should not be final.

Best regards,

Tijs

jianlinz:
Hi!
I submit a PR https://github.com/flowable/flowable-engine/pull/827
add a SecurityService  have a default impl SecurityServiceImpl 
when I want cover getCurrentFlowableAppUser method,i will inject  a impl implements SecurityService.
like this
@Service
@Primary
public class PEPSecurityServiceImpl implements SecurityService

PR #827, opened 2018-02-23
https://github.com/flowable/flowable-engine/issues/825

jianlinz:
a PR with issue825

### DIFF
--- a/modules/flowable-ui-common/src/main/java/org/flowable/app/security/SecurityUtils.java
+++ b/modules/flowable-ui-common/src/main/java/org/flowable/app/security/SecurityUtils.java
@@ -12,6 +12,7 @@
  */
 package org.flowable.app.security;
 
+import org.flowable.app.service.security.SecurityService;
 import org.flowable.idm.api.User;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.context.SecurityContext;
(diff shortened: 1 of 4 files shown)


tijsrademakers:
Thanks for the PR, I've removed final from the SecurityUtils class. The other changes in the PR are not applied. Maybe another solution that doesn't require direct Spring application context access would be better.

ruzkant:
@jianlinz If you are using Spring Security, have you considered a custom AuthenticationProvider for spring, so that the user principal is an instance of FlowableAppUser? I have used this successfully, and basically just wrapped the existing one, like this:

public class CustomAuthenticationProvider implements AuthenticationProvider {
    protected AuthenticationProvider wrappedAuthenticationProvider;
    
    public CustomAuthenticationProvider(AuthenticationProvider wrappedAuthenticationProvider) {
        super();
        this.wrappedAuthenticationProvider = wrappedAuthenticationProvider;
    }

    @Override
    public boolean supports(Class< ? > authentication) {
        return CustomAuthentication.class.isAssignableFrom(authentication);
    }
    
    @Override
    public Authentication authenticate(Authentication authentication) throws AuthenticationException {
        Authentication wrappedAuthResult = wrappedAuthenticationProvider.authenticate(authentication);
        CustomPrincipal customPrincipal = wrappedAuthResult.getPrincipal();
        FlowableAppUser appUser map(customerPrincipal);
        CustomAuthentication auth = CustomAuthentication(appUser);
        return auth;
    }
    
    protected FlowableAppUser map(CustomPrincipal principal) {
        // map here
    }
}


Details will depend on whether you have a custom Authentication or using one of the Spring developed implementations.

PR #830, opened 2018-02-23
Add capability to customize the WSDL importer of flowable-cxf

cdeneux:
When using web-service task in my process definition, it is needed to import WSDL of invoked web-services.

I need to use a custom URI resolver to find WSDLs of my services. So I should be able to customize `flowable-cxf` with my own URI resolver through a custom WSDL importer.

Please find here my contribution to be able to configure the WSDL importer at process engine configuration.


tijsrademakers:
Thanks for the PR

cdeneux:
Thanks Tijs to have accepted my contribution !

PR #834, opened 2018-02-25
Use java.util.Base64 instead of org.apache.commons.codec.binary.Base6…

PascalSchumacher:
…4/org.springframework.security.crypto.codec.Base64.

Remove commons.codec dependency of modules with do not use it anymore.


tijsrademakers:
Thanks!

PR #838, opened 2018-02-26
Replace catch-exception with assertj and update mockito to latest from 1.x

filiphr:
catch-exception is not compatible with latest 1.10.19 from mockito

This also adds AssertJ which would allow us to write even better tests ;).


PascalSchumacher:
:+1: :smiley:

PR #845, opened 2018-03-02
Add IntelliJ idea code style settings and templates

filiphr:
The IntelliJ code style was created by importing the Eclipse formatter.

The following was changed to be in balance with the Eclipse defaults:
* Order of imports (defined by looking into couple of new classes and their formatting)
* Never use wildcard imports (default with Eclipse)
* Parameter description on new line

The file templates have been done by looking into the eclipse file templates and generating similar ones for IntelliJ.

The addition of the AssertJ is just my own preference of an assert framework (even JUnit uses AssertJ 😉). The default from IntelliJ are the JUnit assertions.

I have no particular preference for the settings, I just want to have consistency, and that is the idea of this PR 😄


filiphr:
One preference that I would like to add is chained method calls. For example instead of having:

```
Map<String, Object> result = ruleService.createExecuteDecisionBuilder().decisionKey("simple").variables(variables).executeWithSingleResult();
```
To have:

```
Map<String, Object> result = ruleService.createExecuteDecisionBuilder()
                        .decisionKey("simple")
                        .variables(variables)
                        .executeWithSingleResult();
```

filiphr:
It seems I have more than one preference :). I am not entirely sure for the `CONTINUATION_INDENT_SIZE` of `16` it seems a bit too big. 

The same example from before with `8` would look like:

```
Map<String, Object> result = ruleService.createExecuteDecisionBuilder()
                .decisionKey("simple_decision")
                .variables(variables)
                .executeWithSingleResult();
```

PR #846, opened 2018-03-02
OAuth2

ghonia:
Adding oAuth2 Support

To Enable oAuth, you need to add an implementation of OAuthAuthorizationRequestor and OAuthAuthorizationHandler on the spring context (foe example under org.flowable.idm.extension.bean). The org.flowable.app.oauth.OAuthController will pick them up. The Login page will display the list of requestors under the username/password login form.

[Archive.zip](https://github.com/flowable/flowable-engine/files/1775727/Archive.zip) is an example implementation of one Requestor and one Handler to login to "DocuSign"

The result looks like this
screen shot 2018-03-02 at 4 37 32 pm


ghonia:
@tijsrademakers,
Do you have any thoughts about this?

tijsrademakers:
@ghonia, sorry for the delay. We need to give a test drive and then the PR can be merged. I expect we can test it in the next days.

tijsrademakers:
@ghonia a couple of questions that I have from the PR:

- StartProcessInstanceWithFormCmd 
  formVariables.put("FLOWABLE_REMEMBER_ME", variables.get("FLOWABLE_REMEMBER_ME")); 
  For which purpose is this needed?

-  values.put(CookieConstants.COOKIE_NAME, cookie);
+        startRequest.setValues(values);

   For which purpose is this needed?

Thanks!

filiphr:
@ghonia I am in the process of moving the current apps that flowable has to use Spring Boot. We might even migrate them to Boot 2 (which brings Spring Security 5). Do you think that this might require a different way of implementing the OAuth support?

ghonia:
@tijsrademakers 

After successfully authenticating with one oAuth provider, OAuthController.java will save the returned OAuth token in the "data" field of the flowable Token. Now, the logged in user needs to execute a process which needs access to this access_token. So, /process-instances end point was modified to always append the cookie (if not null) to the variables of StartProcessInstanceWithFormCmd.

Now, a delegate inside a process that was initiated by a user who authenticated using oAuth2 can find the cookie id by using execution.getVariable("FLOWABLE_REMEMEBR_ME") and decode it just like FlowableCookieFilter does.

You will also find that RemoteToken has been modified to expose the "data" field

ghonia:
@tijsrademakers

I needed this because trying SecurityUtils and the TokenService from inside the delegate didn't work. Do you have any better suggestions on where to store the oAuth2 access-token and how to expose it to delegates?

ghonia:
@filiphr 

Yes, using Spring 5, then some of this PR will be redundant

As you of course know Spring Security 5 has built in support for oAuth2. If we move to Security 5, we can comment out the OAuthController.java (and it's supporting two interfaces OAuthAuthorizationRequestor.java and OAuthAuthorizationHandler.java) and use the Spring built in httpSecurity.oauth2Login():

tijsrademakers:
@ghonia we have now merged the Spring Boot 2 (and Spring Security 5 with it) to master. So all the UI apps are now based on Spring Boot 2. Could you have a look at your PR and see if you could modify it to Spring Boot 2 and Spring Security 5? Sorry for the inconvenience.

PR #854, opened 2018-03-03
Add auto configuration for the Idm engine

filiphr:
Moves the creation of the Idm engine into a separate auto configuration

### DIFF
--- a/modules/flowable-idm-spring/src/test/java/org/flowable/spring/test/authentication/SpringPasswordEncoderTest.java
+++ b/modules/flowable-idm-spring/src/test/java/org/flowable/spring/test/authentication/SpringPasswordEncoderTest.java
@@ -14,6 +14,7 @@ package org.flowable.spring.test.authentication;
 
 import static junit.framework.TestCase.assertNotNull;
 import static junit.framework.TestCase.assertTrue;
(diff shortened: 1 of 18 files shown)


modules/flowable-idm-spring/src/test/java/org/flowable/spring/test/authentication/SpringPasswordEncoderTest.java
```diff
@@ -71,8 +71,10 @@ public void testValidateSpringPasswordEncoder() {
 
         IdmEngineConfiguration defaultIdmEngineConfiguration = IdmEngines.getDefaultIdmEngine().getIdmEngineConfiguration();
 
-        assertTrue(defaultIdmEngineConfiguration.getPasswordEncoder() instanceof ClearTextPasswordEncoder);
-        assertTrue(passwordEncoder instanceof SpringEncoder);
+        assertThat(defaultIdmEngineConfiguration.getPasswordEncoder())
```
filiphr: I would say that this test was wrong before. When creating the `IdmEngine` with spring we should be getting the one which is used in spring when doing `IdmEngines.getDefaultIdmEngine()` and not create a new one

PR #855, opened 2018-03-04
Add auto configuration for the CMMN engine

filiphr:


modules/flowable-spring-boot/flowable-spring-boot-starters/flowable-spring-boot-starter-basic/src/test/java/org/flowable/test/spring/boot/cmmn/CmmnEngineAutoConfigurationTest.java
```diff
@@ -0,0 +1,134 @@
...
+        });
+    }
+
+    private void assertAutoDeployment(AnnotationConfigApplicationContext context) {
+        CmmnRepositoryService repositoryService = context.getBean(CmmnRepositoryService.class);
+
+        List<CaseDefinition> caseDefinitions = repositoryService.createCaseDefinitionQuery().orderByCaseDefinitionName().asc().list();
+        assertThat(caseDefinitions)
+            .extracting(CaseDefinition::getKey)
+            .containsExactly("case1", "case2", "case3", "case4");
```
filiphr: This is the reason why this fails. For some reason the order is different on Travis. I can change the keys, but it is strange why Travis behaves like this. When I run the test locally it is green

PR #881, opened 2018-03-13
instant time support for timer

martin-grofcik:


tijsrademakers:
Hi Martin,

Could you also add a unit test for a timer start event that uses an expression with a bean that gives back an Instant, and a boundary timer event with an Instant variable?

martin-grofcik:
Java 8 tests pass locally.

PR #887, opened 2018-03-14
Cmmn user event listener

DennisFederico:
The CmmnRuntimeService part of UserEventListener


modules/flowable-cmmn-converter/src/main/java/org/flowable/cmmn/converter/CmmnXmlConverter.java
```diff
@@ -18,26 +18,7 @@
...
-import org.flowable.cmmn.model.PlanFragment;
-import org.flowable.cmmn.model.PlanItem;
-import org.flowable.cmmn.model.PlanItemDefinition;
-import org.flowable.cmmn.model.ProcessTask;
-import org.flowable.cmmn.model.Sentry;
-import org.flowable.cmmn.model.SentryOnPart;
-import org.flowable.cmmn.model.Stage;
-import org.flowable.cmmn.model.Task;
-import org.flowable.cmmn.model.TimerEventListener;
+import org.flowable.cmmn.model.*;
```
tijsrademakers: Hi Dennis, could you do an import of all the classes instead of the .* import. We don't use the import .* in the Flowable source code

modules/flowable-cmmn-converter/src/main/java/org/flowable/cmmn/converter/export/PlanItemDefinitionExport.java
```diff
@@ -13,16 +13,7 @@
...
-import org.flowable.cmmn.model.DecisionTask;
-import org.flowable.cmmn.model.HumanTask;
-import org.flowable.cmmn.model.Milestone;
-import org.flowable.cmmn.model.PlanItemDefinition;
-import org.flowable.cmmn.model.ProcessTask;
-import org.flowable.cmmn.model.ServiceTask;
-import org.flowable.cmmn.model.Stage;
-import org.flowable.cmmn.model.Task;
-import org.flowable.cmmn.model.TimerEventListener;
+import org.flowable.cmmn.model.*;
```
tijsrademakers: same here

modules/flowable-cmmn-converter/src/main/java/org/flowable/cmmn/converter/CmmnXmlConverter.java
DennisFederico: Sure... will do

PR #895, opened 2018-03-16
Bootiful rest app

filiphr:
As the title says migration of the Flowable Rest Application into an executable Spring Boot war.


modules/flowable-app-rest/src/main/java/org/flowable/rest/conf/DatabaseConfiguration.java
```diff
@@ -1,157 +0,0 @@
...
-            if (maxPoolSize != null) {
-                dataSource.setMaximumPoolSize(maxPoolSize);
-            }
-
-            // Time in milliseconds to indicate when a connection will be removed from the pool. 
-            // Use 0 to never remove an idle connection from the pool 
-            Long idleTimeout = environment.getProperty("datasource.max-idle-time", Long.class);
-            if (idleTimeout != null) {
-                // Property has been historically expressed in seconds, but Hikari expects milliseconds
-                idleTimeout = idleTimeout * 1000;
```
filiphr: This confused me a bit.  Why multiple the new value with 1000?

modules/flowable-app-rest/src/main/resources/static/docs/swagger-ui.js
```diff
@@ -21457,169 +21457,169 @@ function request(RequestConstructor, method, url) {
 module.exports = request;
 
 },{}],162:[function(require,module,exports){
-
```
filiphr: This are some new line changes which I can't seem to revert due to the setup of my repo. If someone can revert them I would have no issue

PR #896, opened 2018-03-16
Add support for a triggerable service task.

roberthafner:


roberthafner:
Joram, Here is a second revision of this feature. I'm still tracking down a failure in a test. But I thought I would share now to let you take a peak. For now, I named the new attribute 'triggerable'.

roberthafner:
Joram, 

I resolved the test failure. The latest changes include the updates we discussed via email. Please review. 

thanks,
Rob

tijsrademakers:
Hi Rob,

Thanks for the PR. I've made a couple of changes with the variable handling (no support for transient variables) and a separate job handler type. Hope these changes are ok with you.

PR #904, opened 2018-03-21
Support expressions for OptionsFormField 

ruzkant:
So basically I added an optionsExpression field to the OptionsFormField class and updated a few command values to process the expression. Note the expression must evaluate to a List<Option> as that is the field on OptionsFormField, or if a string it must evaluate to the equivalent json.

I also updated the modeler app to support setting an expression instead of a statically defined list of options. 

I just used google translate for the non-english language bundles and have no idea how good it is.

The task app required a minor alteration to support using id field as the value for radio buttons, which was hardcoded to only use the option.name value.

I also changed the tomcat port in the task app pom for postgres to be the same as h2 and mysql which avoids port clash with idm app when running the apps with mvn tomcat7:run.


ruzkant:
Looks like one of the travis tests failed, but there were some broken pipe socket exceptions, so does not look related to the pull request...

```
Caused by: java.net.SocketException: Broken pipe (Write failed)
	at java.net.SocketOutputStream.socketWrite0(Native Method)

```

PascalSchumacher:
Yes, that looks like an unrelated failure which occurs from time to time. 😞

ruzkant:
The tests run fine on my system if all I do is add a small delay to binary calls:

ruzkant:
The tests run fine on my side if I add a small delay for the binary requests where it tends to fail.

I was trying to step through the code in eclipse and found that the issues does not occur when I do that, which is why I added a little delay. Not great, but maybe at least a temporary solution if it works in Travis...

public CloseableHttpResponse executeBinaryRequest(HttpUriRequest request, int expectedStatusCode) {
        try {
            Thread.sleep(1000);
        } catch (InterruptedException e) {
            e.printStackTrace();
        }
        return internalExecuteRequest(request, expectedStatusCode, false);
    }

ruzkant:
@PascalSchumacher  I added a delay as mentioned in a previous comment. While I seriously dislike this type of solution it is test code and I have to admit by the time I discovered it I ran out of steam for looking into finding the real cause.

ruzkant:
Uploading a simple test app for this:
[Options Expression Test.zip](https://github.com/flowable/flowable-engine/files/1845427/Options.Expression.Test.zip)

tijsrademakers:
Thanks for the PR and an additional thanks for the example app, that helps a lot to test it easily. Nice usage of the data objects btw.

PR #909, opened 2018-03-23
Fix-RTL-labeling-problem

s-noroozi-cs:
In modeler application, setting English character in name attribute for BPMN elements such as userTask, serviceTask,etc , all things are good. But, When using right-to-left alignment languages such as Persian (also know as Farsi), Arabic, Hebrew, showing output in UI were un-readble, because align is left-to-right and the alphabets were distinct from together. Changeling "letter-spacing" attribute from negative value to zero solved the problem.


s-noroozi-cs:
close due to the fix-rtl-labeling-problem #910, changes should be apply on "oryx-editor" source code

PR #914, opened 2018-03-25
Add modeler app multi tenancy support

ruzkant:
This pull request adds tenant_id to modeler database tables and provides a TenantProvider interface through which the application access and applies tenancy.

The default tenant provider will look up the tenant id from the Environment (eg. app's properties file) and if not found uses the user's tenant_id (this was newly added by @martin-grofcik and is not yet supported in idm app) and uses null as last resort. This should be backwards compatible with existing installations which would neither have the tenant_id configuration property nor the user tenant_id. Also once tenant_id is implemented on users, the configuration property allows overriding this.

The mybatis xml files were updated to use spaces, as there was a mix of spaces and tabs in the file. Let me know if this is not ok.

I preferred adding tenant_id to both model and model_history as they are often queried separately and this avoids joining, but left it out from model_relation as joins are typically by id. In some situations like using Postgres RLS enforcement of tenancy it is simpler and safer to just add tenant_id to all tables, but I noticed this was not the general approach with other tables.

### DIFF
--- a/modules/flowable-rest/src/test/java/org/flowable/rest/service/BaseSpringRestTestCase.java
+++ b/modules/flowable-rest/src/test/java/org/flowable/rest/service/BaseSpringRestTestCase.java
@@ -273,6 +273,10 @@ public class BaseSpringRestTestCase {
      * IMPORTANT: calling method is responsible for calling close() on returned {@link HttpResponse} to free the connection.
      */
     public CloseableHttpResponse executeBinaryRequest(HttpUriRequest request, int expectedStatusCode) {
+        try {
+            Thread.sleep(1000);
+        } catch (InterruptedException e) {
+        }
         return internalExecuteRequest(request, expectedStatusCode, false);
     }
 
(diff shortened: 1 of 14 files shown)


ruzkant:
Added a delay to BaseSpringRestTestCase.executeBinaryRequest method to resolve failing tests. For further comments see this pull request where the same fix was applied: https://github.com/flowable/flowable-engine/pull/904

tijsrademakers:
Thanks for the PR, I've modified the publish logic a bit so that the deploy uri is not changed with the tenant id prepended to the deploy url. Instead we could maybe use a deploy url property that contains the tenant id value?

ruzkant:
@tijsrademakers Oh, my apologies, I didn't mean to leave the tenant id prepended as well after adding the parameter, that's a bug. Thanks for picking it up and fixing it :)

My first implementation was for myself and it used the subdomain as tenant id and it was reworked to be generally useful as is for the current system without requiring that. So yes that could be statically done like you described, although in my case I was using a different TenantProvider implementation to use the subdomain.

I would also like to make the admin app restrict to a specific tenant, but that is a lot more work as the UI provides inputs for tenant and there are way more api's called that would need tenant_id specified. If you have ideas of how to do this more easily than touching large parts of the code base I'd appreciate it :)

PR #954, opened 2018-04-04
Update raphael.js from 2.1.0 to 2.2.1

alagane:
Use the last minified version found here: https://dmitrybaranovskiy.github.io/raphael/
And the direct link:
https://raw.githubusercontent.com/DmitryBaranovskiy/raphael/master/raphael.min.js
Reduce size from 291K to 91K


tijsrademakers:
Thanks for the PR, merged the PR manually and also applied the change to the other UI apps using Raphael JS.

PR #960, opened 2018-04-05
Task builder

martin-grofcik:


tijsrademakers:
@martin-grofcik can you have a look at the parentTaskEntity.forceUpdate(); line? It should not be necessary to call forceUpdate. 
Can you also verify if it's necessary to add the flowable-task-service-api dependency to flowable common? Don't see anything in this PR that would require this.

PR #962, opened 2018-04-05
Make sure that the old property sources are loaded before the spring boot application configuration

### LINKED ISSUE #961: flowable-ui-app.properties are loaded after the application.properties
The Spring Boot default application.properties are with a higher order than our backwards compatible flowable-ui-app.properties. This should not be the case

filiphr:
Fixes #961

### DIFF
--- a/modules/flowable-app-rest/pom.xml
+++ b/modules/flowable-app-rest/pom.xml
@@ -606,5 +606,10 @@
         <groupId>javax.xml.bind</groupId> <!-- Needed for swagger doc generation -->
         <artifactId>jaxb-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-test</artifactId>
+      <scope>test</scope>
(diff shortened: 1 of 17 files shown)


PascalSchumacher:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.6.1:testCompile (default-testCompile) on project flowable-ui-common: Compilation failure: Compilation failure: 
[ERROR] /home/travis/build/flowable/flowable-engine/modules/flowable-ui-common/src/test/java/org/flowable/app/properties/BackwardsCompatiblePropertiesLoaderTest.java:[15,35] package org.assertj.core.api does not exist
```
I guess another module needs an assertj-core dependency?  😉 😃

filiphr:
Thanks for pointing it out. I committed that by mistake. The `FlowableIdmApplicatiinTest` is actually that tests this part. I'll remove it tomorrow

filiphr:
I've fixed the issues (I think). There was another place where a dependency was missing.

PR #969, opened 2018-04-06
Dockerfile maven plugin

jacarui:
`docker-maven-plugin` is descoped and Spotify recommends moving to Dockerfiles using `dockerfile-maven-plugin`. I changed it for image creation only, creating new Dockerfiles and pointing docker scripts to new files.

Pending: move image publication to `dockerfile-maven-plugin`.


yvoswillens:
@jacarui thanks for the work on the docker maven plugin changes.

couple of questions / remarks;

* Is there a reason to move the Docker related files from src/main/docker to /docker? otherwise I would prefer to leave them there. (including the new Dockerfile in the app module's root)
* Also for the app.yml to docker-compose.yml rename
* In the plugin configuration there is a <WAR_FILE> declaration. As far as I can see this is not used (f.e. in the Dockerfile). Is this on purpose?

jacarui:
Some answers:
- It's a style issue. I'm used to build images from different stacks (java/javascript) and for me Docker is an agnostic tool which should remain in a different level and don't follow Java conventions. I thought this way of organising files was the popular convention for docker files because is how projects normally do it, but I can't find any documentation regarding this. Dockerfile remains at pom level as recommended on new maven plugin to use other Docker tools that rely on Dockerfiles.
- 'docker-compose.yml' is the standard name so you can run "docker-compose up" without specifying file name ("docker-compose up -t app.yml"). Also is more descriptive.
- You are right, I'm not using that argument. I can remove it or use it instead of hardcoding war name in the Dockerfile.
- I already updated those. Did I forget some path?

yvoswillens:
* Style is something that is hard to debate ;) But if there are no tool specific reasons to move it I would prefer to leave them in /src/main/docker for now (as this a Java / Maven project). I read in several issues on GitHub that more people want to specify a custom location for the Dockerfile. And also that currently this can cause issues with the Docker build context. So I think you're right and that now it's better to place the Dockerfile in the module's root.

* You have a point with the docker-compose rename. 

* If you could use the WAR_FILE reference instead of the hardcoded war name that would be great.

* I missed the updated docker start scripts.

yvoswillens:
@jacarui thanks for the changes

Should the pushDockerImage profile sections also be changed to make use the 'new' maven plugin?

jacarui:
I've tried to migrate it but I've found some issues:
- I had to store dockerhub auth settings on maven settings.xml.
- couldn't publish both lastest and version image in one go.
- It's supposed to use deploy maven goal, which collides with nexus publication.

I think it doesn't payoff the migration for publication without changing profiles, etc.

yvoswillens:
Made some changes to this PR and created new branch / PR #981
Closing this one.

PR #975, opened 2018-04-08
Cockroach db compatibility

ruzkant:
[UPDATED]
This branch will attempt to add cockroachdb support to flowable.

 _I would like to get some feedback on preferences in regards to the implementation options._

Cockroachdb's core strategy seems to be to focus on being postgres compatible for ORM's and other tools. Although they may in future start supporting tools specifically so that cockroach's unique features can be better exploited.

I have run some tests with manually altered schema and this works, but automated schema upgrades present some challenges especially around transactions. 

**Implementation options**
**1. Create a new database in all places it is required, new scripts, new properties.**
This approach is challenging for the following reasons:
a. The JDBC driver returns product as postgresql and even the versions are postgres versions. The recommended way to detect cockroach over postgres is to run the following query: select version() from which CockroachDB can be extracted.

The posthres driver hardcodes product name, but could retrieve it from a handshake parameters called crdb_version as used by other cockroach tools like pgasync. I tested making this change to the driver, it's simple and works, but I am not sure if a pull request would be accepted for this, even though it seems more correct than hardcoding. I suspect it would break tools that work with the driver currently (it sounds like hibernate works to some extent).

b. Liquibase might need a custom extension as there is none yet. This is not particularly hard to do and it has the concept of priority when overriding databases. Need is maybe a strong word, because so far I have been able to make changes to changesets that do work on the postgres implementation.

c. More work is required to keep schemas up to date, especially the non-liquibase ones.

**2. Try to align postgres and cockroachdb implementations, to be compatible with each other.** 
This approach is challenging for its own reasons.
a. Foreign keys require indexes but you can't create them in the same transaction, not without postgres incompatible syntax. One option is to execute foreign keys in a second transaction which is not quite ideal (this approach is part of my initial commits). I opened an issue at cockroach here: https://github.com/cockroachdb/cockroach/issues/24578
So this may be resolved in future.

b. Other such issues may arise in future, and one the differentiation then becomes a bit more haphazard.

c. More care needs to go into postgres scripts to make sure they don't disobey cockroach semantics.

**Summary of issues found**
I opened issues at cockroach for the following:
1. A bug when calling jdbc metadata and then doing create table and insert queries in ddl scripts, sets off a sequence of events where it hangs.
https://github.com/cockroachdb/cockroach/issues/24578

This requires starting a fresh transaction right before ddl scripts are run.

2.  The issue with foreign keys and corresponding indexes not being allowed in the same transaction has had some discussion here where it seems it might be resolvable in a future update: https://github.com/cockroachdb/cockroach/issues/24626

**Comments**
* there is some advantage to making the scripts equal for postgres and cockroach, one could have a postgres backup database in the event of some  significant cockroachdb issue. I haven't tested it, but cockroach dumps claim to be loadable in other db's.
* I am not particularly happy with executing the sql ddl scripts in two transactions. 
* With liquibase the transaction issue is less of a concern as each changeset seems to execute in a transaction. 
* Moving the sql based migrations to liquibase is probably also an option, if that is a longer term goal for flowable.
* One can wait for some of the bugs to be resolved before taking this further

### DIFF
--- a/modules/flowable-engine-common/src/main/java/org/flowable/engine/common/impl/db/AbstractSqlScriptBasedDbSchemaManager.java
+++ b/modules/flowable-engine-common/src/main/java/org/flowable/engine/common/impl/db/AbstractSqlScriptBasedDbSchemaManager.java
@@ -241,6 +241,7 @@ public abstract class AbstractSqlScriptBasedDbSchemaManager implements DbSchemaM
         String exceptionSqlStatement = null;
         DbSqlSession dbSqlSession = getDbSqlSession();
         Boolean originalAutoCommit = null;
+        boolean localTransactionInProgress = false;
         Connection connection = null;
         try {
             connection = dbSqlSession.getSqlSession().getConnection();
@@ -318,9 +319,11 @@ public abstract class AbstractSqlScriptBasedDbSchemaManager implements DbSchemaM
                                     connection.commit();
                                 }
                                 connection.setAutoCommit(false);
+                                localTransactionInProgress = true;
                             } else if("commit".equalsIgnoreCase(sqlStatement)) {
                                 LOGGER.trace("Calling commit");
                                 connection.commit();
+                                localTransactionInProgress = false;
                             } else {
                                 Statement jdbcStatement = connection.createStatement();
                                 try {
@@ -355,6 +358,14 @@ public abstract class AbstractSqlScriptBasedDbSchemaManager implements DbSchemaM
             LOGGER.debug("flowable db schema {} for component {} successful", operation, component);
 
         } catch (Exception e) {
+            if(connection != null && localTransactionInProgress) {
+                try {
+                    LOGGER.debug("rolling back local transaction");
+                    connection.rollback();
+                } catch (SQLException sqle) {
+                    LOGGER.error("Could not rollback local transaction in schema changes", sqle);
+                }
+            }
             throw new FlowableException("couldn't " + operation + " db schema: " + exceptionSqlStatement, e);
         } finally {
             try {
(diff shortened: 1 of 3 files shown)


ruzkant:
This commit is just some first steps into this to get feedback before I go too far along a certain route. I am especially uncertain about the full implications of the changes to AbstractSqlScriptBasedDbSchemaManager as I am not clear on all the possible transaction semantics with the different engines, spring, standalone, as well as mybatis settings. 

Basically it seems to be running all updates in a transactional command and it is not really fine grained enough to solve the foreign key in a separate transaction issues discussed above.

ruzkant:
**On Indexes and foreign keys**
Apparently  indexes get created asynchronously at the end of a transaction which is why you can't create foreign key in the same transaction. There is an alternative syntax which works, but it doesn't work with postgres.

See this issue I opened on cockroach repo:
https://github.com/cockroachdb/cockroach/issues/24626

ruzkant:
I ran into another issue with liquibase where it is unable to use the jdbc metadata to check for existing indexes. I notice quite a few open items on cockroachdb github related to supporting different functions. The one for the indexes is related to:
https://github.com/cockroachdb/cockroach/issues/16971
https://github.com/cockroachdb/cockroach/issues/16491
https://github.com/cockroachdb/cockroach/issues/10495

It feels like the wild west out there, so I think I am going to close for now, I think quite a few issues will be solved in the near future, and for now will make do with static scripts to load the db schema.

BramGruneir:
@ruzkant Please let us know what issues you require and we can prioritize them.  Starting in our next release (as in what we're coding now), we're are putting a much bigger focus on external tool compatibility.

BramGruneir:
Thanks!  I'm actively tackling _pg_expandarray right now. 

I'll take a look through all of those other queries.  

Feel free to just comment on those already open issues.  It will greatly help our prioritization process.

ruzkant:
I am still testing cockroachdb for my own use, and built a simple compatibility layer, not suitable for general consumption. A layer of indirection solves every problem in computer science :)

I proxy the postgres driver, connection and statement classes, and detect and reorder ddl statements, and run them in appropriate transactions in the background. Introduce the odd index. This at least gets the apps started and works without any modifications to the current sql scripts or liquibase changelogs.

Just ran into correlated subquery issue on task application, and see correlated subquery support is not in 2.0 yet. I expect this to be a very difficult issue to get around if its use is prevalent, which I am not sure if it is.
https://github.com/cockroachdb/cockroach/issues/24684

ruzkant:
For anyone stumbling onto this... support for cockroachdb officially added in 6.4.2, yay!

https://github.com/flowable/flowable-engine/pull/1716

BramGruneir:
Wonderful!

paulhh:
Thanks to Cockroach Labs actively working to get it working with Flowable.  Nice piece of collaboration.

PR #977, opened 2018-04-10
spring dev profile enhancements

yvoswillens:
Because we don't wan't dev specific property files in the build artifacts the dev configurations are moved to configuration classes.


yvoswillens:
closed for #998

PR #1007, opened 2018-04-20
fix correct filenames for app war’s

stuba:
Not sure has it been plan to update all wars to app.war. This anyway fixes predefined docker compose files.


yvoswillens:
Hi.
For 6.3.1 we've moved to a new way of building the docker images (see #981) based on a Dockerfile.
Part of this change is the name change of the executable artifact used in the images.

This change is therefor not necessary IMHO.

Regards,

Yvo

PR #1020, opened 2018-04-26
Issue#1004 Add getters and setters in RuleEngineExecutorImpl to allow…

AminaZoheir:
… for customizing only the RuleEngineExecutor but not the hitpolicies, objectmapper and expressionmapper


tijsrademakers:
Thanks for the PR, we have merged the PR manually with some small changes.

Issue #1023, opened 2018-05-02
Colloection IN operations perform Contains not IN


AminaZoheir:
Collection A IN Collection B
Should do Collection B containsAll Collection A, but it does the opposite.

This applies to all collection operators (IN, NOTIN, ANY, NOTANY) incase of input type = collection.

tijsrademakers:
We refactored the collection operator names to make them more self explanatory. You can see the commit here https://github.com/flowable/flowable-engine/commit/281632cfb59d4a3f5f82d2d351d4bf65b57f4d71. We welcome any feedback,

PR #1024, opened 2018-05-02
Issue #1023 Remove condition for switching values if type is collection

AminaZoheir:
Changed IN operator behavior to perform IN not CONTAINS incase of collection input.


AminaZoheir:
Done in #1035

PR #1035, opened 2018-05-03
Issue#1033

AminaZoheir:
Add contains operations and change in operators behavior with collection type


tijsrademakers:
The DMN collection functions have been refactored already. This was done based on the changes you included in this PR, so thanks for triggering this refactoring

PR #1047, opened 2018-05-10
Default properties

filiphr:
* Adds a Spring Boot `EnvironmentPostProcessor` that allows using a `flowable-default.properties` which would be before the `application.properties`. This means that users can override properties by just putting an `application.properties` into their tomcat `libs` folder.
* Moves the UI and Rest app `application.properties` to `flowable-default.properties`
* Updates the defaults for the UI apps as discussed with @jbarrez 
* Add tests to the UI apps to validate that the ordering of the properties works correctly


filiphr:
I forgot to mention the `flowable-default.properties` can be added as a Spring Boot properties file as well. Have a look [here](https://youtrack.jetbrains.com/issue/IDEA-133445) for how to do it

PR #1055, opened 2018-05-13
Periodic code cleanups: null check not needed before 'instanceof'

dbmalkovsky:


dbmalkovsky:
Need to redo this request; it is faulty.  Closing

Issue #1073, opened 2018-05-20
Claim button remains disabled on all tasks after one task is claimed


aronim:
Module: flowable-ui-task-app
Page: /workflow/#/tasks

Scenario:
- I am a member of a group "Test Group"
- There are two User Tasks assigned to the group "Test  Group"
- I navigate to the Workflow Tasks Page
- I filter for "Tasks where I am one of the candidates"
- I am presented with a list containing the two User Tasks
- I click on the first task in the list
- I am presented with the task details with a claim button
- I click on the claim button
- The claim button is momentarily disabled while the remote request completes
- The claim button is replaced with a Save Button and a Complete Button
- I click on the second task in the Task List
- I am presented with the task details with a disabled claim button.

Problem:
The claimButtonDisabled is set to false on the creation of the TaskDetailController and is set to true when one task is claimed, however it is never set to false again. 

Proposed Solution:
Set claimButtonDisabled to false when the selectTask is changed.

tijsrademakers:
Merged your PR, thanks

PR #1074, opened 2018-05-20
#1073 Claim button remains disabled on all tasks after one task is claimed

aronim:
Resetting claimButtonDisabled when selectedTask changes


tijsrademakers:
Thanks, good catch

PR #1080, opened 2018-05-24
Fix add group to task

martin-grofcik:


martin-grofcik:
another possibility is to do not allow to create owner/assignee groups  <-> task relationship

PR #1084, opened 2018-05-28
Update DDL SQL to support the utf8mb4 character set

guerlab:


tijsrademakers:
Hi,

This PR does a lot of changes to the schema generation sql, including varchar length changes. We can't merge a PR with so many changes to the schema generation sql. If you think we should definitely consider supporting utf8mb4, please let us know and describe which changes are absolutely necessary for this.

Thanks

Issue #968, opened 2018-04-06
Inconsistency in the FormService


DavidPLamas:
Hi,

I'm using version 6.3.0-SNAPSHOT of your project. 

While mostly using the REST API, i'm facing a problem that is very annoying. 
I cannot POST/flowable-rest/form-api/form/model without having to send the "formDefinitionKey" property. I always get the following 404 error:
![image](https://user-images.githubusercontent.com/8373746/38429183-983817bc-39b5-11e8-8c59-c2f7e1a693e1.png)

So, after investigating for a while I found what could be the problem. Maybe I'm wrong but it seems that the parameters are being used in an incorrect order: 
![image](https://user-images.githubusercontent.com/8373746/38429382-12cafc74-39b6-11e8-9002-c45468b7eea2.png)

![image](https://user-images.githubusercontent.com/8373746/38428594-117cd498-39b4-11e8-9657-501410f61c77.png)

![image](https://user-images.githubusercontent.com/8373746/38428892-c96d64f0-39b4-11e8-9ede-2dacfee5db05.png)

In the 3rd image it looks like the parentDeploymentId is being initialized instead of the formDefinitionId. Further down in GetFormModelWithVariablesCmd.java file, it tries to find FormDefinition but it doesn't find any because neither formDefinitionKey nor formDefinitionId are defined.

Am I misunderstanding something ?

PascalSchumacher:
I guess this issue can be closed as the pull request with the fix (https://github.com/flowable/flowable-engine/pull/1087) has been merged and released with flowable version `6.4.0`.

PR #1087, opened 2018-05-29
Issue#968 Fix variable order calling formService method

jacarui:
Issue #968 points that `getFormModelWithVariablesByKeyAndParentDeploymentId` is being used with incorrect variable order.


tijsrademakers:
Thanks Javi!

Issue #1022, opened 2018-04-29
Replace hardcoded context roots from app-cfg.js


zwets:
When deploying the Flowable UI Apps under a different context root from the default `/flowable-*`, it is possible to make everything work using changes in property files (e.g. `flowable.common.app.idm-url` is configurable), except for one party pooper: the file `./src/main/resources/static/scripts/app-cfg.js` has hardcoded context root strings:

```
FLOWABLE.CONFIG = {
        'onPremise' : true,
        'contextRoot' : '/flowable-idm',
        'webContextRoot' : '/flowable-idm',
    'datesLocalization' : false
};
```

If these static strings weren't there, then the app wars could be transparently deployed under a different context root. Currently, putting an app on a different context root requires modifying the war file.

PascalSchumacher:
Indeed, it would be nice to have this changed/fixed.

tijsrademakers:
If there are ideas on how to do this in a different way please add it here to the PR.

ruzkant:
How about making it a JAX-RS resource and replace variable? Or use spring mvc templates, but that's an additional dependency probably.

paulstapleton:
I've come across this problem before as well and fixed it with a little JavaScript in` app-cfg.js`. It's a bit of a hacky solution where the context is parsed from the URL stored in the browser, but it has been working for me when using an alternative context name in tomcat.

There may well be problems with it that I am not aware of, but I'll leave the code here and you can decide if its something that is usable or not.

```
var FLOWABLE = FLOWABLE || {};

FLOWABLE.CONFIG = function($) {

    var config = $("script[src*='app-cfg.js']");
    var context = config[0].src.replace(location.origin, "").replace("/scripts/app-cfg.js", "").split("?", 1)[0];

    return {
           'onPremise' : true,
           'contextRoot' : context,
           'webContextRoot' : context,
           'datesLocalization' : false
     };

}(jQuery);
```

PascalSchumacher:
No idea if this is a good solution (I'm not a front-end devleoper), what about using:

`var baseURL = window.location.protocol + '//' + window.location.host;`

source https://stackoverflow.com/questions/35116446/angular-relative-paths-in-http-requests

PascalSchumacher:
I guess this can be closed now that #1100 has been merged?

tijsrademakers:
Thanks for the ping, this is implemented with the PR indeed.

PR #1100, opened 2018-06-04
Change to dynamic context root for ui-idm-app #1022

vzickner:
Use current `window.location.pathname` to figure out the current context root instead of hardcoded context roots. Tested with Mozilla Firefox and Google Chrome.

### DIFF
--- a/modules/flowable-ui-idm/flowable-ui-idm-app/src/main/resources/static/scripts/app-cfg.js
+++ b/modules/flowable-ui-idm/flowable-ui-idm-app/src/main/resources/static/scripts/app-cfg.js
@@ -17,10 +17,12 @@
 
 var FLOWABLE = FLOWABLE || {};
 
+var pathname = window.location.pathname.replace(/\/$/, '');
+
 FLOWABLE.CONFIG = {
 	'onPremise' : true,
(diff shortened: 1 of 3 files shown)


PascalSchumacher:
I guess the same should be done for the other ui-apps (modeler, admin, task).

vzickner:
Hi @PascalSchumacher, I have added also the adaption for task and modeler. For admin there is no `app-cfg.js`. Unfortunately, flowable-task is using subfolder, therefore it was necessary to enhance the regex for filtering. It's now not that readable as it was before.

martin-grofcik:
I 've tested several context path configurations. In the case when IDM app is mapped to the root.
I am not able to login. POST to https://app/authentication fails.

vzickner:
Hi @martin-grofcik, I improved the regex to also handle root as potential deployment location. Previously, in that case, the context root was set to `/` and all REST endpoints are using `context-root + /path/to/resource`. It tried to call `//path/to/resource` which is translated to `http://path/to/resource`. To avoid this it is necessary that the context-root does not end with a slash.

tijsrademakers:
@martin-grofcik could you test the updates of Valentin?

Issue #1039, opened 2018-05-07
HttpTask URL not encoded when saved


Heurazio:
**Description**
When using `HTTP Tasks` in the modeler, the `Request URL` is not correctly URL-Encoded when saving the process model. 

**Problem**
This results in errors starting the process (HTTP 500) when there are e.g. not encoded query parameters in the `Request URL`. This can be found in the logs:

```
org.flowable.engine.common.api.FlowableException: Invalid URL exception occurred
    ...
    Caused by: java.net.URISyntaxException: Illegal character in query at ...
   ...
```

**Linked Topics**
[Process won't start with HTTP 500 error](https://forum.flowable.org/t/error-starting-new-process-instance-http-500/1956)
[Correct URL Encoding for HTTP Task](https://forum.flowable.org/t/http-task-url-encoding/1982)

vzickner:
The problem in the sample you mentioned is that the space character is not valid in `URI`-objects (the URI is `https://example.com/v3/domain/messages?subject=Test 2&text=Test message 2`). We should either validate it during modeling or escape the parameters afterwards.
@tijsrademakers @jbarrez which option would you suggest?

tijsrademakers:
@vzickner I think we should make sure that we escape the parameters. It should be ok to fill-in a value with a space in the modeler.

PascalSchumacher:
I guess this can be closed now that #1101 has been merged in https://github.com/flowable/flowable-engine/commit/22f0d379b3c91391b5c335215e1572358ca3b8f0?

Kirinfm:
Hi @tijsrademakers ,
For Example, we use flowable to grant members permission. GitLab api use `https://gitlab.com/api/v4/projects/flowable%2Fflowable-engine/members` .
What should I do.

PR #1101, opened 2018-06-04
Fix encoding of URL before conversion to URI object (#1039).

vzickner:
Fix not encoded parts of the URL by decode the original path, query and ref and afterwards encode those again (default behavior of `URI` constructor).


jbarrez:
Merged manually (see https://github.com/flowable/flowable-engine/commit/22f0d379b3c91391b5c335215e1572358ca3b8f0)

PR #1102, opened 2018-06-05
Update ContentItem lastModified timestamp during save

DennisFederico:
ContentItem lastModified timestamp should be updated always on Save


DennisFederico:
Need to wait a bit to fix the unit test, will reopen later

PR #1106, opened 2018-06-06
Issue#1105 Map new model name and key to process identifier and name on save and process identifier on clone

AminaZoheir:


tijsrademakers:
Thanks!

PR #1109, opened 2018-06-06
Task assignment manager refactoring

martin-grofcik:


modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/cfg/DefaultTaskAssignmentManager.java
```diff
@@ -108,6 +112,9 @@ public void deleteGroupIdentityLink(Task task, IdentityLink identityLink) {
 
     @Override
     public void addUserIdentityLinkToParent(Task task, String userId) {
-
+        if (userId != null && ScopeTypes.CMMN.equals(task.getScopeType()) && StringUtils.isNotEmpty(task.getScopeId())) {
```
martin-grofcik: We can consider whether we should keep it consistent with the process engine or not.

PR #1111, opened 2018-06-06
Missing events when using ChangeStateBuilder.

roberthafner:
Tijs/Joram, Here is an example of some of the events that I think are missing when changing the state of a process. 

Rob


roberthafner:
Replaced by PR 1185 which fixes the missing events and adds additional test coverage.

PR #1143, opened 2018-06-20
HttpResponseHandler & HttpRequestHandler BpmnError propagation

aturkovic:
Maybe I am mistaken, but it seems like `BpmnError`s thrown by `HttpRequestHandler` or `HttpResponseHandler` should be propagated instead of wrapped as `FlowableException` to allow custom implementations of these interfaces to affect the flow. Catching all `FlowableException`s with `<flowable:mapException>` does not seem like the proper solution since we cannot differentiate between different reasons that caused them to be thrown if we want to handle them differently: 

- Exception while invoking HttpRequestHandler
- Exception while invoking HttpResponseHandler
- HTTP exception occurred
- IO exception occurred
- Invalid URL exception occurred

Another way of catching exceptions from `HttpRequestHandler` or `HttpResponseHandler` would be to make a new `FlowableException` superclass, but according to the [documentation](https://flowable.org/docs/userguide/index.html#_exception_strategy), this is not the preferred way in order to avoid a big exception hierarchy.


tijsrademakers:
Thanks for providing this PR. Yes the BpmnError should be propagated like you implemented it.

PR #1156, opened 2018-06-27
Feature: rootCause filtering on mapException

alturkovic:
New feature as discussed [here](https://forum.flowable.org/t/extending-mapexception-to-allow-filtering-based-on-rootcause/2147).

This PR adds support for additional `mapException` filtering through a new attribute called `rootCause`. Only exceptions that were caused by the configured `rootCause` will be caught by the error `boundaryEvent`. 

Flag `includeChildExceptions` is still respected, so the combination of a specific exception to catch, what root cause of the exception to handle and whether child exceptions will be caught should all be compatible.


alturkovic:
I have just realized I reused my master branch for this PR. I will close it and reopen a clean one...

PR #1161, opened 2018-06-28
Identity link refactoring

mkiener:
Adding java-doc to identity links, refactoring existing (still backwards compatible) links to  fully support the future, generic scoped objects and definitions.
Fixed updating links with dynamic process injection where identity links might change.


n0mer:
@mkiener `TaskIdentityLinkResourceTest` test is broken, maybe you might consider fixing it?

image

PR #1163, opened 2018-06-28
Http request body encoding

alturkovic:
Http request body should be encoded with a configurable charset instead of using the default one from StringEntity


rgorzkowski:
Hi,
I have only recently started using Flowable but I also met encoding problem with http task.

I see you changed HttpActivityExecutor.java, HttpActivityBehaviorImpl.java and HttpRequest.java
but what about HttpTaskJsonConverter and stencilset-bpmn.json (maybe something more) ? 
Is it intentionally unchanged? Maybe I do not understand something.

filiphr:
@rgorzkowski you are right. There has been no intention to leaving it unchanged. I have created https://github.com/flowable/flowable-engine/issues/1211. In case you are interested in providing a contribution for it we can help you out

alturkovic:
My bad, didn't fully understand what modeler changes were necessary, will add a new PR to fix this.

PR #1173, opened 2018-07-06
Fix the the problem of method enableDuplicateFiltering() sometime will be invalid.

zhanjr:
When I use StrongUuidGenerator to generate uuid,orderByDeploymentId() may be wrong.
You used orderByDeploymenTime() in [if] at line 83,but used orderByDeploymentId() in [else] at line 91.


tijsrademakers:
Thanks, good catch

PR #1185, opened 2018-07-19
Fix: Events sent when changing the state of a process.

roberthafner:
* Add missing ACTIVITY_CANCELLED events.
* Add missing ACTIVITY_STARTED event when embedded subprocess begins.
* Add missing ACTIVITY_STARTED event when call activity begins.

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/agenda/ContinueProcessOperation.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/agenda/ContinueProcessOperation.java
@@ -12,7 +12,8 @@
  */
 package org.flowable.engine.impl.agenda;
 
-import java.util.Collection;
+import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 
 import org.flowable.bpmn.model.Activity;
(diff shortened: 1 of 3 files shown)


tijsrademakers:
Hi Rob,

Also for this PR, sorry for the delay in reviewing.
Could you fix the merge conflicts, then I'll do some testing and merge it.

Thanks

roberthafner:
Hi Tijs,

The conflicts in the PR have been resolved. There was some changes to the existing tests that made some of the event checking I added no longer valid. For now I removed the validation from those tests since I think the remaining event validation checks in the other test methods provide good coverage for the issues that I found. In addition, a number of tests were added in the latest flowable code.  I did not add (event) checks to the new test methods. At some point I'd like to add additional checks to the other test methods but that will have to wait until I have some more spare cycles.

Thanks!
Rob

tijsrademakers:
Hi Rob,

Thanks for resolving the conflicts. In the PR there's a change to execute the boundary event activity behaviours after executing the activity behaviour of the Activity where the boundary event is connected to. Is that really needed for this PR? Can it be solved differently as well?

Best regards

roberthafner:
Hi Tijs,

Yes, the change to the execution of boundary event activity behaviors is necessary since the boundary events for a flow element are received prior to the events for the flow element in which it is bound. For an event based client such as ours, the client does not have knowledge that the flow element exist (or is active) when the boundary event is received and thus does not understand how to process it.

Removing the change will result in a number of tests failing. 
testSetCurrentActivityToActivityWithTimerForSimpleProcess
testSetCurrentActivityToTaskWithTimerInSubProcess
testSetCurrentActivityToTaskWithTimerInSubProcessAndExecuteTimer
testSetCurrentActivityWithTimerToActivityWithTimerSimpleProcess
testSetCurrentExecutionToActivityWithTimerForSimpleProcess
testSetCurrentExecutionToTaskWithTimerInSubProcess

For example, in the case of testSetCurrentActivityToActivityWithTimerForSimpleProcess, line 370 will result in a TIMER_SCHEDULED event bound to the UserTask "firstTask" being received first prior to the ACTIVITY_STARTED event for the UserTask "firstTask".

Sorry, it might not have been clear that this change is needed since the root GitHub comment wasn't updated to include the latest comment from the commit message that also includes:

* Ensure ACTIVITY_STARTED event is sent before boundary event.

See the comment here: https://github.com/flowable/flowable-engine/pull/1185/commits/2fc4700fd555060032977173b4773037cacbe413

With respect to your question regarding whether the issue could be resolved in another way, I suspect there may be other ways to resolve the issue. However, I was trying to minimize the potential impact of the change by allowing the boundary event executions to be created in the original order and then executing the boundary behaviors after the behavior of the flow element (in which they are bound) to ensure the events would be sent in the desired order. In general, it's unclear to me why the boundary event behavior would execute before the behavior of the flow element in which they are bound. Was that intentionally done for some use case that I am overlooking? It wasn't obvious to me that it was intentional since there weren't any tests that failed when the order was changed. That being said, I'm definitely open to alternative approaches that resolves the event order issue. 

Regards,
Rob

tijsrademakers:
Thanks for the detailed explanation Rob. I understand the reasons for doing the implementation like it's done in the PR. This sequence of execution is better from a pure BPMN perspective. So it's good change to do I think.

PR #1208, opened 2018-08-03
Fix start of a process via REST API and message name.

vzickner:
Previously when starting a process with a message through the REST
interface, the engine expected to be the message id same to the message
name. In case they were not the same, it failed for the message name
with a FlowableException 'No message start event found' since the
message name was searched as message id. When using the message id
instead of the name to start the process, it failed with 'Cannot start
process instance by message: no subscription to message with name'.

Now the provided message name is converted to a message id, in case it
exists. As a fallback, it will use the message name as message id,
in case this API is used outside of the REST API.


vzickner:
Closed in favor of commit 706993b40687c7350bfc8fc19e2ac2438c81a9cf

PR #1212, opened 2018-08-07
Extending modeler support for body encoding as described in https://g…

### LINKED ISSUE #1211: Support for Http request body encoding in the modeler
With PR https://github.com/flowable/flowable-engine/pull/1163 support for a body encoding for the HTTP request was added to Flowable.

This issue is for extending this support up to the modeler via the `HttpTaskJsonConverter` and the `stencilset-bpmn.json`

alturkovic:
Fix for PR ['Http request body encoding'](https://github.com/flowable/flowable-engine/pull/1163) which did not add the necessary modeler properties.

Fixes #1211


filiphr:
The failing test is due to the missing changes in the BPMN specific `HttpTaskJsonConverter`

alturkovic:
Is there a reason there are duplicated `httptask` properties in `stencilset_cmmn.json`? I see there are two of each, for an example: `httptaskrequestbodypackage` is declared twice.

filiphr:
>Is there a reason there are duplicated httptask properties in stencilset_cmmn.json?

I think it is just a copy paste mistake

alturkovic:
> I think it is just a copy paste mistake

1. Would you like me to delete those?
2. Would you like me to add CMNN functionality in this PR also?
3. I have noticed I didn't save encoding variable in the execution when `request.isSaveRequest()`. Would you mind if I fix that also in this PR?

filiphr:
Yes to all questions 😀. Feel free to do everything as part of this PR.

In case you want some special way of structuring your commits feel free to squash and rearrange as you see fit. Let us know once you are done so we can have another look at it. 

Thanks for working on this

alturkovic:
No problem, I did add the functionality, so I should cover it thoroughly. Let me know if I missed something (again) :)

filiphr:
Can you have a look at the tests on Travis? They are failing (I think due to the new fields in the converters)

alturkovic:
My bad, I ran it locally with `mvn clean install` before pushing and it didn't show it for some reason? Nevermind....

I am looking into the tests and I it looks like I am still missing something. I believe there is an additional field in the test now due to the default value for `bodyEncoding`? 

However, if I change `httptaskrequestmethod` value in `test.httpservicetaskmodel.json` and run tests in `HttpServiceTaskConverterTest`, I can see that changed value in both tests if I catch it in the debugger on line 58 of the said test. But I am not seeing the same behavior if I change `httptaskrequestbodyencoding`. Test `convertJsonToModel` shows my custom value correctly, but `doubleConversionValidation` shows the default (`UTF-8'). 

Could you please help point me in the right direction?

filiphr:
I'll have a look tomorrow morning and try to help you out. Will look into why it didn't work when you run it locally (maybe the module is not in the default modules)

filiphr:
@alturkovic have a look at the `ServiceTaskJsonConverter` the converting from `BpmnModel` into JSON is handled there. That's the reason why you don't see your change

alturkovic:
Exactly what I was looking for, thanks!

alturkovic:
`mvn clean package` did not run the test again, after seeing it fail on Travis, I had to manually run it. Is there a better way to run the build locally?

tijsrademakers:
Hi, on Travis we use the profile distro to test all modules and not just the core modules. So you can run mvn clean install -Pdistro to do it in a similar way as it's done on Travis

PascalSchumacher:
`mvn clean install -Pdistro -DskipTests=false` should be very close to what travis run, see: https://github.com/flowable/flowable-engine/blob/master/.travis.yml#L23

filiphr:
@alturkovic there is one last things which worries us about this PR. That is the default value of `UTF-8` when converting the JSON. This would mean that for all users and their models they will now have `UTF-8` as their encoding the next time they deploy them. This is not backwards compatible. Would it be OK for you if we leave this without a default value? This would lead again to the test having 2 field extensions only

alturkovic:
No problem, that actually makes a lot of sense... I will get to it this evening or tomorrow morning.

filiphr:
Thanks for the PR @alturkovic. I have merged it into master

PR #1227, opened 2018-08-15
Support for JUnit Jupiter in the flowable-engine

filiphr:
Migrate flowable-engine and other dependent modules tests from JUnit 3 / 4 to Junit Jupiter

* Move used assertions from `TestCase` into `AbstractTestCase` (which is no longer extending `TestCase`) in order not to migrate all tests
* Add `@CleanTest` annotation which instructs the internal extensions to clean all deployments of a test
* Add `@EnsureCleanDb` annotation which instructs the internal extensions to perform an assertion for a clean DB and drop and create if it wasn't clean
* Add `InternalFlowableExtension` - a basis for other internal extensions that offers the following possibilities:
  * Perform a deployment when `@Deployment` is present before each test method
  * Validate history data after each test
  * Delete history jobs and deployment after each test
  * Assert and ensure clean db after each test (or after all test if `TestInstance.Lifecycle#PER_CLASS` is used)
  * Support for injecting the ProcessEngine, ProcessEngine services, ProcessEngineConfiguration
  and the id of the deployment done via `@Deployment` int tests and test lifecycle methods
* Add `LoggingExtension` - an extension which performs the test logging for the Flowable tests
* Add `PluggableFlowableExtension` - an extension which creates a default ProcessEngine and caches it for all tests
* Add `ResourceFlowableExtension` - an extension that starts a new ProcessEngine from a configured resource
before each test (needs to be used with `@RegisterExtensions`)
* Add `InternalFlowableSpringExtension` - an extension that uses `SpringExtension` to get the ProcessEngine from the Spring ApplicationContext
* Adapt `TestHelper` to user test class, method and the `Deployment` annotation
* Add `@DeploymentId` annotation for injecting the id of a deployment in a test method or test lifecycle methods
* Do not use `@Deployment` in ProcessDefinitionEventsService (reason: the listener is required during the deployment)
* Use `@EnabledOnOs` for the ShellTaskTest (Should result in 2 ignored tests on each machine)
* Remove empty test from StandardAgendaFailingTest and disable endless loop test
* Adapt `LDAPTestCase` to close the `ApacheDSContainer` after all tests
* Modules migrated to use JUnit Jupiter
  * flowable-camel
  * flowable-crystalball (partially)
  * flowable-cxf
  * flowable-dmn-spring-configurator
  * flowable-engine
  * flowable-form-spring-configurator
  * flowable-groovy-script-static-engine
  * flowable-http
  * flowable-ldap-configurator
  * flowable-spring

Modules that have not been migrated will be done in subsequent commits.

I have weird tests failing locally, let's see what Travis says

### DIFF
--- a/modules/flowable-app-engine-spring/pom.xml
+++ b/modules/flowable-app-engine-spring/pom.xml
@@ -124,6 +124,21 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>com.h2database</groupId>
             <artifactId>h2</artifactId>
             <scope>test</scope>
--- a/modules/flowable-app-engine/pom.xml
+++ b/modules/flowable-app-engine/pom.xml
@@ -75,6 +75,21 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.vintage</groupId>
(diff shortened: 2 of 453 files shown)


lamtrhieu:
Hi @filiphr 
I am really interested in viewing this PR but the file changes is really too much for me. Why don't you break it down into smaller ones instead ?

filiphr:
Hey @lamtrhieu thanks for your interested in the PR. The PR is actually already done in the smallest possible pieces (See the multiple commits). However, due to the scope of the PR (migrating the entire testing framework and test of the flowable-engine) there are more than 2000 tests that are affected by this change. Have a look at all files which are not ending on `Test.java` for the actual more meaningful changes

dbmalkovsky:
@filiphr This change looks wrong in
modules/flowable-camel/src/test/java/org/flowable/camel/AsyncPingTest.java

 -   @Override
 +   @Test
    public void tearDown() throws Exception {
        List<Route> routes = camelContext.getRoutes();
        for (Route r : routes) {

I think it should be @AfterEach instead of @Test

filiphr:
You are completely right @dbmalkovsky. Thanks a lot for letting us know. I have just fixed that in https://github.com/flowable/flowable-engine/commit/b65bfb572b914588e49d90aa2faa84b068a5b554.

PR #1232, opened 2018-08-17
Fix parallel execution of CallActivity

enchobelezirev:
Currently there was a problem when some Activity is being used with ParallelMultiInstanceActivityBehavior. The problem was that an FlowableOptimisticLockingException was being thrown and this prevents the execution of the concurrent executions(the retry of the failed job was not efficient because the whole execution is being retried which leads to retry of whole steps, gateways or even ProcessInstances). This PullRequest introduces a way to overcome this situation by using a custom Monitoring operation. The operation is "watching" the concurrent executions progress and based on it is deciding whether the parallel processing has finished.

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/FlowableEngineAgenda.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/FlowableEngineAgenda.java
@@ -12,6 +12,8 @@
  */
 package org.flowable.engine;
 
+import java.util.List;
+
 import org.flowable.engine.impl.persistence.entity.ExecutionEntity;
 
 /**
@@ -40,4 +42,7 @@ public interface FlowableEngineAgenda extends Agenda {
     void planDestroyScopeOperation(ExecutionEntity execution);
 
     void planExecuteInactiveBehaviorsOperation();
+
+    void planMonitorParallelMultiInstanceOperation(ExecutionEntity executionEntity, List<? extends ExecutionEntity> executions);
+
 }
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/agenda/DefaultFlowableEngineAgenda.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/agenda/DefaultFlowableEngineAgenda.java
@@ -12,6 +12,8 @@
  */
 package org.flowable.engine.impl.agenda;
 
+import java.util.List;
+
 import org.flowable.common.engine.impl.agenda.AbstractAgenda;
 import org.flowable.common.engine.impl.context.Context;
 import org.flowable.common.engine.impl.interceptor.Command;
@@ -25,17 +27,20 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * For each API call (and thus {@link Command}) being executed, a new agenda instance is created. On this agenda, operations are put, which the {@link CommandExecutor} will keep executing until all
- * are executed.
+ * For each API call (and thus {@link Command}) being executed, a new agenda
+ * instance is created. On this agenda, operations are put, which the
+ * {@link CommandExecutor} will keep executing until all are executed.
  *
- * The agenda also gives easy access to methods to plan new operations when writing {@link ActivityBehavior} implementations.
+ * The agenda also gives easy access to methods to plan new operations when
+ * writing {@link ActivityBehavior} implementations.
  *
- * During a {@link Command} execution, the agenda can always be fetched using {@link Context#getAgenda()}.
+ * During a {@link Command} execution, the agenda can always be fetched using
+ * {@link Context#getAgenda()}.
  *
  * @author Joram Barrez
  */
 public class DefaultFlowableEngineAgenda extends AbstractAgenda implements FlowableEngineAgenda {
-    
+
     private static final Logger LOGGER = LoggerFactory.getLogger(DefaultFlowableEngineAgenda.class);
 
     public DefaultFlowableEngineAgenda(CommandContext commandContext) {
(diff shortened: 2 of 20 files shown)


filiphr:
@enchobelezirev thanks a lot for the PR. Can I please ask you to remove the changes that are not related to the PR. Changes such as reformatings which have not been changed. 

For example I saw something like:

```
processEngineConfiguration.getEventDispatcher().removeEventListener(testListener);
```

to 

```
processEngineConfiguration.getEventDispatcher()
    .removeEventListener(testListener);
```

Thanks a lot in advance. This would make it much easier and faster for us to review the changes. If you can't please do let us know.

enchobelezirev:
Hello and thanks for the comment,
I will reformat the code. Sorry for that :)

enchobelezirev:
Hello,

I have used the xml configuration files which are located in the ./ide-settings/eclipse in order to reformat the code. Could you please check it whether it is OK now?

tijsrademakers:
@enchobelezirev thanks for creating this PR. Could you point me to a unit test that is failing in the current Flowable BPMN Engine implementation? I would like to see the case where the parallel multi instance construct is giving the optimistic lock exceptions. Thanks!

enchobelezirev:
Hello @tijsrademakers and thank you for the reply. First of all, sorry for the long delay - I was Out of Office the last month. Basically, what I did here to produce the issue was to create a simple Unit test which you can find here: https://github.com/enchobelezirev/flowable-test
The test uses the CallActivity and is running in parallel. 
You can run it and observe the result of the test -> in the console output you should be able to see a lot of FlowableOptimisticLocking Exceptions.
Write to me again if you need something more from me :)

enchobelezirev:
Hello,

Any update here? Is there any chance to merge this PR? I accept any recommendations how to implement the fix.

Thanks,
Encho

jbarrez:
@enchobelezirev: Thanks for the PR and for the excellent unit test! 

Using the unit test we were able to find the root cause of the problem. Let me briefly explain. 

All steps of the subprocess are asynchronous and exclusive. As the parent callactivity has parallel multi-instance characteristics, this means that the tasks of each subprocess are executed exclusively for each _subprocess instance_. The exclusive lock is held on the child process instance, not the root one (for performance). 

Now, when the last task of the subprocess is executed, in the same transaction also the end event will be executed. This will cause the child process instance to be deleted and control is given back to the parent process instance. This is where the problem arises: if there are multiple concurrent child process instances finishing at the same time (as is here the case), the _completion_ of the callactivity will happen in context of the parent process instance. As the exclusive lock was on the child process instances and not on the parent, there are now multiple concurrent updates to the parent executions, causing the _FlowableOptimisticLockingException_. 

Do note that the process eventually finished correctly, there is no error wrt functional execution here (but getting the exceptions is indeed not nice).

The fix provided in this PR would indeed fix this problem, however the logic is quite heavy to fix this particular (edge) case. In fact, a similar problem exists when it comes the joining of asynchronous paths in a parallel gateway. The solution there is to create a new exclusive job for the join specifically. 

Given that, we've implemented a fix that is consistent with that approach, please have a look at https://github.com/flowable/flowable-engine/compare/fix_1232. Here, a new attribute (completeAsync) is introduced (it's not possible to determine this situation from the process model as there can be multiple sequence flow into the end event. Also this way is the most performant, not interfering with regular usages). When true, a new exclusive job is created that will handle the completing of the callactivity using an exclusive lock on the parent process instance. With it, the unit test you provide doesn't throw any optimistic lock exception anymore.

@enchobelezirev : please have a look at the branch and let me know if you have any feedback before I'm merging it master.

@tijsrademakers : can you review the branch?

tijsrademakers:
@jbarrez looks good, and nice catch @enchobelezirev

enchobelezirev:
Hello @jbarrez  and @tijsrademakers,

First of all, I want to thank you for looking into this.
I have some questions.
First, in the given situation when some subprocess finishes it will trigger the EndExecutionOperation which will make a job locked by the parent execution which in this case is the CallActivty. This is really awesome but my question is, when the leave method of the MultiInstanceActivityBehavior will be called? Without the logic, the leave method was called for each ending execution and in the method the  variable nrOfCompletedInstances was being incremented and based on the result it was determined whether the parent to continue or not. Will the new logic reach the leave method and increment the variable? I am asking because the other thing I have observed is FlowableOptimisticLockingException during the increment of the variable.

Second, if the fix in  https://github.com/flowable/flowable-engine/compare/fix_1232 is enough then my PR is not needed?

And last, @jbarrez I do not understand why the logic is quite heavy for the fix?

Really, thanks a lot!!! I am really happy that I was able to help overcoming this problem!

Thanks,
Encho

enchobelezirev:
Also, I will use this thread here to ask you if you think that using custom behaviour of tasks is helpful and usable. You can find the forum post: https://forum.flowable.org/t/multi-instance-call-activity-inter-relations/2598 :)

Thanks a lot - I appreciate every your comment :) !!!

jbarrez:
@enchobelezirev : 

- The leave will indeed be called asynchronously when the end event is being handled (in the new job). This means that the problem of the optimistic lock exception for variables (and historic variables) is also solved as these will be done exclusively (we did indeed see this exception too before).
- Indeed, this PR is not needed anymore, I'll close it once it's merged to master (will do that soon). 
- The "heavy" was alluding to creating a new operation for every call activity. Even though it technically is indeed correct, it's something we try to avoid and keep the set of operations small. So heavy probably wasn't the right word for it :-)

Many thanks for your quick response!

enchobelezirev:
Thanks @jbarrez, then I am more than happy with the fix. :)

Many thanks that you were able to look into the problem and create a fix for it!!

jbarrez:
Branch merged. Closing this PR.

enchobelezirev:
Hello,

When release of the fix will be available? I want to consume it into our project.

Thanks,
Encho

filiphr:
@enchobelezirev right now we are looking for a release date some time early next week

PR #1238, opened 2018-08-20
Spelling corrections

dbmalkovsky:
Also corrected: doesn?t -> does not

### DIFF
--- a/docs/public-api/README.md
+++ b/docs/public-api/README.md
@@ -1,6 +1,6 @@
 # Flowable Rest API Documentation
 
-This project contains all informations & tooling about Flowable Public Rest API.
+This project contains information and tooling about Flowable Public Rest API.
 
 ## Introduction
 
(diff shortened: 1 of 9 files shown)


filiphr:
@dbmalkovsky I think that doesn?t is automatically created and comes from `doesn’t` in the different resources. [Here](https://github.com/flowable/flowable-engine/blob/8431b4b1632fd89ac9fb0e22ce3bb77b8a6d3800/modules/flowable-rest/src/main/java/org/flowable/rest/service/api/runtime/task/TaskIdentityLinkResource.java#L49) for example. I think that the quote is wrong. In any case it can be named `does not` instead in any case. Just mentioning as next time we generate it it might become `doesn?t` again 😄

dbmalkovsky:
@filiphr In my organization we typically hand craft our swagger and then write the code instead of generating it.  I was always taught that in professional/formal/documentation scenarios that contractions were at best frowned upon and at worst not acceptable. In the examples here they appear to cause character set issues (let us not forget about Microsoft smart quotes). Maybe we need to find the root occurrences and  fix them there once and for all.

PR #1239, opened 2018-08-20
Allow IdGenerator to be overriden via spring configuration.

roberthafner:


filiphr:
@roberthafner thanks for the PR. We added support for this in a slightly different way. Have a look at https://github.com/flowable/flowable-engine/commit/562bac942bd00d96359a892b71413696f1dfad03. We use Spring `ObjectProvider`, so if there is an `IdGenerator` bean defined, then we are going to use it, otherwise we will use the `StrongUuidGenerator` (no need to expose the bean from our side). Hope this works for you.

PR #1241, opened 2018-08-23
Add the labels to sequence flow elements in BPMN diagram overview

vzickner:
In case the model contains a label for a sequence flow, it will be now added to the overview diagram. Furthermore, the label is also visible in the admin and the debug application.


tijsrademakers:
Thanks a lot, great PR

PR #1245, opened 2018-08-25
Migrate IDM and Form engine tests to Jupiter

filiphr:


n0mer:
smth with `org.flowable.osgi.blueprint.BlueprintBasicTest`

filiphr:
Thanks for letting me know @n0mer. I am exposing the osgi additional imports and exports for the common module now. It should be green now

PR #1249, opened 2018-08-27
Java 11

PascalSchumacher:
Make flowable build on Java 11 (and 12).

Requires an update to Spring and Spring Security 5.1, Spring Boot 2.1 and JPA 2.2.

Tests of flowable-ldap-configurator are skipped, because ApacheDS does not support Java 11 yet.

Travis builds on Java 11 and Java 12 are now longer allowed to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ matrix:
     dist: xenial
   - jdk: openjdk11
     dist: xenial
+  - jdk: openjdk12
   - jdk: openjdk-ea
     branches:
       only:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,6 @@ matrix:
       - master
     if: TRAVIS_PULL_REQUEST != false
   allow_failures:
-  - jdk: openjdk11
   - jdk: openjdk-ea
   - name: "Flowable 5 tests"
 
--- a/modules/flowable-app-rest/src/test/java/org/flowable/rest/app/FlowableRestApplicationSecurityTest.java
+++ b/modules/flowable-app-rest/src/test/java/org/flowable/rest/app/FlowableRestApplicationSecurityTest.java
@@ -73,7 +73,11 @@ public class FlowableRestApplicationSecurityTest {
             "metrics-requiredMetricName",
             "scheduledtasks",
             "httptrace",
-            "mappings"
+            "mappings",
+            "caches",
+            "caches-cache",
+            "health-component",
+            "health-component-instance"
         )
     );
 
@@ -89,7 +93,7 @@ public class FlowableRestApplicationSecurityTest {
     @Test
     public void nonAuthenticatedUserShouldNotBeAbleToAccessActuator() {
         String actuatorUrl = "http://localhost:" + serverPort + "/flowable-rest/actuator";
-        ResponseEntity<Object> entity = restTemplate.getForEntity(actuatorUrl, Object.class);
+        ResponseEntity<String> entity = restTemplate.getForEntity(actuatorUrl, String.class);
 
         assertThat(entity.getStatusCode())
             .as("GET Actuator response status")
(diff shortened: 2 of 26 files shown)


PascalSchumacher:
Currently broken, because build on master is broken. :-(

tijsrademakers:
Hi @PascalSchumacher ,

What is the status of this PR to your opinion? I see you changed the Spring Boot 2 dependency version to 2.1.0.M2. Is that necessary for the Java 11 support?

PascalSchumacher:
Spring 5.1 is the first Spring version to officially support Java 11.

With Spring 5.0 some tests fail on Java 11.

Spring Boot 2.1 is used because Spring Boot 2.0 does not support Spring 5.1.

Some commits of this pull request could be merged right now (Not sure how much that will help anybody). Other commits require offical releases of Spring/Spring-Security/Spring-Boot that will only happen in a few weeks/months (I could not find a planned release date for Spring Boot 2.1).

This pull request is a proof of concept. I created it so see how much has to be changed to make the build pass on Java 11. I did not do any manual testing of the flowable wars.

tijsrademakers:
Thanks for the feedback Pascal

PascalSchumacher:
According to https://spring.io/blog/2018/09/13/spring-boot-2-1-m3-available-now Spring Boot 2.1 will be released at the end of october.

PascalSchumacher:
I updated the pull request to use Spring Boot `2.1.0.RELEASE`.

The pull request now uses only release versions of dependencies.

PascalSchumacher:
I build the idm, modeler, task and admin apps and deployed them to a Tomcat 9. I started Tomcat with Java 11 checked that everything deployed fine, logged in and clicked around in the apps. Then I repeated the same after starting Tomcat with Java 8.

filiphr:
@PascalSchumacher is it OK if we revert https://github.com/flowable/flowable-engine/pull/1571 and add the update of mockito here?

The reason is that the tests in the different UI apps are now failing (at least in IntelliJ) due to conflicting versions of ByteBuddy. We are importing the spring-boot-dependencies in the [flowable-spring-boot](https://github.com/flowable/flowable-engine/blob/c8e9072ee1732c0666a71d72aed466c00ec4286a/modules/flowable-spring-boot/pom.xml#L36-L40) pom and the apps are using that. This leads to having Mockito 2.24.0 and ByteBuddy 1.7.11 (defined in the spring-boot-dependencies). Mockito 2.24.0 needs ByteBuddy 1.9.7. Spring Boot 2.1 manages ByteBuddy 1.9.7.

PascalSchumacher:
Sure. Sorry about the breakage. 😞

filiphr:
>Sure. Sorry about the breakage. 😞

It is not a problem, it is actually our fault that we don't have the UI apps in the travis build.

>What about adding the tests for the UI apps to the travis build?

I think this would be a good idea, we would be able to avoid a problem like this in the future.

PascalSchumacher:
I have rebase this branch onto master and the mockito update actually breaks the Java 12 build:

```
[ERROR] testParseMessagedDefinitionWithExtension  Time elapsed: 0.545 s  <<< ERROR!
org.mockito.exceptions.base.MockitoException: 
Mockito cannot mock this class: interface java.util.Map.
Mockito can only mock non-private & non-final classes.
If you're not sure why you're getting this error, please report to the mailing list.
Java               : 12
JVM vendor name    : Oracle Corporation
JVM vendor version : 12-ea+30
JVM name           : OpenJDK 64-Bit Server VM
JVM version        : 12-ea+30
JVM info           : mixed mode, sharing
OS name            : Linux
OS version         : 4.4.0-101-generic
Underlying exception : java.lang.IllegalArgumentException: org.mockito.codegenMap$MockitoMock$493793903 must be defined in the same package as org.mockito.codegen.InjectionBase/package
	at org.flowable.engine.test.bpmn.event.MessageEventDefinitionWithExtensionElementsTest.testParseMessagedDefinitionWithExtension(MessageEventDefinitionWithExtensionElementsTest.java:39)
```

I guess it is the same problem that caused JUnit to revert the update: https://github.com/junit-team/junit5/commit/754f1eb0f106ab043d79de546d49769690a731b6

Let's revert and wait for a fix.

filiphr:
According to members of the mockito team (https://twitter.com/TimvdLippe/status/1094615569062604801) seems like we are mocking `java.util.*` classes we should review the test that fails and remove that mocking.

filiphr:
I've just merged your PR @PascalSchumacher. Thank you very much for your work on it

PascalSchumacher:
Thanks for merging!

PR #1254, opened 2018-08-29
Update docker compose config files.

rgorzkowski:
Replace dots in environment properties with underscore


yvoswillens:
Thanks for the PR @rgorzkowski

PR #1256, opened 2018-08-30
Compensation Throwing Event for Service Task  - Documentation or Implementation Bug?

roberthafner:
I did not find a test case which only compensates a single service task (without a subprocess being involved). Based on the documentation in section 8.2.31 it appears for this use case the compensate throwing event should be defined like such:
`<intermediateThrowEvent id="throwCompensation">
    <compensateEventDefinition activityRef="bookHotel" />
</intermediateThrowEvent>
`

where the value in activityRef refers to the activity id of the service task that completed and will have compensation run on it. However defining the definition in this manner does not produce the expected results as the compensation handler never executes.  

The attached test case shows that defining the compensate throwing event with the activityRef attribute having a value of the id of the service task that is the compensation handler provides the expected behavior. For example:

`<intermediateThrowEvent id="throwCompensation">
    <compensateEventDefinition activityRef="undoBookHotel" />
</intermediateThrowEvent>
`

Given this it's unclear whether the example in the documentation or the behavior is incorrect. Can someone verify which is correct and possibly add this test case for future reference?

Regards,
Rob

``

### DIFF
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/event/compensate/CompensateEventTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/event/compensate/CompensateEventTest.java
@@ -49,6 +49,18 @@ public class CompensateEventTest extends PluggableFlowableTestCase {
 
     @Test
     @Deployment
+    public void testCompensateServiceTask() {
+
+        ProcessInstance processInstance = runtimeService.startProcessInstanceByKey("compensateProcess");
+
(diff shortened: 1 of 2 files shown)


roberthafner:
I see version 6.4.0 is scheduled to be released soon. I'm hoping this issue can be resolved in that release.

tijsrademakers:
Hi Rob,

Sorry for the delay. The issue has been fixed, you can now use the activity id for which you want to trigger the compensation. The fix has been implemented with backwards compatibility to the original behaviour. Thanks for finding the issue and creating the BPMN XML example.

PR #1272, opened 2018-09-10
modify @Primary  annotation  to @ConditionalOnMissingBean annotation

qiudaoke:
When I use dev in my local environment  and i use @Primary in 
DataSource。Program error

### DIFF
--- a/modules/flowable-app-rest/src/main/java/org/flowable/rest/conf/DevelopmentConfiguration.java
+++ b/modules/flowable-app-rest/src/main/java/org/flowable/rest/conf/DevelopmentConfiguration.java
@@ -35,7 +35,7 @@ public class DevelopmentConfiguration {
     protected static final String DATASOURCE_PASSWORD = "flowable";
 
     @Bean
-    @ConditionalOnMissingBean(DataSource.class)
+    @ConditionalOnMissingBean
     public DataSource developmentDataSource() {
         return DataSourceBuilder
             .create()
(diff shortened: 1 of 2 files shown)


filiphr:
@qiudaoke I am not 100% sure what you are trying to do. Are you using `flowable-app-rest` as a dependency in your own project? 

This `@Primary` here is just to make sure that it can have a different db in development mode and it is mostly for development of this particular app meant.

In case you want to build your own rest application you should use `flowable-spring-boot-starter-rest` and create your own Spring Boot application

qiudaoke:
@filiphr 
 I'm pretty sure that I used a custom DataSource in the dev environment.
`flowable-ui-common` the module 。There is a class in this module. org.flowable.ui.common.conf.DevelopmentConfiguration。

I am sure that when there are multiple @Primary annotations, they are in conflict.

@Profile({"dev"})
public class DevelopmentConfiguration {
    @Bean
    @ConditionalOnMissingBean
    public DataSource developmentDataSource() {
        return DataSourceBuilder
            .create()
            .driverClassName(DATASOURCE_DRIVER_CLASS_NAME)
            .url(DATASOURCE_URL)
            .username(DATASOURCE_USERNAME)
            .password(DATASOURCE_PASSWORD)
            .build();
    }

tijsrademakers:
Hi, 

We only use the DevelopmentConfiguration with the dev profile active and it's only for local development and testing within the Flowable project. You can use any other profile name when starting your application and make sure you don't run into @Primary issues, by default it's not used at all. So I'm closing the PR for this reason.

PR #1276, opened 2018-09-11
Support JSON variables in the REST api

LEDfan:
Hi

We are using flowable-rest and want to use json variable types.
It seems that everything was put in place to support it, since we can use them in flowable-modeler and flowable-task, however in flowable-rest and flowable-admin it gives some troubles.
With this PR I try to fix three problems:
 
 1.  Creating variables of type json doesn't work using REST. I.e. this fails:
```
curl --user admin:test -X PUT --header "Content-Type: application/json" http://localhost:8080/flowable-rest/service/runtime/process-instances/.../variables -d '[{"name": "input", "type": "json", "value": {"test": 123}}]'
```
This returns:
```
{"message":"Bad request","exception":"Variable 'input' has unsupported type: 'json'."}
```

 2. Reading variables of type json will return them as `serializable`
 3. Flowable-admin also shows the variables as `serializable`

This PR will make sure that for 2 and 3 type `json` is returned and the actual json is returned instead of a placeholder.

image

Do you think this fix makes sense? Should I change something?

### DIFF
--- a/modules/flowable-common-rest/src/main/java/org/flowable/common/rest/variable/JsonArrayRestVariableConverter.java
+++ b/modules/flowable-common-rest/src/main/java/org/flowable/common/rest/variable/JsonArrayRestVariableConverter.java
@@ -14,7 +14,12 @@
 package org.flowable.common.rest.variable;
 
 import org.flowable.common.engine.api.FlowableIllegalArgumentException;
+
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
 
 
 public class JsonArrayRestVariableConverter implements RestVariableConverter {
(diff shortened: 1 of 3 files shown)


tijsrademakers:
Hi,

Thanks for the PR and the clear description. It's a good enhancement indeed. What I was wondering is why the VariableType LinkedHashMap and ArrayList are used. When a json variable is retrieved from the database it will be returned as a Jackson ArrayNode or ObjectNode (https://github.com/flowable/flowable-engine/blob/master/modules/flowable-variable-service/src/main/java/org/flowable/variable/service/impl/types/JsonType.java). So I'm a bit confused about this part. Could you share why it's done like this?

Thanks

LEDfan:
I just pushed a commit which changes the VariableType from `LinkedHashMap` and `ArrayList` to `ObjectNode` and `ArrayNode`.

The problem was that when using the REST api the json is converted to `ArrayList` or `LinkedHashMap`, (I think) because of this call https://github.com/flowable/flowable-engine/blob/1ccb67e77dbae20f2f7ebb75230bb3c87b15dc30/modules/flowable-rest/src/main/java/org/flowable/rest/service/api/runtime/process/BaseVariableCollectionResource.java#L90.

However as you told the rest of flowable is storing JSON using the Jackson classes. So in my original patch it would still return json not coming from the rest api (e.g. using the `saveResponseVariableAsJson` on a HTTP Task) as `serializable`. With the new commit this would also be returned as json.

I'm glad that you are interested in the patch :+1:

tijsrademakers:
Thanks for the PR. I've made some small changes to it, hope that's ok. Let me know if you have any feedback or questions about this.

LEDfan:
Sure, thanks for taking care of the remaining issues!

Just one last question: do you have any idea when this will be released into a stable version of flowable?

Issue #1271, opened 2018-09-07
Maven surefire rerunFailingTestsCount does not work with JUnit Jupiter


filiphr:
Since the migration to JUnit Jupiter the `rerunFailingTestsCount` is not taken into consideration.

According to https://maven.apache.org/surefire/maven-surefire-plugin/examples/rerun-failing-tests.html seems like this configuration is only supported on JUnit 4.x. There is https://github.com/junit-team/junit5/issues/1558 that has a discussion to add support for flaky tests.

It might be worth considering upping the timeout on the flaky tests to a value higher than 5000ms

PascalSchumacher:
Not sure how helpful it will be in this case, but in general https://github.com/awaitility/awaitility is great for making asynchronous tests both stable and fast.

filiphr:
I think it would be the same, as we are already doing something similar in the places where the flaky tests fail. In any case thanks for pointing it out.

PR #1288, opened 2018-09-15
flowable-image-generator: task names missing (box too small?)

qiudaoke:
In DefaultProcessDiagramCanvas.drawTask(), the boxHeight is computed as follows:

int boxHeight = height - 16 - ICON_PADDING - ICON_PADDING - MARKER_WIDTH - 2 - 2;

In my environment (bpmn files created using Eclipse BPMN2 Modeler with default settings), this leads to the situation, that in method drawMultilineText() the condition

currentHeight + height > boxHeight

becomes true (for a single line of text), and thus task names are completely left out.

I would suggest to add an else-block in method drawMultilineText(),
in order to draw at least one line of text.

> 
**if(currentHeight + height > boxHeight) {
        if (!layouts.isEmpty()) {
           [...]
        } else { // at least, draw one line
            layouts.add(layout);
            currentHeight += height;
        }
        break;
}**

Changed implementation in flowable-image-generator/DefaultProcessDiagramCanvas.drawMultilineText().

**empty box before the change**
image

**box incl. label after commit**
image


PascalSchumacher:
Test failure:

>[ERROR] testOneTaskProcess  Time elapsed: 6.823 s  <<< ERROR!
>org.flowable.common.engine.api.FlowableException: time limit of 5000 was exceeded

seems unrelated and caused by the build instability after the JUnit 5 update (see: https://github.com/flowable/flowable-engine/issues/1271)

qiudaoke:
@tijsrademakers 
Can we combine this PR?

tijsrademakers:
Thanks!

PR #1290, opened 2018-09-15
Increase timeouts for a lot of asynchronous tests from 5 to 7 seconds make b…

PascalSchumacher:
…uild more stable on travis-ci.

This change is necessary because the maven surefire plugin option rerunFailingTestsCount does not work for junit 5 tests. (#1271)


tijsrademakers:
Already merged with the Spring upgrade PR, thanks

PR #1316, opened 2018-09-26
BPMN stencilset changed to add the category attribute to the UserTask

llyorshch:
This change enables the category attribute for the UserTask in the modeler.


tijsrademakers:
Thanks!

PR #1318, opened 2018-09-26
modify ui-task-app zh-CN.json to Support Chinese language 

qiudaoke:
modify ui-task-app zh-CN.json to Support Chinese language


jbarrez:
Thanks, that was fast!

PR #1323, opened 2018-10-02
Adding support to query tasks with formKey

ilgrosso:
In Apache Syncope we embed Flowable as BPMN engine and we frequently need to query active forms, e.g. tasks where `FORM_KEY_ IS NOT NULL`.

We have a local extension of `TaskQueryImpl` with such a capability, but I thought it could be of general use.


tijsrademakers:
Thanks a lot for the PR, definitely makes sense for general use. I think the or query has a small mistake:

or RES.FORM_KEY_ = #{orQueryObject.withFormKey}

this probably should be something like:

or RES.FORM_KEY_ IS NOT NULL

Could you validate this?

ilgrosso:
> Could you validate this?

Argh, you're completely right - let me fix that with an additional commit...

ilgrosso:
@tijsrademakers please take a look again, it should work now...

tijsrademakers:
Thanks!

ilgrosso:
Thanks to you for merging :-)

PR #1336, opened 2018-10-09
expose single variable accessors in process and case instance interfaces

mkiener:


tijsrademakers:
Closing PR due to inactivity

PR #1339, opened 2018-10-10
add customize font

qiudaoke:
I plan to use Chinese fonts, but the default font is Arial


tijsrademakers:
Yes good idea to be able to customize this with Spring boot properties. Thanks for the PR

PR #1343, opened 2018-10-11
Process instance migration v2

DennisFederico:
Gateways, basic multi instance support. Supports for changing assignee in activity migration mapping


filiphr:
@DennisFederico why is the change in the flowable.cfg.xml needed?

DennisFederico:
@filiphr the change in flowable.cfg.xml its not needed, must have "snuck" in with last the commit to my branch after the PR creation. otherwise CI test wouldn't fail. Will revert that on my next commit

PR #1351, opened 2018-10-20
flowable-osgi: Use depends-maven-plugin to let BlueprintBasicTest use dependency versions defined…

PascalSchumacher:
… in pom.xml

The `org.apache.aries.blueprint.core`, `org.apache.aries.proxy`, `org.apache.aries.util` could have been kept in the tests, but I think it is best to gather all versions in one place. If you disagree I will update the pull request accordingly.


tijsrademakers:
That's indeed more convenient, thanks

PR #1356, opened 2018-10-21
AppDeployer keeps publishing same version number

ruzkant:
This manifests in a way in which the apps are deployed without version number increases, because of the incorrect parameter causing the method AppDeployer.getMostRecentVersionOfAppDefinition to return null all the time.


tijsrademakers:
Thanks, nice catch

PR #1360, opened 2018-10-22
Fix bug in dmn CollectionUtil

amporsim:


yvoswillens:
Thanks for the PR @amporsim !

PR #1364, opened 2018-10-23
Use UTF-8 instead of JVM default encoding.

PascalSchumacher:


tijsrademakers:
Hi Pascal,

I think we can't do some of these changes without breaking backwards compatibility. Maybe we can make it an option in the engine configuration so it can be set?

PascalSchumacher:
Hi Tijs,

the flowable code already contains a lot of places where UTF-8 encoding is used. Therefore I think it would be for the best to use it consistently.

Of course your knowledge of flowable is miles ahead of mine, so you are in a better position to judge this.

I will revert any changes you believe to be too risky or we can just close the whole pull request.

PR #1366, opened 2018-10-23
Use the caught exception as cause when constructing a new exception r…

PascalSchumacher:
…ather than swallowing it.


tijsrademakers:
Hi Pascal,

In the REST resources we don't expose the full stack traces in most cases. Some of the changes in this PR would change this. Is there a specific reason to also do it for the REST resources?

PascalSchumacher:
No specific reason. I will revert the REST resources parts.

PR #1376, opened 2018-10-27
ProcessDiagramGenerator should create appropriate symbols for each SubProcess

qiudaoke:
Before fix
image
After fix
image


tijsrademakers:
Thanks for the PR. Could you maybe remove the following change:

-  if (isTriggeredByEvent) {
+ if (type.equals(EventSubProcess.class)) { 

And for the transaction element add a separate method to draw it and call that from the Transaction activity or use another boolean to pass if the sub process is a transaction?

qiudaoke:
@tijsrademakers 
hi,
I made some changes。

tijsrademakers:
Thanks

PR #1400, opened 2018-11-16
add getDeployedStartForm and GetDeployedTaskFormCmd  api in FormService

qiudaoke:


tijsrademakers:
Hi,

The call to GetDeploymentResourceCmd uses the form key to lookup the start form or task form. This could be true for your use case, but I don't think this is a generic solution. Can you think of a generic solution?

tijsrademakers:
Closing the PR with the comment made some weeks back. Please comment if there is a more generic solution possible.

PR #1403, opened 2018-11-19
Fixed bug where defaultFlow was not set for Inclusive Gateway when co…

paulstapleton:
…nverting BPMN model to JSON

Hi,

I have seen a bug where the BPMN JSON converter module was not setting the defaultFlow flag on inclusive gateways. The code only did this for exclusive gateways.
This occurs when converting from a BPMN model to JSON and I have seen this when exporting and importing in the flowable-modeler tool.

I have also updated a unit test showing the problem and a patch to fix this.

Please let me know if I need to make any changes.

Regards,
Paul Stapleton


dbmalkovsky:
I thought we agreed that per the discussion in this forum [post](https://forum.flowable.org/t/coding-style-and-standards/154) that wild card imports were not standard.  Is that still true?

tijsrademakers:
Yes, that's still valid. @paulstapleton thanks for the PR, could you maybe change the wildcard imports to listing every import statement separately?

paulstapleton:
Thanks for the feedback. I wasn't previously aware of the coding standards. I'll try my best to follow them in the future.

I have now updated my commit. Please let me know if there are any more changes needed.

tijsrademakers:
Thanks for the changes and the PR. And we should make our code styling more known. We will work on this.

PR #1409, opened 2018-11-22
RU activity instance implementation

martin-grofcik:


modules/flowable-engine/src/main/java/org/flowable/engine/impl/history/DefaultHistoryManager.java
```diff
@@ -354,7 +345,7 @@ public void recordHistoricDetailVariableCreate(VariableInstanceEntity variable,
             HistoricDetailVariableInstanceUpdateEntity historicVariableUpdate = getHistoricDetailEntityManager().copyAndInsertHistoricDetailVariableInstanceUpdateEntity(variable);
 
             if (useActivityId && sourceActivityExecution != null) {
-                HistoricActivityInstanceEntity historicActivityInstance = findActivityInstance(sourceActivityExecution, false, false);
+                HistoricActivityInstanceEntity historicActivityInstance = findHistoricActivityInstance(sourceActivityExecution, false);
```
martin-grofcik: The problem is that runtimeActivityInstance does not always exist for this update. That's why we have to rely on HistorycActivityInstance. Another possibilty is to create runtime activityInstance form the currently exiting historic activityInstance

PR #1413, opened 2018-11-27
Add Send Message symbols in the modeler - Flowable release 6.4.0

wberges:
Hello,

Is it possible to add 2 missing `Send Message` BPMN symbols in the Modeler?
- Send Task
- Intermediate Message Throwing Event

Here is additional info:
https://forum.flowable.org/t/send-task-does-not-appear-in-process-modeler/453/8

Thanks a lot for your very good work :)
Best Regards
William


wberges:
Hello Tijs,

I wouldn't say that missing BPMN symbol is a bug, just a lack, isn't it? :)
But to be honest, I don't know clearly how to choose between all these forums/topics, so certainly I posted in the bad one :)

What I want is explained here: https://forum.flowable.org/t/send-task-does-not-appear-in-process-modeler/453/7
I want to be able to send a message to a dedicated instance, not to broadcast my message (signal purpose).
> BPMN Message and Signal are both events; the main difference is that a Message is triggered to a single target while a Signal is broadcast to multiple targets. In other words, the Message is a point-to-point communication while the Signal is a broadcast communication.

In the [BPMN 2.0 spec](https://www.omg.org/spec/BPMN/2.0/PDF), I wasn't able to find the restriction you mention (only Signal allowed to send messages to other process instances).
In other modelers, a Send Task exists allowing to communicate with other pools.
It's also strange to have a Receive Task without its counterpart (Send Task) :)

And meanings of `Send Task` and `Intermediate Message Throwing Event` should be the same, except that a Send Task should be able to execute code in addition to sending a message (though, event with listeners...).

But perhaps I'm wrong?

Best Regards
William

tijsrademakers:
Hi William,

Ok thanks for the explanation. The main reason for asking is that the PR has 1124 file changes in it :-). So I'll close the PR, but we can still continue the discussion here.

You say you would like to send the message event to a dedicated instance, and I think that's also a correct meaning according to the spec, but how do you determine the correct dedicated instance? Other then within the same process instance I don't see this easily possible.

I understand your explanation of the expected send task behaviour. I think we can provide such an implementation. We can start with the engine implementation and then expose it in the Modeler as well.

Best regards,

Tijs

wberges:
Hi Tijs,

Thanks for answering so quickly :)

In my mind: 
- Message is supposed to be associated/correlated to a pair {Message catching event, process parameters (process definition, process Id, business key, process variables...) }, allowing a point-to-point connection. 
It means that you must first find the right process instance using the association parameters (using a Query?) before firing the Message. 
- Signal is supposed to be sent to all Signal catching events without any other association value than the Receive Signal event (broadcast). Parameters may be used, but only as input, not for association/correlation.

It's for me the main difference between a Signal and a Message.
Do you agree with this vision?

Cheers
William

PR #1423, opened 2018-11-28
added X_FORWARDED_PROTO==https for secure cookies when using a proxy

mattydebie:
I noticed the cookies were not secure cookies when flowable is acting behind a load balancer. So I added a check for when the header `X_FORWARDED_PROTO` is set.


mattydebie:
Spring's `HttpHeaders` class does not contain a constant for the `X-FORWARDED-PROTO` header ([issue here](https://jira.spring.io/browse/SPR-14232)). 
I can however just turn it into a plain string if that's better?

tijsrademakers:
Ok, yes if you could turn it into a plain string that would be good, thanks.

Issue #1415, opened 2018-11-27
Admin: state change and process migration mapping values grow without refresh


lsmall:
I am exploring the Change State and Process Instance Migration features and I noticed that if I do not refresh the screen or logout/login, the values remain in the session and accumulate. Since the "Change state" button doesn't appear until the relevant source and target tasks are selected each time, I expected the selected values would be cleared if I closed the dialog without changing the state. The same is true for process migration. Since the "Add mapping" button only shows when the source and target tasks are selected, I would not expect previous selections to show. However, when I select the "Add mapping" button without refreshing, I see multiple mappings (previous and current):

Mapping 1 from (Task A) to (Task A) - Mapping 2 from (Task A) to (Task A)

I cannot see the multiple values for state change, but both migration and state change display multiple confirmation dialogs (one for each mapping) when this issue occurs. The migration fails, but the state change succeeds. Either way, errors in the log show that certain values were not valid or found. My UI experience is limited, so I do not know how to fix this issue.

lsmall:
Shouldn't the dialog close after executing the migration or state change? It seems like the user can execute the same one again which is no longer valid if the previous requests succeeded. Also, I noticed the same request was being sent for each confirm dialog. So if 2 mappings were there, the user gets 2 confirm dialogs and each request has the same migration payload containing all mappings.

tijsrademakers:
The current handling of the change state and migration dialogs is very basic and you are right that it needs more logic to handle multiple changes in a good way.

lsmall:
I think I have resolved removing the change state/migration buttons after the command executes.

lsmall:
I was able to hide the execution buttons, but not the original issue with redundant values. I had a peer with extensive UI development experience review the code and she recommended it be refactored/rewritten. In other words, it appears to be a related to how it is implemented in general and not specific to this feature.

Opened new pull request [1433](https://github.com/flowable/flowable-engine/pull/1433) just to  hide execution elements after state change or migration request completes.

PR #1424, opened 2018-11-29
flowable-1415 hide state/migration buttons after execution

lsmall:
https://github.com/flowable/flowable-engine/issues/1415#issuecomment-442892035

Hide the change state/migration buttons after the command executes


lsmall:
Still testing

PR #1431, opened 2018-12-02
Update JavaMail to version 1.6.3 (use new maven coordinates)

PascalSchumacher:


PascalSchumacher:
JavaMail `1.6.3` does not contain any bugfixes or enhancements and dependencies like `commons-mail` still use the old maven coordinates. Therefore I think it is best to not merge this for now.

PR #1444, opened 2018-12-11
Making Admin UI tenant aware.

mattydebie:
My attempt on making the Admin UI tenant aware.

I added the tenant fields to the IDM UI, so that one can enter and update the tenant ID for a user. This tenant ID is now also passed to the UI.

The admin UI adds the tenantID filter right before sending the request to the flowable-rest. This way you can not 'disable' it and it should be safe.

I don't think this fixes everything but it could be a step in the right direction.

Every comment or remark is welcome.


tijsrademakers:
Sorry for the delay in responding. The approach that is taken in the PR looks fine. But except for GET calls and the historic task query, there are also several other calls like a POST deployment call and a lot of other calls that require a tenantId in the body. What are your thoughts about this?

mattydebie:
Yeah I agree that it's not fully tenant-aware yet. But I will look further into it now that I know you will accept the way I handled it :smile: 
One thing I did notice is that I forced tenant awareness on lists by always filtering on `tenantID` if present. This means that if someone were to guess a *taskID* for example, he can stil view it by entering it manually in the browser. 
Any help on that front?

yvoswillens:
I think the preferred implementation would be backend only and centralised.
By adding the tenant ID to the requests only in the backend you avoid issues like you mention (guessing the task ID).
Based on the request method (POST, GET etc) the tenant ID could be added as a request parameter or as a request body property.
That last approach needs some verification; will this be sufficient for all requests.

mattydebie:
@yvoswillens I'll try something along those lines. Thanks for the input!

mattydebie:
@yvoswillens When adding the tenantId to the query params for every request, you can still 'guess' the ID's. The following request will still return the process definition, even though it does not share my tenant ID 
```
http://localhost:8080/flowable-rest/service/repository/process-definitions
?deploymentId=ca14889b-2306-11e9-9893-0242ac120003&tenantId=matty
```
the above request is composed in the FlowableClientService

mattydebie:
@yvoswillens @tijsrademakers  The latest commit shows how I would go about it by using the `x-tenant` header.
With 20fa82a, the ProcessDeployments/-Definitions are made completely tenant-aware. If you think this way is acceptable, I'll handle the rest of the endpoints as well.

mattydebie:
Hey @yvoswillens, @tijsrademakers could you confirm wether you like this approach? Sorry for bothering yet again.. :stuck_out_tongue:

yvoswillens:
Hi @mattydebie 
Will get back to you before the end of the week.

mattydebie:
any progress?

yvoswillens:
Hi @mattydebie 
I think that getting the tenantID from the User object in REST backend SecurityContext has several advantages over sending it from the client in a header.

Could you also take a look at the 'REST API interceptors'. 
These interfaces can be implemented to execute the needed logic; like determining (and evaluating) the tenant id and f.e. adding it to the query object.
This approach will probably be cleaner.

Also; because this will be a very intrusive change it needs to be possible to disable this 'tenant check' with a property. (Or perhaps even the other way around; disabled by default)

Let me know what you think or if I need to elaborate on certain point.

Thanks!

VeridianDynamics:
Hi @yvoswillens 
my name is Jan and I'm Matty's colleague. It appears that I have inherited this PR from him and I'm going to implement tenant checking for the process-instances endpoint as well. I have a few questions about what you said here.

>I think that getting the tenantID from the User object in REST backend SecurityContext has several >advantages over sending it from the client in a header.
Do I understand correctly that this would mean autowiring SecurityContext in the relevant REST resource, fetching the User object from it and based on its tenantID, modify the query?

>Could you also take a look at the 'REST API interceptors'.
>These interfaces can be implemented to execute the needed logic; like determining (and evaluating) >the tenant id and f.e. adding it to the query object.
>This approach will probably be cleaner.
I saw some of those interceptor interfaces, but I don't know how these would be used. I haven't seen any implementation within Flowable itself, so I suspect that the user is supposed to implement their own and somehow include it at runtime?

Thanks!

yvoswillens:
Hi @VeridianDynamics 

Yes. Both assumptions are correct.
You can get the user from the Spring Security context (auto wiring is one of the options).

The interceptor interfaces were introduced to provide clean hook points for custom logic.
Beans implementing these interfaces that are available on the context will be wired in.
This way you can manipulate the query before executing it. For example adding the tenant id.

Regards,

Yvo

VeridianDynamics:
Hi @yvoswillens 
I was successful in writing a custom interceptor and getting the security context, but unfortunately, the Context only contains an instance of `org.springframework.security.authentication.UsernamePasswordAuthenticationToken`. This token contains only an IP address I'm connecting from, the authorities ('access-idm, etc...') and the Principal name ("admin"). There is no tenant information and no User object. Can you please advise on how I can obtain it?

Thanks,
Jan

yvoswillens:
Hi.

You should be able to get the user (FlowableUser) by with UserDetailsService.loadUserByUsername(principalName).

Regards,

Yvo

VeridianDynamics:
Hi @yvoswillens 
thanks a lot for the information, this does indeed work. Or so I thought yesterday. When I tried what you suggested, I was able to get the right user. However, when I rebuilt the project with Maven today, the principal is suddenly always "rest".

I looked at where this comes from and found that the credentials used for the internal request are lifted from a `org.flowable.ui.admin.domain.ServerConfig` and the more I look at it, the more I'm wondering how is it even possible that I was able to get the 'admin' user login earlier - it makes sense that I see the 'rest' user instead, it's what's used internally to get the data from the actual rest endpoint, so it does not make sense to get the security context of the logged-in user while the actual request is being performed from localhost as 'rest' user.

Is it possible that what I experienced earlier was in fact a bug or a potential security vulnerability? Or maybe I'm just missing some crucial aspect of how Spring applications are composed?

Thanks,
Jan

yvoswillens:
Hi Jan,

I'm not sure if I understand what you're saying.
You have the Flowable Admin app running. You login to the Admin app with a certain user. There you have the Flowable API endpoints defined; with a username and password.
When you access a view in Flowable Admin the data is fetch from a certain Flowable API with that defined username / password.
And you're saying you are seeing the Admin app user (not the one defined with the API endpoint) in the SecurityContext of the API endpoint?

VeridianDynamics:
Hi @yvoswillens 
Thanks for the comment, I think the mystery is almost solved. I was not aware that it's possible to set up the REST API endpoint in Flowable Admin (I assume this is the ServerConfig in backend) and it has changed the user used from 'admin' to 'rest' all by itself for some reason. Pretty weird, but I'm 100% positive I didn't touch this. See a screenshot: https://i.imgur.com/5sdHGKg.png

Anyway, since this user is always used to perform the internal REST call (the setting is 'Admin UI'-wide), there is no way to know who is the person using the Admin UI in Flowable REST. And it makes perfect sense, it could be running on a completely separate machine. Unless I'm missing something, it seems that sending it in the header is quite reasonable if we want to filter out other tenants when using the Admin UI; for separate REST calls performed with the correct user against Flowable REST, the interceptor should work just fine.

Thanks!
Jan

tijsrademakers:
Closing the PR due to lack of activity. Please reopen when you still would like to see this functionality added to Flowable Admin.

PR #1446, opened 2018-12-11
Remove unused variable

dbmalkovsky:
It appears that during the refactoring of database connections
(4f28b4f43cdfce3abac0497a9919d3bbabd7f306) and (5c2c28e4fe8b1a9f6a8bd551f41ed83105367f68) the code for closing the connections was moved into individual routines and is no longer needed
in the parent routine.


dbmalkovsky:
My version failed on JDK8 this time and the message is now:  [ERROR] There was an error in the forked process GC overhead limit exceeded.   It would appear that it might dependent on which VM the job is run on.

PR #1454, opened 2018-12-16
  Fix Swagger/OpenAPI queryVariable operation enum case

### LINKED ISSUE #1437: Unsupported variable query operation: EQUALS
flowable-rest.war 6.4.0 
POST /query/historic-process-instances
when the parameters like :
{
  "variables": [
    {
      "name": "applyUserId",
      "operation": "EQUALS",
      "value": {},
      "type": "string",
      "variableOperation": "EQUALS"
    }
  ]
}

the respose is :
{
  "message": "Bad request",
  "exception": "Unsupported variable query operation: EQUALS"
}

PascalSchumacher:
Fixes "Unsupported variable query operation: EQUALS" (fixes #1437)

Not sure this is correct, my knowledge of Swagger/OpenAPI is limited.

### DIFF
--- a/modules/flowable-cmmn-rest/src/main/java/org/flowable/cmmn/rest/service/api/engine/variable/QueryVariable.java
+++ b/modules/flowable-cmmn-rest/src/main/java/org/flowable/cmmn/rest/service/api/engine/variable/QueryVariable.java
@@ -17,6 +17,8 @@ import org.flowable.common.engine.api.FlowableIllegalArgumentException;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 
+import io.swagger.annotations.ApiModelProperty;
+
 /**
  * @author Frederik Heremans
  */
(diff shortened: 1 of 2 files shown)


PascalSchumacher:
Any feedback on this pull request?

PascalSchumacher:
Build failed on Java 9 and 10 because it took to long. I guess #1595 need to be merged first.

filiphr:
@PascalSchumacher I think this is not the most correct place for the change. That file is semi automatic generated. I think that we need to annotated the `QueryVariable#operation` with Swagger annotations for the correct name. 

The correct names are defined in the [JSON query variable format](https://www.flowable.org/docs/userguide/index.html#restQueryVariable) section of the documentation. They are camelCase.

PascalSchumacher:
@filiphr Thank you very much!

I have updated the pull request. 

The `flowable-swagger-process.yaml` generated by `flowable-oas-generator` now seems to be correct (as far as I can tell).

By the way: There is a third `QueryVariable` class located in the `flowable-ui-task-logic` module, but that module does not have a `swagger-annotations` dependency, so I did not update this class.

PR #1483, opened 2019-01-01
Move errorLogging profile to parent POM, in order to tests if it spee…

PascalSchumacher:
…ds up travis build.


PascalSchumacher:
No (noticeable) effect.

PR #1484, opened 2019-01-02
HSQL create table command requires the 'default' before 'NOT NULL'.

garybrc:
As per the HSQL documentation, the 'default' clause is required before 'NOT NULL'. This fix is required to allow the HSQL database to be created for unit testing.


jbarrez:
Thanks!

PR #1489, opened 2019-01-03
AbstractEngineConfiguration: Remove unused javafx.beans.binding.When …

PascalSchumacher:
…import to fix build on Java 10+.


jbarrez:
Woops! Good catch!

PR #1508, opened 2019-01-07
add jdbcStatementTimeout and jdbcBatchProcessing

qiudaoke:


tijsrademakers:
Hi @qiudaoke, this change is doing the following by default:

if (isJdbcBatchProcessing()) {
        configuration.setDefaultExecutorType(ExecutorType.BATCH);
}

So we are using that batch executor type with this change by default. From reading the MyBatis documentation the default is the simple executor type. Could you elaborate a bit more about this change?

PR #1510, opened 2019-01-08
Updated Date query parameter example

verheyenkoen:
ISO-8601 date without seconds is not accepted by the API.


tijsrademakers:
thanks

PR #1523, opened 2019-01-11
Improve select queries: remove unnecesary condition to ignore task variables

inakihn:
Improve select queries: remove unnecesary condition to retrieve
only process variables ignoring task variables


inakihn:
That has to do with new line characters:
image

Execution.xml has LF
HistoricProcessInstance.xml had CRLF, and probably my intellij changed it to LF. Should I leave it as it was?

tijsrademakers:
Yes please leave the file as it was and only do the changes necessary for the PR. Thanks!

inakihn:
I will try to redo the PR :)

PR #1524, opened 2019-01-11
Improve select queries: remove unnecesary condition to retrieve only process variables 

inakihn:
Improve select queries: remove unnecesary condition to retrieve only process variables ignoring task variables.


inakihn:
HistoricProcessInstance.xml is shown as being completely changed, although I have tried to only change the relevant line:

image

filiphr:
You would need to revert the changes in the file, then do the change (do it in a simple text editor if you can't with IntelliJ) commit and then commit and force push to your branch.

PR #1526, opened 2019-01-11
Improve select queries: remove unnecesary condition to retrieve only process variables ignoring task variables

inakihn:


inakihn:
Finally managed to not change the new line characters on HistoricProcessInstance.xml
I had to do it outside intellij with vi.

tijsrademakers:
Thanks!

PR #1533, opened 2019-01-16
Made sure you can send objects to flowable-rest form-data

mattydebie:
When sending a JSON object to the form-data endpoint /flowable-rest/service/form/form-data:
```sh
curl -H "Content-Type: application/json" --request POST -u rest:rest -d '{ "taskId": "<< TASKID >>", "properties": [{ "id": "test", "value": { "test": true }, "type": "json" }, { "id": "stringtest", "value": "thisstringhere", "type": "string" }, { "id": "boolval", "value": true, "type": "boolean" }] }' http://localhost:8080/flowable-rest/service/form/form-data
```

You get this error:
```json
{
  "message": "Bad request",
  "exception": "JSON parse error: Cannot deserialize instance of `java.lang.String` out of START_OBJECT token; nested exception is com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize instance of `java.lang.String` out of START_OBJECT token\n at [Source: (PushbackInputStream); line: 1, column: 113] (through reference chain: org.flowable.rest.service.api.form.SubmitFormRequest[\"properties\"]->java.util.ArrayList[0]->org.flowable.rest.service.api.form.RestFormProperty[\"value\"])"
}
```

This PR makes sure the `form-data` endpoint can receive JSON Nodes and that the VariableTypes correctly recognizes the object as JSON Node. Also, as far as I could tell, every variable entered into the form-data got treated as a `String`, so a boolean would have been saved as `"true"`.

### Before
|Name|Type|Value|
|---|---|---|
|aBoolean|String|"true"|
|jsonVal|Serializable|(Binary)|

### After
|Name|Type|Value|
|---|---|---|
|aBoolean|boolean|true|
|jsonVal|json|{"test": true}|

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/FormService.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/FormService.java
@@ -48,7 +48,7 @@ public interface FormService {
     /**
      * Start a new process instance with the user data that was entered as properties in a start form.
      */
-    ProcessInstance submitStartFormData(String processDefinitionId, Map<String, String> properties);
+    ProcessInstance submitStartFormData(String processDefinitionId, Map<String, Object> properties);
 
     /**
      * Start a new process instance with the user data that was entered as properties in a start form.
@@ -64,7 +64,7 @@ public interface FormService {
      * @param properties
      *            the properties to pass, can be null.
      */
-    ProcessInstance submitStartFormData(String processDefinitionId, String businessKey, Map<String, String> properties);
+    ProcessInstance submitStartFormData(String processDefinitionId, String businessKey, Map<String, Object> properties);
 
     /**
      * Retrieves all data necessary for rendering a form to complete a task. This can be used to perform rendering of the forms outside of the process engine.
@@ -84,10 +84,10 @@ public interface FormService {
     /**
      * Completes a task with the user data that was entered as properties in a task form.
      */
-    void submitTaskFormData(String taskId, Map<String, String> properties);
+    void submitTaskFormData(String taskId, Map<String, Object> properties);
 
     /** Save the data that was entered as properties in a task form. */
-    void saveFormData(String taskId, Map<String, String> properties);
+    void saveFormData(String taskId, Map<String, Object> properties);
 
     /**
      * Retrieves a user defined reference to a start form.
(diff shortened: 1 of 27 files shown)


tijsrademakers:
@mattydebie can you make sure the unit tests run ok for the changes? You can use the check profile to do this in your local dev environment.

mattydebie:
@tijsrademakers Sure, I'll take a look!

mattydebie:
@tijsrademakers If this is OK for you and there is nothing to add, let me know and I'll squash the commits.

tijsrademakers:
@mattydebie yes the changes look fine

mattydebie:
The commits are squashed into one!

tijsrademakers:
@mattydebie sorry for not raising it before, but after looking at the changes again, I see this change in JsonType:

if (value == null || value instanceof LinkedHashMap) {

Using a LinkedHashMap here is not ok, it should be Map probably. But this changes the way Maps are handled and introduces a backwards compatibility issue for instances where a Map is expected instead of a JSON object. 
Could you add a new Variable type supporting the conversion from Maps, stored as JSON and transformed back to Maps when they are fetched again? And we need to makes this an option you can disable in the engine configuration to make sure we can be backwards compatible.

There are also a couple of cases where a toString() is done on the Object value without doing a null check first. It's probably a rare case when a form property value is null, but it would be good to be on the safe side for the toString() calls.

mattydebie:
> Could you add a new Variable type supporting the conversion from Maps, stored as JSON and transformed back to Maps when they are fetched again? And we need to makes this an option you can disable in the engine configuration to make sure we can be backwards compatible.

Weren't maps being handled as Serializable? If so, we could add a config flag that makes sure the json is being skipped?

tijsrademakers:
@mattydebie yes Maps are handled as Serializable indeed. Adding a config flag that makes sure the JSON variable type is skipped is also fine yes.

mattydebie:
@tijsrademakers the changes for the JsonType are ready, though i can't seem to find where I can implement the config flag. And what prefix do you want it to be? (I found the `FlowableAppProperties.java` but can't seem to get it to work).. Any pointers would be great :grin:

VeridianDynamics:
Hi @tijsrademakers 
I guess I've inherited this PR as well, can you let me know what needs to be done to make it acceptable for a merge?

Thanks!

NicoP-S:
Hi i'm new to flowable.
I am just trying to do a POST to form/form-data with some boolean properties.
Like @mattydebie  mentioned these values are converted to strings `"true"`. 

Is there a workaround for this issue. I can't believe this is not working via rest!??

tijsrademakers:
Closing this PR due to lack of activity. Sorry for the lack of communication on our end, please reopen the PR if you think that this functionality should still be added and we will prioritize it.

PR #1536, opened 2019-01-17
ELResolver tests variable update with historic variable update

martin-grofcik:
just additional tests


tijsrademakers:
Please add the necessary history level check before working with historic data

PR #1541, opened 2019-01-20
Delete outdated checkspring maven profile.

PascalSchumacher:


tijsrademakers:
This checks the basic spring modules, so good to keep for now

PR #1552, opened 2019-01-25
Fix gather completed flows function's bug : RuntimeDisplayJsonClientResource.java

yongkyun:
If model is using Parallel Gateway and execute, some sequences flow after parallel gateway are missing because of destination flow id check code.
Code just check next activity. I have modified check code that destination flow id is in activity list


tijsrademakers:
This only works with the new functionality that stores the taken sequence flows as well in the activity instance tables. A solution is needed that will work also for existing data without the sequence flow info being available.

PR #1570, opened 2019-02-02
Add MS-SQLServer support in flowable docker images

Elgarni:
Currently, when running any docker image and setting `SPRING_DATASOURCE_DRIVER-CLASS-NAME` to `com.microsoft.sqlserver.jdbc.SQLServerDriver`, the app encounters an exception:
```java
.....
.....
java.lang.IllegalStateException: Cannot load driver class: com.microsoft.sqlserver.jdbc.SQLServerDriver
```
This is expected, as only Postgres driver is present in the classpath. 

**This PR** adds SQL Server JDBC driver dependency  to the following modules:
- flowable-ui-admin
- flowable-ui-idm
- flowable-ui-modeler
- flowable-ui-task
- flowable-app-rest

The dependency was added to the build profiles : `docker` and `docker-deps`

### DIFF
--- a/modules/flowable-app-rest/pom.xml
+++ b/modules/flowable-app-rest/pom.xml
@@ -501,6 +501,12 @@
           <artifactId>postgresql</artifactId>
           <scope>compile</scope>
         </dependency>
+        <dependency>
+          <groupId>com.microsoft.sqlserver</groupId>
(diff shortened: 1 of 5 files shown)


tijsrademakers:
The docker images don't come preconfigured with all the different db drivers. We can't make an exception for a specific driver version of MSSQL

Issue #1575, opened 2019-02-05
Using the datastore rather than the file system for Content Storage


jmoraleda:
I have looked through the source and have found the way to configure the Content Store Engine to use the datasource that is being used to store everything else rather than the filesystem when using the content-api. Have I missed this? If not, this is a feature request. I am willing to help implement it the project at large would find it useful.

tijsrademakers:
The supported option is filesystem for the open source content engine. You could also look at flowable.com for the commercial options for Flowable, because there is support for database persistence in that version. Otherwise we welcome PRs for this feature of course.

billerby:
@tijsrademakers "A PR is welcome". It seems @jmoraleda made a PR. However its in open state since quite a long time and without any comments whatsoever. Will you look in to the Pull Request sometime soon and perhaps merge it into flowable?

Thanks!

tijsrademakers:
@billerby correct, but the PR is currently outdated and is also lacking some things that are mentioned in the PR comments. So right now some time needs to be invested to update the PR, before it can be considered for merging.

billerby:
@tijsrademakers Ahh, missed the comments... sorry then I fully understand.

billerby:
@tijsrademakers hmm, can't see those comments, or is the the travis_ci comments you are refering to?

tijsrademakers:
@billerby what I mean mostly is that the PR is currently out of date and can't be merged, but there are also some things to discuss like the table structure, how the unit tests run for all the supported databases etc.

PR #1585, opened 2019-02-09
Added database backed content storage to content engine.

jmoraleda:
I implemented the feature we discussed a few days ago at https://github.com/flowable/flowable-engine/issues/1575.

Now there are two content storage implementations in the content engine and which one is chosen depends on configuration. The alternatives are:

      flowable.content.storage.type=file

or

      flowable.content.storage.type=database

`file` is the default if the property is not given and uses the existing `SimpleFileSystemContentStorage`. So adding this feature will not break any existing deployment.

`database` will make the content engine use a new `SimpleDatabaseContentStorage` where the persisted data is stored in a new table in the database: `ACT_CO_STORAGE_ITEM`.

Implementation details: 

1. For `SimpleFileSystemContentStorage` I have renamed the properties: `flowable.content.storage.root-folder` and `flowable.content.storage.create-root` as `flowable.content.storage.fs.root-folder` and `flowable.content.storage.fs.create-root`. Using the older names still works, but will issue a deprecation warning in the logs.

1. No additional configuration beyond specifying `flowable.content.storage.type=database` is required to use the database backed content storage.

1. The new table in the database `ACT_CO_STORAGE_ITEM` uses liquibase and mybatis just as the rest of the flowable system, so it will be created automatically as needed. I have *not* updated the global scripts to add it, though.

1. `SimpleDatabaseContentStorage` does not implement the `getMetaData` method of the `ContentStorage` interface. This is the same that `SimpleFileSystemContentStorage` does. See note below.

-----------------

NOTE on the last comment: Evidently the `getMetaData` method of interface `ContentStorage` is not being used anywhere since there is no implementation for it.

In fact, if I understand the intention correctly, the method itself is not correctly specified in the interface, since it should take a `String id` parameter to retrieve the metadata for a given item, but as currently specified it receives no parameters.

I wonder if the `getMetaData` method of interface  should be fixed by adding a `String id` parameter or deleted from the interface altogether. Does it serve any purpose? Since all the MetaData is available in the `ContentItem`, and I see no need to store it redundantly in the `StorageItem`.

Thank you for a great project.

### DIFF
--- a/modules/flowable-content-engine/src/main/java/org/flowable/content/engine/ContentEngineConfiguration.java
+++ b/modules/flowable-content-engine/src/main/java/org/flowable/content/engine/ContentEngineConfiguration.java
@@ -48,6 +48,8 @@ import org.flowable.content.engine.impl.persistence.entity.TableDataManager;
 import org.flowable.content.engine.impl.persistence.entity.TableDataManagerImpl;
 import org.flowable.content.engine.impl.persistence.entity.data.ContentItemDataManager;
 import org.flowable.content.engine.impl.persistence.entity.data.impl.MybatisContentItemDataManager;
+import org.flowable.content.engine.impl.storage.db.SimpleDatabaseContentStorage;
+import org.flowable.content.engine.impl.storage.db.entity.StorageItemEntityManager;
 import org.flowable.content.engine.impl.storage.fs.SimpleFileSystemContentStorage;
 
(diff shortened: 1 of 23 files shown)


jmoraleda:
I made a mistake selecting the from branch in my git repo, so I am closing this pull request and opened a new one pointing to the correct branch here: https://github.com/flowable/flowable-engine/pull/1591

Issue #1586, opened 2019-02-10
SaveContentItemCmd in the content engine does not delete old content data when resetting content data for an item


jmoraleda:
The SaveContentItemCmd in the Content Engine
https://github.com/flowable/flowable-engine/blob/master/modules/flowable-content-engine/src/main/java/org/flowable/content/engine/impl/cmd/SaveContentItemCmd.java

does not delete old content data when invoked repeatedly for the same `contentItem` with `inputData` not null. This leads to the old data being leaked in the ContentStorage without any item owning it.

I will fix this by updating the data from the ContentStorage if it already existed rather than inserting new data and will submit a pull request.

tijsrademakers:
Issue is resolved

PR #1587, opened 2019-02-10
Fixed ContentStorage leak when changing the Data associated with an existing item in the Content Engine

jmoraleda:
Fixes bug https://github.com/flowable/flowable-engine/issues/1586

Problem: The SaveContentItemCmd in the Content Engine created a new entry in the `contentStorage` when changing an item's `contentData`. The old `contentData` remained as a leaked object in the `contentStorage`.

In this fix, when a `contentData` already exists in the `contentStorage` associated with the content item, the existing `contentData` object is updated rather than a new one being created.


tijsrademakers:
Thanks for the PR, manually merged it with https://github.com/flowable/flowable-engine/commit/59b50049b81a49a6a15934f556b4cf65c7582f4a

PR #1589, opened 2019-02-10
Add flowable-ui-admin, flowable-ui-idm, flowable-ui-modeler and flowa…

PascalSchumacher:
…ble-ui-task modules to distro profile, so that these are build and tested on travis.

In order to prevent accidental breakages like my pull request with the Mockito `2.24.0` update (which broke the ui apps tests).

To sure if this is your prefer another way to solve this?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1467,7 +1467,11 @@
 				<module>modules/flowable-jms-spring-executor</module>
 				<module>modules/flowable-crystalball</module>
 				<module>modules/flowable-ui-common</module>
+				<module>modules/flowable-ui-admin</module>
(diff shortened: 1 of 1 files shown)


PascalSchumacher:
Build currently fails because of the mockito update in: https://github.com/flowable/flowable-engine/commit/3ff0874f6715f065d233723f7de3f5889e905480

tijsrademakers:
The Travis build is not very stable lately and adding even more modules to the current build doesn't feel like the right approach. Maybe we can use a separate build that tests these UI modules instead?

PascalSchumacher:
I believe the build will be stable again after recent fixes, but let's wait and see if that is really the case.

PR #1596, opened 2019-02-12
Speed-up (travis) build by increasing the maximum memory for unit tes…

PascalSchumacher:
…ts to 3 GB.

Alternative approach to the one taken in https://github.com/flowable/flowable-engine/pull/1595


PascalSchumacher:
Maybe it is just the normal variance on travis, but the build of #1595 seems to be faster that the two build of this pull request.

PR #1605, opened 2019-02-14
flowable-ldap-configurator: Remove javax.annotation-api test dependen…

PascalSchumacher:
…cy, which is no longer necessary after the move to UnboundID LDAP for tests.


filiphr:
Thanks @PascalSchumacher, nice catch

PR #1621, opened 2019-02-17
Performance: use addAll()

dbmalkovsky:
When calling method in a loop (e.g. collection.add(x))
replace with calling a bulk method (e.g. collection.addAll(listOfX).


PascalSchumacher:
Is it even better to use to pass the collection as an argument to the constructor (which should be possible in most of the cases)?

dbmalkovsky:
@PascalSchumacher I think you are correct.  Once this gets merged I will make a second pass. Thanks.

PR #1624, opened 2019-02-18
DateFormat is not thread-safe and should not be used as a constant f…

PascalSchumacher:
…ield.


tijsrademakers:
Thanks for creating the PR. Instead of instantiating SimpleDateFormat every time we are now using DateTimeFormatter, which is threadsafe

PascalSchumacher:
Yes, that is a better solution. Thanks!

PR #1636, opened 2019-02-25
Update Spring Framework and Spring Security to version 5.2.0 and Spri…

PascalSchumacher:
…ng Boot to version 2.2.0

Just for fun. 😄 Compile works locally, lets see if the tests pass.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -13,19 +13,27 @@
 		<version>7</version>
 	</parent>
 
+	<repositories>
+		<repository>
+			<id>spring-snapshot</id>
+			<name>Spring Snapshot Repository</name>
+			<url>https://repo.spring.io/snapshot/</url>
+		</repository>
+	</repositories>
+
 	<packaging>pom</packaging>
 	<version>6.5.0-SNAPSHOT</version>
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<jdk.version>1.8</jdk.version>
-		<spring.framework.version>5.1.5.RELEASE</spring.framework.version>
-		<spring.security.version>5.1.4.RELEASE</spring.security.version>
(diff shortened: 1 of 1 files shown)


filiphr:
>Just for fun. 😄 Compile works locally, lets see if the tests pass.

First milestone in 2 weeks fun indeed 😄. 

Btw @PascalSchumacher you know that Github added this new [Draft Pull Request](https://github.blog/2019-02-14-introducing-draft-pull-requests/) feature? It might fit nicely with such PRs

PascalSchumacher:
@filiphr Thanks, when I last created a draft pull request it looked like travis ci/travis was not build it. Not sure if that is a feature, or if I missed something. Will try again some time.

I created this branch locally to check if any of the performance improvements in Spring Boot 2.2.0 have a noticeable effect when starting the flowable web apps (no measurable effect). I also wanted to check how much faster start-up is with enabled lazy-configuration (see: https://github.com/spring-projects/spring-boot/commit/3a4a431eaae114cf59f445618ccff0cd65c4b647) (around 10% faster, maybe the effect is so small because liquibase is run at start-up).

filiphr:
>@filiphr Thanks, when I last created a draft pull request it looked like travis ci/travis was not build it. Not sure if that is a feature, or if I missed something. Will try again some time.

OK didn't know that. 

>I created this branch locally to check if any of the performance improvements in Spring Boot 2.2.0 have a noticeable effect when starting the flowable web apps (no measurable effect)

We actually noticed some nice speed up with 2.1. Didn't know that there is something in 2.2 as well.

>I also wanted to check how much faster start-up is with enabled lazy-configuration (see: [spring-projects/spring-boot@3a4a431](https://github.com/spring-projects/spring-boot/commit/3a4a431eaae114cf59f445618ccff0cd65c4b647)) (around 10% faster, maybe the effect is so small because liquibase is run at start-up).

I think that the Flowable engines are initialized in any case so you don't see big improvement with the lazy initialization.

PR #1643, opened 2019-03-02
Adapt AsyncHistoryManager to use the new runtime ActivityInstance for getting activity start

### LINKED ISSUE #1637: Adapt AsynHistoryManager to use the new runtime ActivityInstance
When the runtime `ActivityInstance` were added there were new methods added to the `HistoryManager`. The current implementation of `recordActivityEnd(ActivityInstance activityInstance)` uses the old `getActivityStart` to get the activity start based on the execution and activityId. We should add a new method that would get the activity start based on the runtime activity instance id

thiyagu-7:
Fixes #1637

I have added an overloaded method for `getActivityStart` that takes the *runtimeActivityInstanceId* and finds the activity start object based on that. I have also refactored the common parts to a private method and used a Predicate logic to find the required activity start instance.

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/history/async/AsyncHistoryManager.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/history/async/AsyncHistoryManager.java
@@ -20,6 +20,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


filiphr:
@thiyagu-7 on first glance (over my phone) it looks good. I am back from vacation on Monday and will do the merge then. I also need to check the predicate logic, not sure if that might have some performance impact

PR #1648, opened 2019-03-06
master

javierruizflowable:


tijsrademakers:
Hi Javier,

Thanks for the PR.
Could you implement the case instance name query options in the same way as you see here:

https://github.com/flowable/flowable-engine/blob/master/modules/flowable-engine/src/main/java/org/flowable/engine/runtime/ProcessInstanceQuery.java#L278

This provides also the option to search on the exact name or do a like with case option. Then it's consistent with the other query API options.

tiffmaelite:
Hi @tijsrademakers 
As agreed, the options to search for exact name or to do a like with case will be added in a later pull-request. Furthermore, the variable name was changed from `caseInstanceName` to `caseInstanceNameLikeIgnoreCase` according to your request for consistency, and to avoid the need for renaming when once the other options are added.

tiffmaelite:
Hi @tijsrademakers 
As agreed privately, the options to search for exact name or to do a like with case will be added in a later pull-request. Furthermore, the variable name was changed from `caseInstanceName` to `caseInstanceNameLikeIgnoreCase` according to your request for consistency.

tiffmaelite:
Hi @tijsrademakers 
As agreed privately, the options to search for exact name or to do a like with case will be added in a later pull-request. Furthermore, the variable name was changed from `caseInstanceName` to `caseInstanceNameLikeIgnoreCase` according to your request for consistency.

PR #1660, opened 2019-03-12
add querying by formKey

### LINKED ISSUE #1642: Expose querying by formKey for the TaskInfoQuery
The `Task` and `HistoricTaskInstance` have the `formKey` persisted. It would be good if we expose a query option in the [`TaskInfoQuery`](https://github.com/flowable/flowable-engine/blob/master/modules/flowable-task-service-api/src/main/java/org/flowable/task/api/TaskInfoQuery.java) (that would allow it to query both in `Task` and `HistoricTaskInstance`.

LazarGrnarov:
Fixes #1642

### DIFF
--- a/modules/flowable-task-service/src/main/java/org/flowable/task/service/impl/HistoricTaskInstanceQueryImpl.java
+++ b/modules/flowable-task-service/src/main/java/org/flowable/task/service/impl/HistoricTaskInstanceQueryImpl.java
@@ -112,6 +112,7 @@ public class HistoricTaskInstanceQueryImpl extends AbstractVariableQueryImpl<His
     protected Date completedBeforeDate;
     protected String category;
     protected boolean withFormKey;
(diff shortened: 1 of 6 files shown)


modules/flowable-task-service/src/main/java/org/flowable/task/service/impl/HistoricTaskInstanceQueryImpl.java
```diff
@@ -1252,6 +1253,20 @@ public HistoricTaskInstanceQuery taskWithFormKey() {
         return this;
     }
 
+    @Override
+    public HistoricTaskInstanceQuery taskFormKey(String formKey) {
+        if(formKey == null){
```
dbmalkovsky: nit pick but the coding "standards" specify "if (" instead of "if(".  There are multiple occurrences in this PR.

PR #1668, opened 2019-03-18
Closing inputStream after process auto deployment.

hidertanure:
I was having some trouble with hot deployment on tomcat after integrate my application with Flowable.
I that case, Flowable was letting all my .bpmn20.xml files open, preventing the server to made the deployment. Needing to stop the server before make a new deploy.

### DIFF
--- a/modules/flowable-cmmn-engine/src/test/java/org/flowable/cmmn/test/async/AsyncTaskTest.java
+++ b/modules/flowable-cmmn-engine/src/test/java/org/flowable/cmmn/test/async/AsyncTaskTest.java
@@ -159,5 +159,44 @@ public class AsyncTaskTest extends FlowableCmmnTestCase {
         assertCaseInstanceEnded(caseInstance);
         assertEquals(0L, cmmnManagementService.createJobQuery().caseInstanceId(caseInstance.getId()).count());
     }
+
+    @Test
+    @CmmnDeployment
+    public void testAsyncExclusiveServiceTaskAfterStageExit() {
+        CaseInstance caseInstance = cmmnRuntimeService.createCaseInstanceBuilder().caseDefinitionKey("testAsyncExclusive").start();
+
+        // Trigger the user event listener first. This will execute the first exclusive async service task
+        cmmnRuntimeService.completeUserEventListenerInstance(cmmnRuntimeService.createUserEventListenerInstanceQuery()
+            .caseInstanceId(caseInstance.getId()).singleResult().getId());
+
+        // There should be one job and it needs to be exclusive
+        Job job = cmmnManagementService.createJobQuery().caseInstanceId(caseInstance.getId()).singleResult();
+        assertThat(job.isExclusive()).isTrue();
+
+        waitForJobExecutorToProcessAllJobs();
+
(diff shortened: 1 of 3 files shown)


hidertanure:
DeploymentBuilderImpl.java, this method public DeploymentBuilder addZipInputStream(ZipInputStream zipInputStream);

It is using the IoUtil.readInputStream expecting to keep the stream open. It is a wrong responsibility to the method. I sujest to create an another IoUtil.readInputStream to keep stream open explicity.

tijsrademakers:
Hi,

Thanks for the PR, I understand your point. But we already take care of closing the input streams when using the IoUtil.readInputStream method in most cases I've looked at. I think we need to find the reason why it wasn't closed properly in your usage and explicitly close it there instead.
Do you know where it didn't do that in your case?

hidertanure:
Hi, i glad to help.

In my case it was leaving open when i just put a .bpmn20.xml file in the process resource folder. When the deploy runs it leave the .xml file locked.

Thats is why i commited the code with the properly close on the read method like the others.

Looking on this test failure, it is using the read method in a wrong way. I sugest to change it, for example, creating a new Stream for the zipEntry to get the bytes.

Or change method name to express explicitly this responsability, to leave the stream open.

What do you think?

hidertanure:
Submitting corrections on "addZipInputStream" methods as suggested.

> I sugest to change it, for example, creating a new Stream for the zipEntry to get the bytes.

filiphr:
@hidertanure I would suggest that you inspect the call hierarchy of the method and appropriately close the streams there. One if the locations is the deployment strategies

hidertanure:
I totaly agree with you. I’ll work on it.

What i was talking about responsability of the method it’s clear now after my last change. 
In a first look it’s seeams like that the Util had to close the stream, but it is exactly the opposit. The InputStream is a external managed resource, different from what is happening in the others Util’s methods.

A question. The openJdk build error it’s something related to the environment, wright?

filiphr:
>What i was talking about responsability of the method it’s clear now after my last change.
In a first look it’s seeams like that the Util had to close the stream, but it is exactly the opposit. The InputStream is a external managed resource, different from what is happening in the others Util’s methods.

I don't think so. The read into bytes method should not close the stream. The other methods are not getting an input stream.

I would say that the current changes are not needed. We only need to make sure that the input stream is closed on the location where it is created.

hidertanure:
> > What i was talking about responsability of the method it’s clear now after my last change.
> > In a first look it’s seeams like that the Util had to close the stream, but it is exactly the opposit. The InputStream is a external managed resource, different from what is happening in the others Util’s methods.
> 
> I don't think so. The read into bytes method should not close the stream. The other methods are not getting an input stream.
> 
> I would say that the current changes are not needed. We only need to make sure that the input stream is closed on the location where it is created.

I didn't make myself clear, but what i mean is the same. The Util don’t need to be change. In my first look i was wrong.

filiphr:
>I didn't make myself clear, but what i mean is the same. The Util don’t need to be change. In my first look i was wrong.

OK, I didn't understand you well, sorry. 

Looking forward on the next commits then 😉

hidertanure:
Sorry guys, i guess i made a mess with my commits...
I didn't undo the previous commit properly.
What's the best approach, closing this request and do a new fork?

PR #1670, opened 2019-03-21
Adjusting DefaultAutoDeployment to close inputStream after deploy

hidertanure:
I was having some trouble with the hot deployment on Apache Tomcat after integrate my application with Flowable.

I that case, Flowable was letting all my resources/processes/*.bpmn20.xml files opened, preventing the Tomcat server to made the deployment. Needing to stop the server before make a new deploy.

After some advices here: [https://github.com/flowable/flowable-engine/pull/1668](https://github.com/flowable/flowable-engine/pull/1668)

I made this new request, to correct the deployment method and adjust  their tests.

### DIFF
--- a/modules/flowable-app-engine-spring/src/main/java/org/flowable/app/spring/autodeployment/DefaultAutoDeploymentStrategy.java
+++ b/modules/flowable-app-engine-spring/src/main/java/org/flowable/app/spring/autodeployment/DefaultAutoDeploymentStrategy.java
@@ -14,6 +14,7 @@
 package org.flowable.app.spring.autodeployment;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.util.zip.ZipInputStream;
 
 import org.flowable.app.api.AppRepositoryService;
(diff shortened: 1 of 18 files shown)


hidertanure:
> Thanks for doing the changes @hidertanure.
> 
> This looks much better, seems like we were not closing those properly on multiple places 😄. I have one last comment in the places where we create `ZipInputStream` I think that we should close that stream and that would close the underlying `InputStream` as well. I see that the close on the `ZipInputStream` does something extra, not just closing the underlying stream.

I guess it won't be a problem because the "try()" it's going to guarantee that original InputStream will be closed properly anyway at the end.
The ZipInputStream reuse the original InputStream.

filiphr:
This is what is there in the parent of the `ZipInputStream`.

```
public void close() throws IOException {
        if (!closed) {
            if (usesDefaultInflater)
                inf.end();
            in.close();
            closed = true;
        }
    }
```

As you can see it is doing something else, not just closing the wrapped `InputStream`. Honestly, I don't know that much about that inflater. Therefore, I would prefer to just close the `ZipInputStream` as well 😄

dbmalkovsky:
Nitpick:  I noticed that there are few classes where an import of a standard Java class was added (e.g., java.io.InputStream) out of the typical ordering of imports for the project.  Standard Java imports are typically first in the list.

tijsrademakers:
Closing PR due to inactivity

Issue #255, opened 2017-02-21
DmnEngine not work


jizhengdaling:
# hi，
My flowable.cfg.xml:
`<bean id="dmnEngineConfiguration" class="org.flowable.dmn.engine.impl.cfg.StandaloneDmnEngineConfiguration">
   <property name="jdbcUrl" value="jdbc:mysql://127.0.0.1:3306/flowable6" />
    <property name="jdbcDriver" value="com.mysql.jdbc.Driver" />
    <property name="jdbcUsername" value="root" />
    <property name="jdbcPassword" value="" />
</bean>`
I build DmnEngine like this:
` DmnEngineConfiguration dmnEngineConfiguration = DmnEngineConfiguration.createDmnEngineConfigurationFromInputStream(in);
		 dmnEngine = dmnEngineConfiguration.buildDmnEngine();
		 repositoryService=dmnEngine.getDmnRepositoryService();`

and then i use this code:
`@Test
	public void deploySingleDecision() {

		dmnRepositoryService
				.createDeployment()
				.name("secondDeployment")
				.addClasspathResource(
						"com/shareniu/flowablestudy/ch3/multiple_conclusions.dmn")
				.tenantId("testTenant").deploy();
	}`

its's wrong 
org.flowable.dmn.xml.exception.DmnXMLException: cvc-elt.1: 找不到元素 'definitions' 的声明。
	at org.flowable.dmn.xml.converter.DmnXMLConverter.convertToDmnModel(DmnXMLConverter.java:161) ~[flowable-dmn-xml-converter-6.0.0.jar:6.0.0]
	at org.flowable.dmn.engine.impl.parser.DmnParse.execute(DmnParse.java:80) ~[flowable-dmn-engine-6.0.0.jar:6.0.0]
	at org.flowable.dmn.engine.impl.deployer.ParsedDeploymentBuilder.createDmnParseFromResource(ParsedDeploymentBuilder.java:89) ~[flowable-dmn-engine-6.0.0.jar:6.0.0]

## multiple_conclusions.dmn 

`<?xml version="1.0" encoding="UTF-8"?>
<definitions xmlns="http://www.omg.org/spec/DMN/20151130" id="simple" name="Simple" namespace="http://activiti.org/dmn">
    <decision id="decision" name="Simple decision">
        <decisionTable id="decisionTable">
          <input>
            <inputExpression id="input1" typeRef="number">
                <text>input1</text>
            </inputExpression>
          </input>
[...]

PR #1700, opened 2019-04-10
Fix DmnXMLConverter exception: cvc-elt.1 (flowable#255)

ouyida3:
now when I visit http://www.omg.org/spec/DMN/20151101 , get 404 error.
Should I change it into http://www.omg.org/spec/DMN/20151101/dmn.xsd?

I get an exception cvc-elt.1 when I use DMN.
So I change the code:`flowable-engine/modules/flowable-dmn-xml-converter/src/main/resources/org/flowable/impl/dmn/parser/dmn.xsd`, and `mvn install` a new jar.
With this new jar, the exception disappeared.

### DIFF
--- a/modules/flowable-dmn-xml-converter/src/main/resources/org/flowable/impl/dmn/parser/dmn.xsd
+++ b/modules/flowable-dmn-xml-converter/src/main/resources/org/flowable/impl/dmn/parser/dmn.xsd
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<xsd:schema xmlns="http://www.omg.org/spec/DMN/20151101" xmlns:xsd="http://www.w3.org/2001/XMLSchema" targetNamespace="http://www.omg.org/spec/DMN/20151101" elementFormDefault="qualified">
+<xsd:schema xmlns="http://www.omg.org/spec/DMN/20151101/dmn.xsd" xmlns:xsd="http://www.w3.org/2001/XMLSchema" targetNamespace="http://www.omg.org/spec/DMN/20151101/dmn.xsd" elementFormDefault="qualified">
 	<xsd:element name="DMNElement" type="tDMNElement" abstract="true"/>
 	<xsd:complexType name="tDMNElement">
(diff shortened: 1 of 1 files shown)


tijsrademakers:
The XSD is the official DMN 1.1 XSD that is provided by the OMG specification website. This XSD can't be changed.

ouyida3:
> The XSD is the official DMN 1.1 XSD that is provided by the OMG specification website. This XSD can't be changed.

But the official DMN 1.1 XSD is different from `flowable-engine/modules/flowable-dmn-xml-converter/src/main/resources/org/flowable/impl/dmn/parser/dmn.xsd`. 
I visit [https://www.omg.org/spec/DMN/1.1](https://www.omg.org/spec/DMN/1.1), and find the XSD:
[https://www.omg.org/spec/DMN/20151101/dmn.xsd](https://www.omg.org/spec/DMN/20151101/dmn.xsd)

You will see `xsd:schema xmlns` is :
`http://www.omg.org/spec/DMN/20151101/dmn.xsd`, not `http://www.omg.org/spec/DMN/20151101`.
So I changed it.
After I changed it, the DMN errer(cvc-elt.1 ) disappeared.

I commented the `cvc-elt.1` Exception details in flowable#255.
https://github.com/flowable/flowable-engine/issues/255

tijsrademakers:
You are right, we are looking into why this is different.

yvoswillens:
Hi @ouyida3 
I just pushed a change to makes the new final 1.2 DMN spec the default when generating DMN XMLs.
Parsing support was already implemented.
This will make use of the 1.2 xmlns "http://www.omg.org/spec/DMN/20180521/MODEL/" from now on.

Regards,

Yvo

ouyida3:
hi, @yvoswillens ,I just tested it.Test success. Thank you very much :)
At the same time, I found that my "xxx.dmn" file
`<definitions xmlns="http://www.omg.org/spec/DMN/20151101/dmn.xsd" ...` runs successfully today, while it failed the day before yesterday.
Maybe OMG changed something these two days?
Anyway thank u very much!

PR #1701, opened 2019-04-10
Flowable-modeller: Display Description of BPMN Elements as Tooltips and some bug fix

imic07:
Flowable-modeller: Display Description of BPMN Elements as Tooltips


tijsrademakers:
thanks for the PR

PR #1705, opened 2019-04-12
Add BPMN and CMMN Elements i18n

imic07:


imic07:
Note: This is not true internationalization which should not affect the normal use of the program but only internationalize the text.

imic07:
Note: This is not really internationalization. Full internationalization should not affect the normal use and upgrading of procedures.

imic07:
there's another way to implement it, so i close this PR.

PR #1715, opened 2019-04-23
Fixed support for includeCaseVariables when retrieving historic cases

paulstapleton:
I have noticed that the Rest API /cmmn-api/cmmn-history/historic-case-instances?includeCaseVariables=true is not working.

The result is returned without any case variables being included - the parameter is being ignored.
It seems that the implementation for the parameter includeCaseVariables is missing. I have added the missing logic, by repeating the same code from the historic-process-instances Rest API.

Please let me know if anything else is needed.

Thanks,
Paul


tijsrademakers:
Hi Paul,

Thanks for the PR. It would be great if you could add a couple of unit tests to show that it's working as expected.

Thanks!

paulstapleton:
I've added a unit test for this now. I also added the same unit test for historic process variables as it also was missing unit testing.

Please let me know if any further changes are needed.

tijsrademakers:
thanks!

PR #1721, opened 2019-04-29
Update commons-lang to version 3.9.

PascalSchumacher:


PascalSchumacher:
Builds on Java 11/12 failed with:

>[ERROR] Failed to execute goal on project flowable-cmmn-engine-configurator: Could not resolve dependencies for project org.flowable:flowable-cmmn-engine-configurator:jar:6.5.0-SNAPSHOT: Failure to find org.flowable:flowable-form-engine-configurator:jar:6.5.0-SNAPSHOT in https://oss.sonatype.org/content/repositories/snapshots/ was cached in the local repository, resolution will not be reattempted until the update interval of sonatype-snapshots has elapsed or updates are forced

This failure also occurred for the last two builds of master.

PR #1723, opened 2019-05-02
Fix fetching of the new user after creation

VeridianDynamics:
I omitted to change the way newly-created users are fetched in the previous PR #1712, so this is a little bugfix. Let's fetch them by the ID (necessarily unique because of DB constraints) instead of email. Presently the code is throwing an error if there are multiple service accounts without an email, because the query returns more than a single result.


tijsrademakers:
Thanks for the PR

PR #1724, opened 2019-05-04
Add chinese document

qiudaoke:


tijsrademakers:
Thanks for the PR and the effort put into this.
The difficulty will always be to keep it in sync with the English documentation. Will you help to keep it in sync? :-)

qiudaoke:
@tijsrademakers  In China, there are probably more than 1,000 Flowable developers (fans), and I have set up a document synchronization team with about 20 members. We will synchronize and compare the official documents in real time. In the early days, many Flowable developers may not be familiar with this process. In order to avoid unnecessary errors, I will help to keep it in sync in the early stage. Afterwards I can lead everyone to attribute the entire PR process. Thank you

tijsrademakers:
@qiudaoke That's great to hear and thanks again for all your efforts.

PR #1734, opened 2019-05-10
Add support for unaqcuiring jobs on AsyncExecutor shutdown

enchobelezirev:
Relates to this forum post: https://forum.flowable.org/t/unacquire-jobs-when-asyncexecutor-shutdown-method-is-called/3855

Also, found that the unacquire method is called on the start of the AsyncExecutor. This PR, makes it to be called on the shutdown as well.

### DIFF
--- a/modules/flowable-job-service/src/main/java/org/flowable/job/service/impl/asyncexecutor/DefaultAsyncJobExecutor.java
+++ b/modules/flowable-job-service/src/main/java/org/flowable/job/service/impl/asyncexecutor/DefaultAsyncJobExecutor.java
@@ -35,22 +35,22 @@ import org.slf4j.LoggerFactory;
 public class DefaultAsyncJobExecutor extends AbstractAsyncExecutor {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(DefaultAsyncJobExecutor.class);
-    
+
     /**
      * If true (default), the thread for acquiring async jobs will be started.
      */
     protected boolean isAsyncJobAcquisitionEnabled = true;
-    
+
     /**
      * If true (default), the thread for acquiring timer jobs will be started.
      */
     protected boolean isTimerJobAcquisitionEnabled = true;
-    
+
     /**
      * If true (default), the thread for acquiring expired jobs will be started.
      */
     protected boolean isResetExpiredJobEnabled = true;
-    
+
     /**
      * Thread responsible for async job acquisition.
      */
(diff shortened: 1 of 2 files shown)


enchobelezirev:
I have used the latest formatter included in the ide-settings folder :)

enchobelezirev:
Hello @tijsrademakers,

Thanks for merging the PR. I have one question, though.
Why have you removed the if from the method?
In our case we do not have a tenantId and thus there is an Exception when using the method. 
In my opinion, the if check is needed there.

Best regards,
Encho

enchobelezirev:
Here is the method from the JobQuery:
```
@Override
    public JobQuery jobTenantId(String tenantId) {
        if (tenantId == null) {
            throw new FlowableIllegalArgumentException("Provided tenant id is null");
        }
        this.tenantId = tenantId;
        return this;
    }
```

enchobelezirev:
Moreover, the tenantId will always be null as the method unlockOwnedJobs is calling the UnacquireOwnedJobsCmd with tenantId = null:
`protected void unlockOwnedJobs() {
        jobServiceConfiguration.getCommandExecutor().execute(new UnacquireOwnedJobsCmd(lockOwner, null));
    }`

jbarrez:
The tenantId, when not set is defaulting to the empty string, see https://github.com/flowable/flowable-engine/blob/master/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/AbstractEngineConfiguration.java#L266 which references https://github.com/flowable/flowable-engine/blob/master/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/AbstractEngineConfiguration.java#L92

Do you have custom changes that make this tenantId null?

enchobelezirev:
No, but the problem is not that the tenantId is null in the AbstractEngineConfiguration, the problem is that the unlockOwnedJobs method in the AsyncExecutor is calling the UnacquireOwnedJobsCmd with tenantId = null. Please, see:
https://github.com/flowable/flowable-engine/blob/master/modules/flowable-job-service/src/main/java/org/flowable/job/service/impl/asyncexecutor/AbstractAsyncExecutor.java#L94

jbarrez:
Ok, thanks. I understand the problem now. I'll apply a fix + add a test for this.

enchobelezirev:
Hey @jbarrez, 
Is there any update on the fix in the UnacquireOwnedJobsCmd ?

jbarrez:
@enchobelezirev : it seems I added it long time ago, but never pushed it. Apologies!
I'll merge your PR instead, thanks!

enchobelezirev:
I thank you, too!

PR #1758, opened 2019-05-22
Http task configuration through spring boot, config to use system properties

reva:
Motivation was to be able to set the 'useSystemProperties' setting on the http client builder. This enables the configuration of proxy settings through system properties (amongst other things).


filiphr:
Thanks for the PR @reva. I've merged this into master and also updated the documentation with the new properties in https://github.com/flowable/flowable-engine/commit/7473f55b963fbb22adf22dc092eea9a62a2f8e05.

Pinging @qiudaoke just for heads up for the new documentation properties, for the Chinese translation 😄

PR #1761, opened 2019-05-23
Ability to update Business Key on cases

Stoeoeoe:
Add ability to update business key in cases


modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/history/async/AsyncCmmnHistoryManager.java
```diff
@@ -88,7 +88,20 @@ public void recordUpdateCaseInstanceName(CaseInstanceEntity caseInstanceEntity,
...
         }
     }
-    
+
+    @Override
+    public void recordUpdateBusinessKey(CaseInstanceEntity caseInstanceEntity, String businessKey) {
+        if (cmmnEngineConfiguration.getHistoryLevel().isAtLeast(HistoryLevel.ACTIVITY)) {
+            ObjectNode data = cmmnEngineConfiguration.getObjectMapper().createObjectNode();
+            putIfNotNull(data, CmmnAsyncHistoryConstants.FIELD_ID, caseInstanceEntity.getId());
+            putIfNotNull(data, CmmnAsyncHistoryConstants.FIELD_NAME, caseInstanceEntity.getBusinessKey());
```
filiphr: This should be `CmmnAsyncHistoryConstants.FIELD_BUSINESS_KEY` instead of `FIELD_NAME`. You can also add a test to `AsyncCmmnHistoryTest` to validate this as well

tijsrademakers:
Thanks for the PR

PR #1766, opened 2019-05-26
Add a property description at the beginning of flowable.http

qiudaoke:


filiphr:
Thanks for the quick update @qiudaoke 😄

PR #1800, opened 2019-06-10
Remove unused private fields and local variables in rest components t…

PascalSchumacher:
…ests.


tijsrademakers:
Hi @PascalSchumacher , is this change https://github.com/flowable/flowable-engine/pull/1800/files#diff-5e7e7b890116cceeb0f845f038bf0d93L247 ok? Did you check that it's not needed anymore?

PascalSchumacher:
Hi Tjis,

the test still passes as the private method is unused, so I think it is o.k.? Or am I missing something?

tijsrademakers:
ok thanks, just wanted to check quickly

Issue #1805, opened 2019-06-12
Engine: zero loop multi-instance with event subscription throws PersistenceException


mwerlitz:
When using event subscriptions (boundary message event) on multi-instance activities (subprocesses, usertasks, ...) and running through the activity with a loop cardinality of zero or an empty collection Flowable throws an PersistenceException.

It looks like the engine tries to insert the event subscription into the database after the execution already continued "skipping the activity" and the original execution does not exist anymore or wasn't inserted at all.

Tested with Flowable 6.4.1

Example Stacktrace:
```
org.apache.ibatis.exceptions.PersistenceException: 
### Error updating database.  Cause: org.h2.jdbc.JdbcSQLException: Referential integrity constraint violation: "ACT_FK_EVENT_EXEC: PUBLIC.ACT_RU_EVENT_SUBSCR FOREIGN KEY(EXECUTION_ID_) REFERENCES PUBLIC.ACT_RU_EXECUTION(ID_) ('9')"
Referential integrity constraint violation: "ACT_FK_EVENT_EXEC: PUBLIC.ACT_RU_EVENT_SUBSCR FOREIGN KEY(EXECUTION_ID_) REFERENCES PUBLIC.ACT_RU_EXECUTION(ID_) ('9')"; SQL statement:
insert into ACT_RU_EVENT_SUBSCR (
            ID_, 
            REV_,
            EVENT_TYPE_,
           	EVENT_NAME_,
           	EXECUTION_ID_,
           	PROC_INST_ID_,
           	ACTIVITY_ID_,
           	CONFIGURATION_,
           	CREATED_,
           	PROC_DEF_ID_,
           	TENANT_ID_
    )
    values (?,
            1,
            'message',
            ?,
            ?,
            ?,
            ?,
            ?,
            ?,
            ?,
            ?
    ) [23506-197]
### The error may involve org.flowable.engine.impl.persistence.entity.EventSubscriptionEntityImpl.insertMessageEventSubscription-Inline
### The error occurred while setting parameters
### SQL: insert into ACT_RU_EVENT_SUBSCR (             ID_,              REV_,             EVENT_TYPE_,             EVENT_NAME_,             EXECUTION_ID_,             PROC_INST_ID_,             ACTIVITY_ID_,             CONFIGURATION_,             CREATED_,             PROC_DEF_ID_,             TENANT_ID_     )     values (?,             1,             'message',             ?,             ?,             ?,             ?,             ?,             ?,             ?,             ?     )
[...]

PR #1807, opened 2019-06-12
Bug 1805: added two test cases for zero loop multi-instances with event subscription

mwerlitz:
The test case shows the problem described in https://github.com/flowable/flowable-engine/issues/1805


tijsrademakers:
Merged PR manually, thanks for the unit tests, this made it easy to fix the issue.

Issue #1806, opened 2019-06-12
Engine: JPA entity proxy object cannot be used as a process variable


mwerlitz:
When using the JPA integration it is possible to use the entities as process variables e.g. in a multi-instance loop. The JPA provider may return a proxy object for the real entity e.g. for lazy loading in one-to-may collections.

Flowables way to get the Id of the entity object via reflection (at least for field access) for storing it as a process variable in `JPAEntityMappings ` is broken when proxy objects are used. Tested with Hibernate 5.3.10 bundled with Spring Boot 2.1.5.

Tested with Flowable 6.4.1

Example Exception:
```
org.flowable.common.engine.api.FlowableIllegalArgumentException: Value of primary key for JPA-Entity cannot be null
at org.flowable.variable.service.impl.types.JPAEntityMappings.getIdString(JPAEntityMappings.java:168) ~[flowable-variable-service-6.4.1.jar:6.4.1]
at org.flowable.variable.service.impl.types.JPAEntityMappings.getJPAIdString(JPAEntityMappings.java:87) ~[flowable-variable-service-6.4.1.jar:6.4.1]
at org.flowable.variable.service.impl.types.JPAEntityVariableType.setValue(JPAEntityVariableType.java:72) ~[flowable-variable-service-6.4.1.jar:6.4.1]
at org.flowable.variable.service.impl.persistence.entity.VariableInstanceEntityImpl.setValue(VariableInstanceEntityImpl.java:139) ~[flowable-variable-service-6.4.1.jar:6.4.1]
at org.flowable.variable.service.impl.persistence.entity.VariableInstanceEntityManagerImpl.create(VariableInstanceEntityManagerImpl.java:50) ~[flowable-variable-service-6.4.1.jar:6.4.1]
at org.flowable.variable.service.impl.persistence.entity.VariableScopeImpl.createVariableInstance(VariableScopeImpl.java:887) ~[flowable-variable-service-6.4.1.jar:6.4.1]
at org.flowable.engine.impl.persistence.entity.ExecutionEntityImpl.createVariableInstance(ExecutionEntityImpl.java:720) ~[flowable-engine-6.4.1.jar:6.4.1]
at org.flowable.engine.impl.persistence.entity.ExecutionEntityImpl.createVariableLocal(ExecutionEntityImpl.java:741) ~[flowable-engine-6.4.1.jar:6.4.1]
[...]

PR #1808, opened 2019-06-13
Bug 1806: added test case for problem getting jpa entity id via reflection from entity proxy object

mwerlitz:
The test case shows the problem described in https://github.com/flowable/flowable-engine/issues/1806


mwerlitz:
Replaced with bugfix propasal.

PR #1809, opened 2019-06-13
Bug 1806: testcase and bugfix proposal

mwerlitz:
Changed getting the JPA entity identifier from reflection to JPA 2.0 API.
This way is compatible with entity proxy objects.

### DIFF
--- a/modules/flowable-engine/src/test/java/org/flowable/standalone/jpa/JPAVariableTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/standalone/jpa/JPAVariableTest.java
@@ -591,4 +591,27 @@ public class JPAVariableTest extends ResourceFlowableTestCase {
         assertTrue(updatedEntity instanceof FieldAccessJPAEntity);
         assertEquals("updatedValue", ((FieldAccessJPAEntity) updatedEntity).getValue());
     }
+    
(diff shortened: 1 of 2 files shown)


tijsrademakers:
Hi,

Thanks for the PR. If this change runs fine with our test suite then it looks fine for me. Can we set it to ready for review, so travis will run all tests for this PR?

Thanks,

Tijs

mwerlitz:
Hi Tijs,

two things to note about the fix:
- After the change Flowable will require a JPA 2.0 provider. If you need to support JPA 1.0 also we will need a version detection to fall back to the original implementation.
- The proposed fix does not modify the Flowable5 Engine (for compatibility)

tijsrademakers:
Closing the PR. For this to be added a choice between a JPA 1.0 and 2.0 needs to be provided so that a Flowable user can choose how to run the JPA integration.

mwerlitz:
I don't think just closing the PR for a bug that breaks current JPA integration is not an appropriate solution.

I could add a JPA version detection so there is no need for the user to explicitly decide between JPA 1.0 and 2.0 integration.
Furthermore as I remember correctly your build currently uses JPA 2.0 libs so there is no real JPA 1.0 compatiability anyway.

PR #1812, opened 2019-06-17
flowable-ui-idm-app, flowable-ui-modeler-app and flowable-ui-task-app…

PascalSchumacher:
… do not use the servlet api directly so remove their javax.servlet-api dependencies.


PascalSchumacher:
Noticed that there are other unnecessary direct dependencies. Closing this in favor of a new pull request.

Issue #1822, opened 2019-06-21
async for throw events


martin-grofcik:
intermediate signal throw event does not support async export/import.
+ intermediate none throw event async attribute

tijsrademakers:
Manually merged PR

PR #1823, opened 2019-06-21
#1822 allowing async flag for intermediate none throw event + signal …

martin-grofcik:
…throw event async flag export/import fix


tijsrademakers:
Merged manually

Issue #1824, opened 2019-06-21
No way to Query Completed Historical Tasks


unamanic:
Currently, tasks that get completed (as opposed to deleted) are left with a null deleteReason. However, attempting to use the API to query for a null does not add the criteria to the query.

unamanic:
PR merged, closing issue.

PR #1825, opened 2019-06-21
#1824 - Add "withoutDeleteReason" to query for null delete reasons.

unamanic:
Fix for issue #1824


unamanic:
Added a test and corrected a bug in the when in an or condition.

PR #1847, opened 2019-07-17
Added indexes which fix slow-performing queries

VeridianDynamics:
Hi,
we run Flowable in a decently large multi-tenant setup and noticed some long-running queries (some of them > 1min) during a load test. Adding the indexes in this PR fixed the issue. It was quite difficult for me to figure out how the DB creation and upgrade is organized, so it may need some additional work. There's only one other PR ( https://github.com/flowable/flowable-engine/pull/429 ) with indexes, so I used it as a guide and tried to follow the current naming conventions.

Regards,
Jan


tijsrademakers:
Closing this PR. These indices make sense in certain use cases when queries with these columns are executed. We don't see these indices as general improvements for all users of the Flowable engines.

Issue #1859, opened 2019-07-26
ModelImageService - Error creating thumbnail


dgkm:
Version: 6.5.0

> I'm getting the following exception when modeler creating a thumbnail upon saving the BPMN diagram.  Is it a known issue and/or a bug?

```2019-07-26 00:22:36.549 ERROR 1 --- [nio-8080-exec-8] o.f.u.m.s.ModelImageService              : Error creating thumbnail image 04019394-9f46-11e9-b69a-0242c0a81003
2019-07-26T00:22:36.562306377Z 
java.lang.NullPointerException: null
at sun.awt.FontConfiguration.getVersion(Unknown Source) ~[?:1.8.0_212]
at sun.awt.FontConfiguration.readFontConfigFile(Unknown Source) ~[?:1.8.0_212]
at sun.awt.FontConfiguration.init(Unknown Source) ~[?:1.8.0_212]
at sun.awt.X11FontManager.createFontConfiguration(Unknown Source) ~[?:1.8.0_212]
at sun.font.SunFontManager$2.run(Unknown Source) ~[?:1.8.0_212]
at java.security.AccessController.doPrivileged(Native Method) ~[?:1.8.0_212]
at sun.font.SunFontManager.<init>(Unknown Source) ~[?:1.8.0_212]
at sun.awt.FcFontManager.<init>(Unknown Source) ~[?:1.8.0_212]
at sun.awt.X11FontManager.<init>(Unknown Source) ~[?:1.8.0_212]
at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method) ~[?:1.8.0_212]
at sun.reflect.NativeConstructorAccessorImpl.newInstance(Unknown Source) ~[?:1.8.0_212]
at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(Unknown Source) ~[?:1.8.0_212]
at java.lang.reflect.Constructor.newInstance(Unknown Source) ~[?:1.8.0_212]
at java.lang.Class.newInstance(Unknown Source) ~[?:1.8.0_212]
at sun.font.FontManagerFactory$1.run(Unknown Source) ~[?:1.8.0_212]
at java.security.AccessController.doPrivileged(Native Method) ~[?:1.8.0_212]
at sun.font.FontManagerFactory.getInstance(Unknown Source) ~[?:1.8.0_212]
at sun.font.SunFontManager.getInstance(Unknown Source) ~[?:1.8.0_212]
[...]

Issue #1856, opened 2019-07-23
Javascript engine not found in docker image 6.4.2


robrez:
Found similar issues were reported in the past.  #1286 stated to have fixed the problem via d1a84e0; however, it looks like a regression has landed in v6.4.2 since the base image was changed on  3425b90

```
org.flowable.common.engine.api.FlowableException: Can't find scripting engine for 'javascript'
	at org.flowable.common.engine.impl.scripting.ScriptingEngines.getEngineByName(ScriptingEngines.java:127) 
```

erbg:
Also the single Docker Images are based on alpine-slim (https://github.com/flowable/flowable-engine/blob/master/modules/flowable-app-rest/src/main/docker/Dockerfile) for instance.

yvoswillens:
We're investigating a fix for the issue. This will apply for all images.

erbg:
Is there any work around we could use in the meantime?

unamanic:
Despite the name, the PR updates all of the Dockerfiles to use the full Alpine openJDK 11 base and adds fonts (which caused another issue).  Since it’s using the full base instead of the slim, it will be bigger, but They do work.

yvoswillens:
New docker images (6.4.2.1) have been published on Docker Hub.
Code changes present on master.

PR #1860, opened 2019-07-27
Update All-in-One Dockerfile

unamanic:
Corrects Issue #1859 and #1856 in all-in-one container, but migrating to the regular alpine base image and adding fonts.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen.
+
+**Describe alternatives you've considered**
+A clear and concise description of any alternative solutions or features you've considered.
+
+**Additional context**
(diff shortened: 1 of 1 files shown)


filiphr:
Awesome work @unamanic. Just thinking out loud now. Why not switch to AdoptOpenJDK 11 docker image?

unamanic:
@filiphr I don't see any reason not to, since JDK 11 is the next LTS release.  However this fix should probably get back-ported to 6.4.2, since the published images contain both of these issues.

unamanic:
Updated Docker image to use  AdoptOpenJDK 11

PascalSchumacher:
I guess this pull request was made redundant by https://github.com/flowable/flowable-engine/commit/0d5e6309f2fda95c60d298ef8cd0bc69bdcd4e66?

yvoswillens:
We're currently testing the referred commit. It's a bit different than this PR. 
When it's verified I will comment on this and close it.

unamanic:
I’m not certain the other pull request solves the issue with JavaScript script tasks, since Nashorn is one of the things pulled from the slim image.

yvoswillens:
for the jdk8 alpine slim that was the case; for the jdk11 alpine slim not.

bwangfdu:
I find current adoptopenjdk/openjdk11:alpine seems not working with Flowable-rest-app 6.4.2 release version. When I start the jar file with "java -jar", it exit with "segmentation fault". The alpine base image and adding openjdk11 with apk works fine.

unamanic:
This Pull request no longer required.

PR #1868, opened 2019-07-30
Remove deprecated description field from @Api()

dbmalkovsky:
Description is not used by Swagger doc and is deprecated.
Also fixed up a some imports to match standard ordering.

The deprecated description argument is now removed from all files.


dbmalkovsky:
I will look at doing this again in the near future; closing for now.

PR #1883, opened 2019-08-06
Speed up test execution by defining the "javax.xml.stream.XMLInputFac…

PascalSchumacher:
…tory" and "javax.xml.validation.SchemaFactory:http://www.w3.org/2001/XMLSchema" system properties to avoid expensive lookup of implemention classes at factory creation time.


PascalSchumacher:
Not sure if this has a noticeable effect.

tijsrademakers:
Didn't have an effect on the test build time from what I can see, so closing it for now

Issue #1909, opened 2019-08-23
Skip Expression Not Implemented in Http Task


unamanic:
**Describe the bug**
The "Http Task" is a "Service Task" an should provide "Skip Expression" functionality.  This is currently not implemented.  If you define an Http Task's skipExpression to an expression that evaluates to true and set `_ACTIVITI_SKIP_EXPRESSION_ENABLED` to true, the Http Task gets executed.
 
**Expected behavior**
If you set an Http Task's skipExpression to an expression that evaluates to true and set `_ACTIVITI_SKIP_EXPRESSION_ENABLED` to true, the task should be skipped.

Note: the code on Service Tasks allows the use of `_FLOWABLE_SKIP_EXPRESSION_ENABLED` in leu of `_ACTIVITI_SKIP_EXPRESSION_ENABLED`

tijsrademakers:
Merged the PR, thanks

PR #1910, opened 2019-08-25
Add support for Skip Expressions to HTTP Tasks #1909

unamanic:
Adds support for Skip Expressions to BPMN HTTP Tasks during to the flowable-http module and Modeler.

#### Check List:
* Unit tests: YES
* Documentation: NO (Skip Expression not documented in other tasks that use them)


modules/flowable-http/src/main/java/org/flowable/http/bpmn/impl/HttpActivityBehaviorImpl.java
```diff
@@ -141,84 +143,95 @@ public boolean verify(String s, SSLSession sslSession) {
...
-            }
+        boolean isSkipExpressionEnabled = SkipExpressionUtil.isSkipExpressionEnabled(skipExpressionText,
+                execution.getCurrentActivityId(), execution, commandContext);
 
-        } catch (Exception e) {
-            if (e instanceof FlowableException) {
-                throw (FlowableException) e;
-            } else {
-                throw new FlowableException(HTTP_TASK_REQUEST_FIELD_INVALID + " in execution " + execution.getId(), e);
+        if (skipExpressionText == null || !isSkipExpressionEnabled || !SkipExpressionUtil.shouldSkipFlowElement(skipExpressionText,
```
tijsrademakers: skipExpressionText == null check is not necessary here, can you remove this?

modules/flowable-http/src/test/java/org/flowable/http/bpmn/HttpServiceTaskTest.java
```diff
@@ -497,6 +500,31 @@ public void testDeleteResponseEmptyBody() {
         assertNull(runtimeService.getVariable(processInstanceId, "myResponse"));
     }
 
+    @Test
+    @Deployment
+    public void testGetWithVariableNameAndSkipExpression() {
+        Map<String, Object> variables = new HashMap<>();
+        variables.put("_ACTIVITI_SKIP_EXPRESSION_ENABLED", true);
```
tijsrademakers: can you change this to _FLOWABLE_SKIP_EXPRESSION_ENABLED, and the same for the other lines using this variable?

tijsrademakers:
Thanks for the PR, I've added two small comments, could you have a look and update the PR?

modules/flowable-http/src/main/java/org/flowable/http/bpmn/impl/HttpActivityBehaviorImpl.java
unamanic: Removed redundant null check and updated test to ensure that Http Tasks without a skip expression are still executed when _FLOWABLE_SKIP_EXPRESSION_ENABLED is set.

modules/flowable-http/src/test/java/org/flowable/http/bpmn/HttpServiceTaskTest.java
unamanic: Updated

PR #1912, opened 2019-08-31
Update Awaitility to version 4.0.0.

PascalSchumacher:
Code changes are not required. Imho using `java.time.Duration` is more readable.


filiphr:
Thanks @PascalSchumacher.

>Code changes are not required. Imho using `java.time.Duration` is more readable.

I agree that `Duration` is more readable. I also prefer using `java.time` as much as possible.

Issue #1919, opened 2019-09-05
Base64 Encoding in Expression


erbg:
Now that in 6.4 you added the possibility of expression Functions it would be very nice to have also a function that converts a Variable to Base64.

https://www.flowable.org/docs/userguide/index.html#expressionsFunctions

This is very usefull in Rest Http Tasks for example.

filiphr:
Done in https://github.com/flowable/flowable-engine/pull/1924

PR #1924, opened 2019-09-07
Base64 Encoding in Expression

erbg:
#### Check List:
* Unit tests: YES 
* Documentation: YES 

#1919 

Hope this is ok for you.


filiphr:
@erbg from an API point of view does it make sense to add a `base64Decode` function as well? That way you use your variable in the process as well.

erbg:
@filiphr I did not added this, because i don't have seen a use case. Normaly you need to encode to Base64 in Rest Requests, but when do you need the other round?

If you prefer i will add it. Tell me.

filiphr:
@erbg I don't think there might be a big use case, but was thinking to have a symmetry.

Perhaps a better alternative would be to have to functions under the `base64` prefix that would be able to do encoding and decoding. 

One use case could be when you want to decode and save variable value.

PR #1943, opened 2019-09-16
Update non-static non-final fields with all upper-case names.

dbmalkovsky:
Added `static` or `final` as needed.


modules/flowable-engine/src/main/java/org/flowable/engine/impl/persistence/entity/CommentEntityImpl.java
```diff
@@ -56,7 +56,7 @@ public void setFullMessageBytes(byte[] fullMessageBytes) {
         fullMessage = (fullMessageBytes != null ? new String(fullMessageBytes) : null);
     }
 
-    public static String MESSAGE_PARTS_MARKER = "_|_";
+    public static final String MESSAGE_PARTS_MARKER = "_|_";
     public static Pattern MESSAGE_PARTS_MARKER_REGEX = Pattern.compile("_\\|_");
```
PascalSchumacher: I guess this should be final too?

dbmalkovsky: Yes; good catch.

PR #1947, opened 2019-09-20
Update Jackson to version 2.10.1

PascalSchumacher:
Jackson 2.10 was just released and will be used by Spring 5.2 and Spring Boot 2.2.

### DIFF
--- a/modules/flowable-http/src/main/java/org/flowable/http/HttpActivityExecutor.java
+++ b/modules/flowable-http/src/main/java/org/flowable/http/HttpActivityExecutor.java
@@ -53,6 +53,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.MissingNode;
 
 /**
  * An executor behavior for HTTP requests.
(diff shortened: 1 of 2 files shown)


PascalSchumacher:
I guess the change to `HttpServiceTaskTest#testDeleteResponseEmptyBody` is not correct. Probably the implementation has to be changed instead to keep the old behavior.

The reason for the behavior change should be one  of the Null related changes listed in https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.10 (maybe https://github.com/FasterXML/jackson-databind/issues/2430).

filiphr:
It probably is that. However, I think that we should fix it in the implementation side not in the test.

Perhaps in `https://github.com/flowable/flowable-engine/blob/c0f05e8c682ded61035a5076ff66ad97524772aa/modules/flowable-http/src/main/java/org/flowable/http/HttpActivityExecutor.java#L118` and check if the return type is `NullNode`.

Or adapt `JsonType`

PascalSchumacher:
Thanks Phillip!

I have updated the pull request and changed the implementation instead of the test (not sure if this is the best solution).

The correct Jacksons issue is `Change of behavior (2.8 -> 2.9) with ObjectMapper.readTree(input) with no content`, see: https://github.com/FasterXML/jackson-databind/issues/2211

PascalSchumacher:
I thought `2.10.0` was already released, but it wasn't yet. Should be just a few days until the release. I will update this pull request as soon as possible.

tijsrademakers:
Thanks for the PR, merged the changes in a separate commit.

Issue #1743, opened 2019-05-14
Multi-instance fails to pass correct element value to form


paulrda:
i have a User Task in my process and i want to execute user tasks **parallel**. i use multi-instance feature and added a List collection array to Collection property. i added a name for element variable and passed element variable to form property to show assigned element value to the form but i can see only last index value of List collection in every form of User Tasks.
Multi-instance feature only works if i change the User Task multi-instance execution to **Sequential**.
I use default form engine and i access the process and view User Task from Flowable-task UI. 
In my Collection Array i have two entries. first index value is "item 1" and second index value is "item 2". after enabling multi-instance this is how forms looks like.

_Process diagram View From Eclipse designer_
<img width="922" alt="eclipse screen shot" src="https://user-images.githubusercontent.com/15071681/57704999-a05f4380-7680-11e9-8af6-b1afe03e9740.png">

_First Form_

<img width="1423" alt="Multi Instance Form 1" src="https://user-images.githubusercontent.com/15071681/57685354-52811600-7655-11e9-9ae2-e54c45119745.png">

_Second Form_

<img width="1423" alt="Multi Instance Form 2" src="https://user-images.githubusercontent.com/15071681/57685396-67f64000-7655-11e9-8adb-964bba93c5e3.png">

There is no errors in logs to post here.
i have created a uni test project for this issue.
[https://github.com/paulrda/Home-Test-Projects/tree/master/flowable-multi-instance](url)

tijsrademakers:
Hi, the current form variable resolvement doesn't work for multi instance yet. You could still fetch the correct variables via the Java API and then invoke the getFormModelWithVariablesByKeyAndParentDeploymentId method in the org.flowable.form.api.FormService to get the correct form.

PR #1949, opened 2019-09-21
Pull form variables from taskService if task is active

unamanic:
Fixes: #1743

#### Check List:
* Unit tests:  NO
* Documentation: NO

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/cmd/GetTaskFormModelCmd.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/cmd/GetTaskFormModelCmd.java
@@ -1,9 +1,9 @@
 /* Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


unamanic:
I'm not entirely sure where sure where a test would go for this change.  There don't appear to be any form tests in flowable-engine, where this change is.

tijsrademakers:
@unamanic thanks for the PR, but looking at the change, before the historic variables were fetched with the process instance id, but with your change this is now done with the task id. This means that process variables will not be taken into account anymore. I think that's not correct. Or did you test the historic case as well and was it working fine?

unamanic:
@tijsrademakers, the referenced issue was caused by the fact that we were pulling by process id instead of by task.  When we pull with the process id instead of the task id, it pulls the current state of the element variable across the process.  This causes us to always get the last element that the process iterated through when creating parallel tasks.

With this change, process variables on completed tasks now show their state when they were completed instead of the most current state of the process in history.  Is that incorrect?  If so, I can change inactive tasks to fetch by process id instead.

tijsrademakers:
@unamanic the problem before was that task specific variables were not taken into account, that is ok now with your change. But now process variables are not used anymore, only task specific variables. It should be a combination of both process and task specific variables, with task variables having precedence over process variables (if the same variable name is used for both).

unamanic:
@tijsrademakers I'll introduce some code to explicitly bring in process variables.  Do you have any thoughts on where a unit test should go?

unamanic:
[Parallel App.zip](https://github.com/flowable/flowable-engine/files/3640174/Parallel.App.zip)
The code now explicitly brings in any process variables that are not present on the task.  

I've attached the App I've been using for testing.  It uses Groovy to create a list called `items` and a process variable called 'p_var'.  It then flows to a parallel user task that iterates over `items` and gives each task the element variable `item`.

PR #1960, opened 2019-10-01
Add logging to UnacquireOwnedJobsCmd

enchobelezirev:
#### Check List:
* Unit tests:  NA
* Documentation: NA

We had the requirement to check which jobs were unacquired by the AsyncExecutor and did not find a way to do it from the logs(the database selection is hard to be performed as the systems on which we are operating are productive and it is hard to access them).
The intention here is to add some logging in order to know which jobs are unacquired by the current AsyncExcutor.


enchobelezirev:
The build is working on my local machine.

filiphr:
The Build via the GitHub Actions is good. We are slowly moving towards that and we have both GtiHub Actions and Travis in parallel in the moment.

enchobelezirev:
Hi @filiphr,
Is there something I can do in order to fix the problem?

filiphr:
Thanks for the changes @enchobelezirev, we've merged it to master

PR #1969, opened 2019-10-08
Fixes BPMN Deadletter-jobs REST API documentation

remibantos:
Fixes BPMN Deadletter-jobs REST API documentation regarding Deadletter Jobs resuming.
Replace execute action with move action in the payload (only 'move' action is supported)

#### Check List:
* Unit tests: NA
* Documentation: YES


filiphr:
Thanks for the fix @remibantos

PR #1978, opened 2019-10-21
Update H2 to version 1.4.200.

PascalSchumacher:
Remove "MVCC=TRUE" setting from connections strings, because it is no longer supported (MVStore has been the default mode since H2 1.4).


PascalSchumacher:
Any feedback on this?

tijsrademakers:
Thanks for the ping, sorry that it was open for so long.

PR #1986, opened 2019-10-26
implement sub case termination during terminate all end event

vzickner:
#### Check List:
* Unit tests: YES
* Documentation: no, since it's the expected behavior to terminate a child case


modules/flowable-cmmn-engine-configurator/src/main/java/org/flowable/cmmn/engine/configurator/impl/cmmn/DefaultCaseInstanceService.java
```diff
@@ -109,4 +114,20 @@ public void deleteCaseInstance(String caseInstanceId) {
...
     }
 
+    @Override
+    public Set<String> findChildCaseIdsForExecutionIds(Collection<String> executionIds) {
+        CmmnRuntimeService cmmnRuntimeService = cmmnEngineConfiguration.getCmmnRuntimeService();
+        CaseInstanceQuery caseInstanceQuery = cmmnRuntimeService.createCaseInstanceQuery()
+                .caseInstanceCallbackType(CallbackTypes.EXECUTION_CHILD_CASE)
+                .or();
+        for (String executionId : executionIds) {
+            caseInstanceQuery.caseInstanceCallbackId(executionId);
```
filiphr: This is not going to work properly. The query will execute only with the last executionId, it is the same as without the or. The reason for that is that the query can have only one `caseInstanceCallbackId`. If you really need to query by all execution ids then add a `caseInstanceCallbackIds` to the query. In order to verify that this is not working you can add a test where you have 2 parallel case tasks and it won't pass.

modules/flowable-cmmn-engine-configurator/src/test/java/org/flowable/cmmn/test/CaseTaskTest.java
```diff
@@ -300,4 +303,37 @@ public void testCaseTaskIdVariableNameExpression() {
...
+                .addClasspathResource("org/flowable/cmmn/test/caseTaskProcessWithTerminateEndEventOnDifferentExecution.bpmn20.xml")
+                .deploy();
+
+        try {
+            // Arrange
+            ProcessInstance processInstance = processEngineRuntimeService.startProcessInstanceByKey("terminateByEndEvent");
+            Task task = processEngineTaskService.createTaskQuery()
+                    .processInstanceId(processInstance.getProcessInstanceId())
+                    .singleResult();
+            Assumptions.assumeThat(task).isNotNull();
```
filiphr: Why are you using assumptions? If for whatever reason the task is null the test would be ignored since the assumption is not met, it won't actually fail

modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/event/end/TerminateEndEventTest.java
```diff
@@ -1060,6 +1072,50 @@ public void testThreeExecutionsArrivingInTerminateEndEvent() {
         assertHistoricProcessInstanceDetails(processInstanceId);
     }
 
+    @Test
+    @Deployment
+    void testWithSubCaseTaskAndEnsureSubCaseIsTerminated() {
```
filiphr: Please don't use mockito. We want to test that the business logic is correct, not that our implementation does the proper calls. If this test needs the cmmn-engine then move it in the cmmn-engine-configurator, where we are testing the integration between the engines

modules/flowable-cmmn-engine-configurator/src/main/java/org/flowable/cmmn/engine/configurator/impl/cmmn/DefaultCaseInstanceService.java
vzickner: Thanks! I will fix it.

modules/flowable-cmmn-engine-configurator/src/test/java/org/flowable/cmmn/test/CaseTaskTest.java
vzickner: To ensure that the test is only failing in cases the test is supposed to fail. The test is not supposed to test if a simple user task is created, the only purpose of this test is to test that in case somebody triggers a terminate end event, the case is terminated.

modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/event/end/TerminateEndEventTest.java
vzickner: As you might have seen, this test also exists as an integration version within the `cmmn-engine-configurator`. I would also expect a test here, since this is part where the behavior of the terminate end event is tested, assuming it has an external `CaseInstanceService` which is doing what it is supposed to do. If you think this part is not necessary, I am happy to remove the test.

modules/flowable-cmmn-engine-configurator/src/main/java/org/flowable/cmmn/engine/configurator/impl/cmmn/DefaultCaseInstanceService.java
vzickner: fixed

modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/event/end/TerminateEndEventTest.java
vzickner: This test is gone, since I moved the implementation to a different location

filiphr:
@vzickner we needed to do some changes in order to avoid querying the case engine for every single execution (there are a lot of executions being created and deleted during the run of a process).

Therefore, I am closing this in favor of PR #2004 (It contains your commits from this PR).

PR #2014, opened 2019-11-14
Add support for collection-based repetition for plan items

mkiener:
Like the multi-instance type in BPMN, the collection-based repetition in CMMN works similar as you can specify a collection variable to be evaluated and used to create plan item instances, even with support for local variables holding the current item and item index.

#### Check List:
* Unit tests: YES
* Documentation: YES


modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/test/AbstractFlowableCmmnTestCase.java
```diff
@@ -173,6 +173,42 @@ protected void assertNoPlanItemInstance(List<PlanItemInstance> planItemInstances
...
+
+            Object itemValue = cmmnRuntimeService.getLocalVariable(task.getId(), "item");
+            Object itemIndexValue = cmmnRuntimeService.getLocalVariable(task.getId(), "itemIndex");
+            assertEquals(itemVariableValues.get(ii), itemValue);
+            assertEquals(itemIndexVariableValues.get(ii), itemIndexValue);
+        }
+    }
+
+    protected void completePlanItems(String caseInstanceId, String planItemName, int expectedCount, int numberToComplete) {
+        // now let's complete all Tasks B -> nothing must happen additionally
```
jbarrez:
```suggestion
```

PR #2029, opened 2019-11-23
feat(jsonConverter|bpmnConverter): handle elementIndexVariable conver…

mo-nick:
JSON converter support for Multi-Instance loop characteristics element index variable

#### Check List:
* Unit tests: YES
* Documentation: NA


modules/flowable-json-converter/src/test/java/org/flowable/editor/language/MultiInstanceConverterTest.java
```diff
@@ -0,0 +1,38 @@
+package org.flowable.editor.language;
```
dbmalkovsky: Please add standard license header; thanks.

mo-nick: done!

PR #2030, opened 2019-11-23
Pre upgrade script

martin-grofcik:


tijsrademakers:
Hi Martin,

I like the idea of execution logic before (and I think also after) each migration. Instead of only supporting script, could you maybe make this more flexible, so that you can also execute a Java class or an expression like we do in several other places?

gro-mar:
Hi Tijs,

yes, I am just checking whether the idea is good to merge. I will extend it.

Martin

gro-mar:
May be it could evolve into some kind of "migration" process....

 `start` --> `Upgrade script/javadelegate.....` --> `mapping` --> `end`

and we can benefit from the audit trail...... .

tijsrademakers:
Cool Martin, I think the idea of having a possibility to add pre upgrade logic is very useful. Will you work on the java delegate / expression support in this same PR?
Supporting migration processes is more a matter of how someone would like to implement it I think, it's already possible to do it in a BPMN process.

martin-grofcik:
Yes, I will try.

martin-grofcik:
@tijsrademakers Hi Tijs. Any update on this?

tijsrademakers:
Hi Martin, the changes look good. I think it would make sense to also add an after migration script /  class / expression as well. With that this PR would be complete and we can merge it master.

martin-grofcik:
https://github.com/flowable/flowable-engine/pull/2116

tijsrademakers:
Closing because of new PR #2116

PR #2033, opened 2019-11-26
Added support for BusinessKeyLike to (Historic)ProcessInstanceQuery

jmoraleda:
Added support for processInstance BusinessKeyLike to HistoricProcessInstaceQuery and ProcessInstanceQuery including REST api. This is discussed at https://forum.flowable.org/t/proposal-for-new-method-processinstancebusinesskeylike-to-historicprocessinstancequery/4886


jmoraleda:
@filiph I added some tests per your suggestion. Thank you for pointing me to the relevant classes. That was very helpful and definitely saved me time.

PR #2060, opened 2019-12-10
Remove unused jackson-datatype-hppc dependency.

PascalSchumacher:
Flowable does not use `High Performance Primitive Collections for Java` (https://github.com/carrotsearch/hppc) so this dependency can be removed.


PascalSchumacher:
Any feedback on this?

tijsrademakers:
Hi Pascal, we need to verify why this dependency was added. Will make sure this gets done in the next week.

PR #2066, opened 2019-12-18
JIRA-2065 WIP: handle process parsing and fail fast on unexpected XML

grexe:
- detect if activeProcess is null in ExtensionElementsParser
- added MessageEventDefinition as known type in IntermediateThrowEventValidator

#### Check List:
* Unit tests: WIP, will add after cleanup tomorrow
* Documentation: TBD


grexe:
this is only for having a better way to discuss issue #2066 - I hope it's fine, will happily clean up and prepare for a real merge based on your feedback.

tijsrademakers:
@grexe , we don't support an intermediate message throw event in the Flowable Engine, so that validator change is not correct. It's not supported because we don't support BPMN collaboration diagrams. Instead an intermediate signal throw event can be used. For the change for the active process null check looking forward to see a BPMN example for this to better understand it.

tijsrademakers:
Closing PR because of the merged PR of Tim (https://github.com/flowable/flowable-engine/pull/2068)

PR #2068, opened 2019-12-19
resolve #2065 NPE parsing collaboration extension elements

### LINKED ISSUE #2065: BPMN20 XML import throws NullPointerException when no process is found
**Describe the bug**
I have designed a BPMN2.0 process model in Signavio modeller and exported it as BPMN 2.0 XML. When trying to import the XML into Flowable 6.4.2, the importer throws a NPE:
```
2019-12-17 15:36:03.256 ERROR 1 --- [nio-8080-exec-8] o.f.u.m.s.FlowableModelQueryService      : Import failed for BigPicture.bpmn

org.flowable.bpmn.exceptions.XMLException: Error processing BPMN document
	at org.flowable.bpmn.converter.BpmnXMLConverter.convertToBpmnModel(BpmnXMLConverter.java:448) ~[flowable-bpmn-converter-6.4.2.1.jar:6.4.2.1]
	at org.flowable.ui.modeler.service.FlowableModelQueryService.importProcessModel(FlowableModelQueryService.java:179) ~[flowable-ui-modeler-logic-6.4.2.1.jar:6.4.2.1]
...
Caused by: java.lang.NullPointerException
	at org.flowable.bpmn.converter.parser.ExtensionElementsParser.parse(ExtensionElementsParser.java:55) ~[flowable-bpmn-converter-6.4.2.1.jar:6.4.2.1]
	at org.flowable.bpmn.converter.BpmnXMLConverter.convertToBpmnModel(BpmnXMLConverter.java:399) ~[flowable-bpmn-converter-6.4.2.1.jar:6.4.2.1]
```

**Expected behavior**
The model should be imported and visualised. It works in e.g. https://bpmn.io online demo.

**Code**
I will try to provide a simple test case for reproducing this bug if needed.

**Additional context**
Using flowable 6.4.2.1 (also tried 6.5.0-SNAPSHOT as of today).
Checked out source and ran the `org.flowable.engine.test.bpmn.usertask.ImportExportTest` with my file.
The problem is that if no process is found in the XML (there is none defined), then in `org.flowable.bpmn.converter.parser.ExtensionElementsParser:parse()`, the if/else block does not set the `parentElement` (`parentElement = activeProcess` will set `null`), and the subsequent call `parentElement.addExtensionElement(extensionElement);`will fail, resulting in a NPE.

tstephen:
#### Check List:
* Unit tests: YES
* Documentation: NA

### DIFF
--- a/modules/flowable-bpmn-converter/src/main/java/org/flowable/bpmn/converter/BpmnXMLConverter.java
+++ b/modules/flowable-bpmn-converter/src/main/java/org/flowable/bpmn/converter/BpmnXMLConverter.java
@@ -293,7 +293,7 @@ public class BpmnXMLConverter implements BpmnXMLConstants {
         model.setStartEventFormTypes(startEventFormTypes);
         model.setUserTaskFormTypes(userTaskFormTypes);
         try {
-            Process activeProcess = null;
+            Process activeProcess = new Process();
             List<SubProcess> activeSubProcessList = new ArrayList<>();
             while (xtr.hasNext()) {
                 try {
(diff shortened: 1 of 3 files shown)


tstephen:
The second commit unwinds the first and is better in my view as it additionally preserves the extension attributes which is of interest should the process ever need to be taken back to the modeling tool

PR #2073, opened 2019-12-19
Unnecessary null check

dbmalkovsky:
Both `equals()` and `containsAny()` handle nulls correctly

#### Check List:
* Unit tests:  NA
* Documentation:  NA


dbmalkovsky:
Too many commits; closing.

PR #2074, opened 2019-12-19
Remove unnecessary 'null' check

dbmalkovsky:
Both `equals()` and`containsAny()` handle null

#### Check List:
* Unit tests: NA
* Documentation:  NA


modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/multiinstance/CounterExecutionListener.java
```diff
@@ -37,7 +37,7 @@ public void notify(DelegateExecution execution) {
         String nrOfCompletedInstances = null == nrOfCompletedInstancesObj ? null : nrOfCompletedInstancesObj.toString();
         LOGGER.debug("Invoke into executionEnd method and execution is {}", execution.getVariables());
         
-        if (nrOfCompletedInstances != null && nrOfInstances.equals(nrOfCompletedInstances)) {
+        if (nrOfInstances.equals(nrOfCompletedInstances)) {
```
filiphr: Seems like this is needed. There are NPEs now in the tests. I think that we need to have 

```
nrOfInstances != null && nrOfInstances.equals(nrOfCompletedInstances
```

in the if instead

dbmalkovsky:
Closing and will reopen after further testing; thanks @filiphr for the catch.

PR #2082, opened 2019-12-24
AsyncSendEventJobHandler: Use correct placeholders in String#format c…

PascalSchumacher:
…all.


filiphr:
Good catch, thanks @PascalSchumacher

PR #2083, opened 2019-12-24
AbstractEngineConfiguration#initDatabaseType: Always close the Prepar…

PascalSchumacher:
…edStatement and the ResultSet.


filiphr:
Thanks @PascalSchumacher

PR #2136, opened 2020-01-21
Remove 10 unused imports

dbmalkovsky:
#### Check List:
* Unit tests:  NA
* Documentation:  NA


filiphr:
Thanks David. I also read your commit for the Duration import, but didn't get around to doing it (thanks for doing it instead).

dbmalkovsky:
NP; just bundled it with a bunch of other ones.

PR #2140, opened 2020-01-22
Spring beans secure js

mo-nick:
Access to beans from Script Tasks with Secure Javascript
discussed [here](https://forum.flowable.org/t/access-to-spring-beans-in-script-task/5209)

#### Check List:
* Unit tests: YES 
* Documentation:  NA


modules/flowable-secure-javascript/src/main/java/org/flowable/scripting/secure/impl/SecureJavascriptUtil.java
```diff
@@ -16,16 +16,22 @@
 import org.mozilla.javascript.Context;
 import org.mozilla.javascript.Scriptable;
 
+import java.util.Map;
```
dbmalkovsky: Project standards are to place `java` imports before `org` imports.  The same change needs to be made in the next file too.

mo-nick: done!

tijsrademakers:
Thanks

PR #2141, opened 2020-01-22
Travis: Remove OpenJDK EA build.

PascalSchumacher:
This build always fails, because Flowable tests use Groovy. Groovy uses ASM. ASM checks the class version in use (incremented with each Java version) and does not support future Java versions.


PascalSchumacher:
Or maybe the whole travis build should be dropped instead? Does it do anything not covered by github actions?

PR #2144, opened 2020-01-26
bug fix: JSON converter for boundary compensation event

mo-nick:
Bug fix in JSON converter for compensation boundary event.

#### Check List:
* Unit tests: YES 
* Documentation: NA

### DIFF
--- a/modules/flowable-json-converter/src/test/java/org/flowable/editor/language/AssociationLinkConverterTest.java
+++ b/modules/flowable-json-converter/src/test/java/org/flowable/editor/language/AssociationLinkConverterTest.java
@@ -1,0 +1,30 @@
+package org.flowable.editor.language;
+
+import org.flowable.bpmn.model.*;
+import org.junit.Test;
+
+import java.util.List;
+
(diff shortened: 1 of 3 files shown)


modules/flowable-json-converter/src/test/java/org/flowable/editor/language/AssociationLinkConverterTest.java
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+package org.flowable.editor.language;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.util.List;
+
+import org.flowable.bpmn.model.*;
```
dbmalkovsky: Another project standard is not to use wildcard imports but rather to enumerate all the classes required.  Thanks.

mo-nick: done

PR #2147, opened 2020-01-28
Use LinkedHashMap for deterministic iterations

contextshuffling:
Tests in `modules/flowable-engine/src/test/java/org/flowable/engine/test/api/runtime/changestate/ChangeStateTest.java` depends on `modules/flowable-engine/src/main/java/org/flowable/engine/impl/runtime/ChangeActivityStateBuilderImpl.java` where the processVariables and localVariables are stored in a HashMap. Tests are using iterators to get the values in the HashMap.  However, HashMap does not guarantee any specific order of entries. Tests can fail if the iteration order changes.

This PR proposes to fix the test by changing HashMap to LinkedHashMap that make the iteration order deterministic.

### DIFF
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/runtime/changestate/ChangeStateTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/runtime/changestate/ChangeStateTest.java
@@ -14,6 +14,7 @@
 package org.flowable.engine.test.api.runtime.changestate;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.tuple;
 
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


contextshuffling:
CI passed JDK8 but failed 11 and 13. 

I do not quite see how this code changes are related to the errros.

unamanic:
Thanks for the contribution.  Here are some thoughts for discussion:
- Outside of making tests easier, I'm not sure that maintaining insertion order is helpful.
- The `processVariables` method allows you to replaces the processVariables map with any implementation of Map<>, so even changing the instances in the class does not guarantee that the implementation will be maintain insertion order.
- If the tests are relying on a side effect that is not guaranteed (order of the keys), perhaps it's more appropriate to just correct the tests.

filiphr:
Thanks for the contribution @contextshuffling.

>CI passed JDK8 but failed 11 and 13.

The GitHub Actions are OK, so I would look at those. Travis is a bit flaky, your changes indeed have no link to the failures.

Regarding the comments from @unamanic, I would agree with them. The map in the implementation is not that important. I would only change the tests not to take order into consideration.

You can use `Assertions.assertThat` from AssertJ (it is already there as a dependency) and a mix of extracting and contains only in any order.

For example (not 100% complete):

```
assertThat(changeStateEventListener.getEvents())
        .extracting(FlowableEvent::getType)
        .containsExactly(
            FlowableEngineEventType.VARIABLE_CREATED,
            FlowableEngineEventType.VARIABLE_CREATED,
            FlowableEngineEventType.ACTIVITY_CANCELLED,
            FlowableEngineEventType.ACTIVITY_CANCELLED,
            FlowableEngineEventType.VARIABLE_CREATED,
            FlowableEngineEventType.VARIABLE_CREATED,
            FlowableEngineEventType.ACTIVITY_STARTED
        );

assertThat(changeStateEventListener.getEvents())
    .filteredOn(stateEvent -> stateEvent instanceof  FlowableVariableEvent)
    .extracting(
        stateEvent -> ((FlowableVariableEvent) stateEvent).getVariableName(),
        stateEvent -> ((FlowableVariableEvent) stateEvent).getVariableValue()
    )
    .containsExactlyInAnyOrder(
        tuple("processVar1", "test"),
        tuple("processVar2", 10),
        tuple("localVar1", "test2"),
        tuple("localVar2", 20)
    );
```

contextshuffling:
@filiphr Sounds reasonable! I will update the PR using the new assertion style

contextshuffling:
I updated the commit. I think this way makes more sense. The original iterator style looks confusing and it looks like it is coded with a specific JDK (run the test and write down the assertion). The order does not even match the setup order. According to the spec, the iteration order could be in any order and thus the test is flaky. With this commit, it also helps for the users to know that the order might depend on the underlying map implementation.

filiphr:
Thanks a lot for adapting this @contextshuffling

PR #2161, opened 2020-02-03
flowable-eventsubscription-service: Fix self assignment in EventSubsc…

PascalSchumacher:
…riptionQueryImpl.

No idea if this is the correct fix, but as it is these two lines to not have an effect, as they assign the field to itself.

### DIFF
--- a/modules/flowable-eventsubscription-service/src/main/java/org/flowable/eventsubscription/service/impl/EventSubscriptionQueryImpl.java
+++ b/modules/flowable-eventsubscription-service/src/main/java/org/flowable/eventsubscription/service/impl/EventSubscriptionQueryImpl.java
@@ -243,9 +243,9 @@ public class EventSubscriptionQueryImpl extends AbstractQuery<EventSubscriptionQ
         }
 
         if (inOrStatement) {
-            this.currentOrQueryObject.createdBefore = createdBefore;
+            this.currentOrQueryObject.createdBefore = beforeTime;
         } else {
-            this.createdBefore = createdBefore;
+            this.createdBefore = beforeTime;
         }
 
         return this;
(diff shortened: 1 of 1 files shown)


modules/flowable-eventsubscription-service/src/main/java/org/flowable/eventsubscription/service/impl/EventSubscriptionQueryImpl.java
```diff
@@ -260,7 +260,7 @@ public EventSubscriptionQueryImpl createdAfter(Date afterTime) {
         if (inOrStatement) {
             this.currentOrQueryObject.createdAfter = createdAfter;
         } else {
-            this.createdAfter = createdAfter;
+            this.createdAfter = afterTime;
```
filiphr: Probably a change is needed 2 lines above as well?

modules/flowable-eventsubscription-service/src/main/java/org/flowable/eventsubscription/service/impl/EventSubscriptionQueryImpl.java
```diff
@@ -245,7 +245,7 @@ public EventSubscriptionQueryImpl createdBefore(Date beforeTime) {
         if (inOrStatement) {
             this.currentOrQueryObject.createdBefore = createdBefore;
         } else {
-            this.createdBefore = createdBefore;
+            this.createdBefore = beforeTime;
```
filiphr: Probably a change is needed 2 lines above as well?

PascalSchumacher:
I have amended the pull request to the change the other two branches.

Thank you for pointing this out!

PR #2183, opened 2020-02-22
Update Java Uuid Generator to version 4.0.

PascalSchumacher:
Only change is that `Java Uuid Generator` now uses `slfj4` instead of  `log4j` logging, see:

https://github.com/cowtowncoder/java-uuid-generator/blob/master/release-notes/VERSION


PascalSchumacher:
Blueprint test fail for some reason. No idea why. ;(

PascalSchumacher:
There seems to be a problem with Java Uuid Generator OSGI Manifest, see: https://github.com/cowtowncoder/java-uuid-generator/issues/37

PR #2185, opened 2020-02-23
ServiceTask with expression triggerable

donkon:
A ServiceTask with an expression should also be triggerable:

The org.flowable.engine.impl.bpmn.behavior.ServiceTaskExpressionActivityBehavior
class would just not call leave when triggerable is true (in execute()) and leaves in trigger().

https://forum.flowable.org/t/how-to-use-triggerable/3016

#### Check List:
* Unit tests: YES 
* Documentation: YES


filiphr:
Thanks a lot for the contribution @donkon

PR #2188, opened 2020-02-26
Use assertj fluent style in flowable-cmmn-engine (part 1)

dbmalkovsky:
Breaking the changes into groups to make review easier.

Also applied the latest code formatting style.


filiphr:
Thanks for the PR David. This one is a bit bigger, so it would take me a bit more time fore reviewing it.

PR #2205, opened 2020-03-04
Clean up dependencies now that modules use assertj fluent style

dbmalkovsky:
Given the recent merges to modules using assertj fluent style remove unused dependencies.

#### Check List:
* Unit tests: NA
* Documentation: NA


filiphr:
I've reverted the changes @dbmalkovsky. It was my mistake merging this. The `junit-vintage-engine` is needed so that JUnit 5 and 4 tests can be run in the same time.

PR #2209, opened 2020-03-06
Use assertj fluent style in flowable-engine.  Part 1

dbmalkovsky:
Working on the `org.flowable.engine.test.api.event` package.

### DIFF
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/event/DatabaseEventLoggerTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/event/DatabaseEventLoggerTest.java
@@ -109,9 +109,13 @@ public class DatabaseEventLoggerTest extends PluggableFlowableTestCase {
                 Map<String, Object> data = objectMapper.readValue(entry.getData(), new TypeReference<HashMap<String, Object>>() {
 
                 });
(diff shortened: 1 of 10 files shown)


modules/flowable-engine/src/test/java/org/flowable/engine/test/api/event/DatabaseEventLoggerTest.java
```diff
@@ -97,193 +99,201 @@ public void testDatabaseEvents() throws IOException {
...
-
-                assertFalse(data.containsKey(Fields.DESCRIPTION));
-                assertFalse(data.containsKey(Fields.CATEGORY));
-                assertFalse(data.containsKey(Fields.OWNER));
-                assertFalse(data.containsKey(Fields.DUE_DATE));
-                assertFalse(data.containsKey(Fields.FORM_KEY));
-                assertFalse(data.containsKey(Fields.USER_ID));
-
-                assertEquals(testTenant, data.get(Fields.TENANT_ID));
+                assertThat(data.get(Fields.ID)).isNotNull();
```
filiphr: For maps we can in theory use `assertThat(data).containsOnlyKeys(...)`.

dbmalkovsky: @filiphr Made the changes per your note and also fixed the ordering of `.as()`.  Thanks

PR #2212, opened 2020-03-07
Simplify the Default*DiagramCanvas#generateImage implementations.

PascalSchumacher:


PascalSchumacher:
Looks like the travis build cache for the Java 13 builds is corrupt:

>[ERROR] error reading /home/travis/.m2/repository/com/sun/xml/bind/jaxb-impl/2.3.2/jaxb-impl-2.3.2.jar; zip END header not found

This caused the build of this pull request and other pull requests today to fail.

PR #2213, opened 2020-03-07
Use assertj fluent style in flowable-engine.  Part 2

dbmalkovsky:
Working on the `org.flowable.engine.test.api.event` package.

### DIFF
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/event/MultiInstanceUserTaskEventsTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/event/MultiInstanceUserTaskEventsTest.java
@@ -1,9 +1,9 @@
 /* Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


modules/flowable-engine/src/test/java/org/flowable/engine/test/api/event/MultiInstanceUserTaskEventsTest.java
```diff
@@ -179,345 +180,340 @@ public void testMultiInstanceCancelledWhenFlowToTerminateEnd() throws Exception
...
         }
 
-        assert(foundMultiExec1);
-        assert(foundMultiExec2);
-        assert(foundRootExec);
-        assert(foundBoundaryExec);
+        assert (foundMultiExec1);
+        assert (foundMultiExec2);
+        assert (foundRootExec);
+        assert (foundBoundaryExec);
```
filiphr:
```suggestion
        assertThat(foundMultiExec1).isTrue();
        assertThat(foundMultiExec2).isTrue();
        assertThat(foundRootExec).isTrue();
        assertThat(foundBoundaryExec).isTrue();
```

dbmalkovsky: @filiphr The corrections are made.  The definition of the `assert` methods in the base class that many tests extend makes finding all the old occurrences difficult.

filiphr: Thanks @dbmalkovsky, yes I know. using the java `assert` was most probably some kind of mistake long time ago. This was the only place that I spotted it

PR #2215, opened 2020-03-08
Provide extensions for instantiating exceptions

wxbty:
When I use a workflow service as a product and provide services for other applications through rpc, the custom javadelegate class of the exclusive gateway in the workflow xml uploaded by the application cannot be obtained, and the flowable lacks the corresponding extension point to push the instantiation exception to the call square

### DIFF
--- a/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/interceptor/InstantiateErrorHandler.java
+++ b/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/interceptor/InstantiateErrorHandler.java
@@ -1,0 +1,7 @@
+package org.flowable.common.engine.impl.interceptor;
+
+import org.flowable.common.engine.api.FlowableClassLoadingException;
+
+public interface InstantiateErrorHandler {
+    Object handle(Exception e, Object[] objects);
+}
--- a/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/interceptor/InstantiateErrorHandler.java
+++ b/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/interceptor/InstantiateErrorHandler.java
@@ -1,7 +1,5 @@
 package org.flowable.common.engine.impl.interceptor;
 
-import org.flowable.common.engine.api.FlowableClassLoadingException;
-
 public interface InstantiateErrorHandler {
     Object handle(Exception e, Object[] objects);
 }
--- a/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/interceptor/InstantiateErrorHandler.java
+++ b/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/interceptor/InstantiateErrorHandler.java
@@ -1,5 +1,20 @@
+/* Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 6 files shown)


filiphr:
Thanks for the PR @wxbty. Can you provide an example of how this would look like with rpc?

Regarding the extension point, I think it would make more sense to do something like the `DelegateInterceptor`. Basically do an `InstantiateInterceptor`  that looks like:

```
public interface InstantiateInterceptor {

    Object instantiate(String className);

}
```

and then this can be used to instantiate the class.

wxbty:
Thanks for your reply, I describe my needs.

We have many business teams that need the help of a flowable workflow engine, but we don't want every application to introduce a flowable jar package and create more than 20 tables. Our current approach is to provide a flowable service (A), and the business application (B) calls the remote interface (dubbo) of A instead of the flowable API, 
and they run like this:
    1. B uploads xml to A and returns processId (0001); 
                 String  pid = remoteUploadService.upload ("xxx.xml");
    2.A deploy (0001);
                 remoteDeployService.deploy (0001);
    3.start process 
     ...
The above is no different from using the flowable api directly, only the difference between local and remote.
But when encountering an exclusive gateway, the problem came. A stored B's xml, but A did not have the JavaDelegate class in the xml, so the process could not go on. Here I hope there are some extension points. When the instantiation of JavaDelegate in xml fails, the class name and context can be sent to B through remote invocation (such as dubbo's callback), so that B can instantiate the class through reflection and continue Business execution continues as defined by the process.
Regarding the naming of the interceptor, I think about it this way, the correct process will go to the interceptor, and exceptions require other things (like spring's filter and exceptionHandler)

modules/flowable-engine/src/main/java/org/flowable/engine/impl/cfg/ProcessEngineConfigurationImpl.java
```diff
@@ -76,12 +76,7 @@
 import org.flowable.common.engine.impl.el.function.VariableLowerThanOrEqualsExpressionFunction;
 import org.flowable.common.engine.impl.el.function.VariableNotEqualsExpressionFunction;
 import org.flowable.common.engine.impl.history.HistoryLevel;
-import org.flowable.common.engine.impl.interceptor.Command;
-import org.flowable.common.engine.impl.interceptor.CommandConfig;
-import org.flowable.common.engine.impl.interceptor.CommandContext;
-import org.flowable.common.engine.impl.interceptor.CommandInterceptor;
-import org.flowable.common.engine.impl.interceptor.EngineConfigurationConstants;
-import org.flowable.common.engine.impl.interceptor.SessionFactory;
+import org.flowable.common.engine.impl.interceptor.*;
```
dbmalkovsky: Project standards do not allow package imports; please leave the individual imports.

modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/interceptor/InstantiateErrorHandler.java
```diff
@@ -0,0 +1,5 @@
+package org.flowable.common.engine.impl.interceptor;
```
dbmalkovsky: Missing required license header at the top of the file; thanks

filiphr:
Thanks for the clarification @wxbty.

Can you please address the other remark I made regarding the `InstantiateInterceptor`. I don't think we need an `InstantiateErrorHandler` if we have the instantiate interceptor.

Regarding your use case, how do you plan to solve transactionality cross machines?

wxbty:
Maybe Interceptor is more in line with the overall style of flowable and has been modified. In addition, all write operations are on the flowable service side (A in the use case). I don't think there is a cross-machine transaction problem, thanks

wxbty:
> Thanks again for the pr @wxbty.
> 
> I see that you already took some of my input and changed the `InstantiateInterceptor`. Can we please rename the `handle` method into `instantiate` since that's what it does.
> 
> On top of that the current tests should pass without any changes. So I would try to pass the interceptor to the `instantiateDelegate` method instead of only the single place.

Thanks for your suggestion  @filiphr 
Modified as recommended to achieve the purpose of renaming and not modifying test cases. But somehow some test cases can’t pass, it seems to have nothing to do with my code, do you know why?

tijsrademakers:
Closing PR because of lack of unit tests. Please reopen when you have time to add unit tests.

PR #2220, opened 2020-03-12
Mapping missing case instance data

mikrobus:
Some CaseInstanceEntity data are not populated when calling `CaseInstanceHelperImpl.createCaseInstanceEntityFromDefinition`. 

In my use case these fields (businessKey and name) cannot be seen in the case-related event logging without calling additional queries.

The fix is to simply add the missing mapping.


mikrobus:
Closed until tests are fixed

PR #2222, opened 2020-03-13
Update RemoteIdmServiceImpl.java

camellibby:
Fix a tiny bug. Revert assert adminPassword.


filiphr:
Good spot @camellibby, thanks

PR #2224, opened 2020-03-15
Use assertj fluent style in flowable-engine.

dbmalkovsky:
All files in the `org.flowable.engine.test.api.mgmt` package are included.


filiphr:
Thanks David

PR #2231, opened 2020-03-19
Change the posted value of people/group fields from userid/groupid to user/group object when save/complete a task to match the backend api's need

silverhjk:
In the flowable-task ui, only the user id or group id will be sent to the backend api when a user save or complete a task that a people or group field is used in the form. But the backend service will check the posted data type, it will discard the data if the data type is not a Map. So the selected user or group can not be set to the process variables.  

#### Check List:
* Unit tests: NO
* Documentation: NO


tijsrademakers:
Thanks, nice catch

PR #2234, opened 2020-03-21
Replace Java 13 with Java 14 build on Github and Travis.

PascalSchumacher:
Java 13 is effectively end-of-life, so replace it with Java 14.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [1.8, 11, 13]
+        java: [1.8, 11, 14]
     steps:
     - uses: actions/checkout@v2-beta
       with:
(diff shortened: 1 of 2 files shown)


PascalSchumacher:
Java 14 build fails because Spring Boot still uses Groovy `2.5.9`.

This change will have to wait until Spring Boot `2.2.6` is released next week.

filiphr:
Hey @PascalSchumacherm thanks as always for the updates.

Regarding the drop of Java 13. Recently there was an interesting development and Azul stepped up to maintain Open JDK 13u as a Mid Term Support (MTS) release. [This](https://mail.openjdk.java.net/pipermail/jdk-updates-dev/2020-February/002535.html) is the announcement from Azul for nominating themselves to maintain Open JDK 13u.

So the question is now should we keep 13 or not? What do you think @tijsrademakers and @jbarrez?

PascalSchumacher:
Hi Filip,

wasn't aware that somebody plans to maintain OpenJDK 13. Thanks for letting me know!

tijsrademakers:
@filiphr I think it's good to keep Java 13  and add Java 14 as an addition.

PR #2235, opened 2020-03-21
Delete unused file azure-pipelines.yml.

PascalSchumacher:


filiphr:
Thanks for the PR @PascalSchumacher, this is actually used for testing with the Oracle Database. Oracle doesn't provide Docker images that we can use with GitHub actions, and at that time GitHub Actions didn't allow to use private docker registries (not sure if that has changed).

We had to build our own Oracle Docker image and publish it to our internal Flowable Docker Registry so we can use it in the tests. Azure Pipelines allows use of private Docker registries.

As soon as GitHub actions allows use of private Docker registries or Oracle publishes public Docker images we can remove this.

PascalSchumacher:
Thanks for the information, Filip!

Issue #2236, opened 2020-03-23
JsonType: Switching variable from 'large' to 'small' json does not work


reva:
When setting the value of a 'small' json variable (saved internally in the text column) to a 'large' value (saved as a 'byte array' internally), the text value does not get deleted. The JsonType variable type returns the 'small' value on the next call instead of the current value.

It should be possible to change the payload size of a json variable.

PR #2237, opened 2020-03-23
Fix issue when switching from small to large json values (#2236)

reva:
See #2236


filiphr:
Thanks for the PR @reva. However, what will now happen is that the bytes (rather the `VariableByteArrayEntity`) will stay, they won't get deleted.

filiphr:
Thanks for the PR Valentin, I have merged it manually (https://github.com/flowable/flowable-engine/commit/db53b45427b2e59f83fe638fe195a9d5f4261ebe) and polished it with https://github.com/flowable/flowable-engine/commit/ecb368a55d2678f74fc7a315c6bf6588329c9f5c

PR #2242, opened 2020-03-27
Exception Mapping in ServiceTaskExpression permit throwing checked Ex…

donkon:
In flowable 6, only unchecked exceptions can be used for the exception mapping feature. 
But if you have an expression like `${myBean.myMethod(..)}`, you typically refer to an external Java API, which declares also checked exceptions (which you propably want to catch in an boundaryerrorevent). If so, I have to rewrite my code to declare everywhere unchecked exceptions or to throw a new BpmnError().

So, I see not reason, why only unchecked exceptions are allowed here.


filiphr:
Thanks for the PR @donkon. Can we please add a test for this? Have a look at `BoundaryErrorMapTest. You can do it there I would say.

tijsrademakers:
Thanks @donkon for the PR, I've applied the fix in a separate commit, therefore closing this PR. See 51b4f1b07cfefa0dcbca1cd80a76b8d4409eccd7

PR #2245, opened 2020-03-27
Disable Flowable 5 tests GitHub Workflow for pull requests, because th…

PascalSchumacher:
…is build always fails.

It causes all recent pull request builds to fail.

Fixing it would be better, but I don't know how to do it.


filiphr:
Seems like there is some ordering problem (I was trying it locally). 

Makes sense to remove it from the PR builds for now

PR #2277, opened 2020-04-12
Speed up builds by skipping compiling of tests during the install run.

PascalSchumacher:


PascalSchumacher:
Does not make sense, as this just moves the compiling of the tests to the test run.

PR #2278, opened 2020-04-13
Cache maven repository to speed-up github actions builds.

PascalSchumacher:
Every time a pom changes the cache is invalidated. Failed builds do not affect the cache. This should avoid the cache corruption problems on travis ci.

### DIFF
--- a/.github/workflows/java-ea.yml
+++ b/.github/workflows/java-ea.yml
@@ -14,6 +14,12 @@ jobs:
       - uses: actions/checkout@v2-beta
         with:
           fetch-depth: 10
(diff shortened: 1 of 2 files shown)


PascalSchumacher:
Hi,

any feedback on this?

tijsrademakers:
Hi @PascalSchumacher , is this still valid with disabling Travis? Is there a big difference in the Github actions build when doing this?

PascalSchumacher:
It is hard to say how much of a difference it will make, without actually testing it.

My guess would be that each build should be around 2-3 minutes faster.

PR #2286, opened 2020-04-21
fix decision table result for hit policies with multiple entries

vzickner:
Behavior before:
  Execution a decision table with the hit policy rule order and output order resulted in either an array or the element, depending on how many rows are hit.
  The user of the Decision/DMN task now needs to lookup which of both exists and pick it accordingly.

Behavior after:
  For decision table with the hit policy rule order and output order the result is always returned as an array. Independent on how much entries are hit.

#### Check List:
* Unit tests: YES 
* Documentation: NO

### DIFF
--- a/modules/flowable-dmn-engine-configurator/src/test/java/org/flowable/dmn/test/runtime/DmnTaskTest.java
+++ b/modules/flowable-dmn-engine-configurator/src/test/java/org/flowable/dmn/test/runtime/DmnTaskTest.java
@@ -1,0 +1,112 @@
+/* Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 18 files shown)


modules/flowable-dmn-engine-configurator/src/test/java/org/flowable/dmn/test/runtime/DmnTaskTest.java
```diff
@@ -0,0 +1,112 @@
...
+import org.flowable.engine.runtime.ProcessInstance;
+import org.flowable.engine.test.ConfigurationResource;
+import org.flowable.engine.test.Deployment;
+import org.flowable.engine.test.FlowableTest;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
```
dbmalkovsky, line 29: FWIW, the ordering of the imports does not match project standards.

vzickner: I changed the order of the imports

tijsrademakers:
@vzickner, thanks for the PR. One thing we have to take care of is backward compatibility. Because this could lead to different runtime behaviour for running process and case instances. Can you maybe add an engine configuration option to enable or disable this new behaviour? The default could be to enable it.

vzickner:
Hi @tijsrademakers, thank you! I added the property `alwaysUseArraysForDmnMultiHitPolicies` to the BPMN and CMMN engine

PR #2305, opened 2020-05-01
Use assertj fluent style in flowable-bpmn-converter module.

dbmalkovsky:
This is part 4 of 5 parts.


filiphr:
Thanks for the PR @dbmalkovsky

PR #2314, opened 2020-05-05
Use assertj fluent style in flowable-dmn-engine module.

dbmalkovsky:
Added `net.javacrumbs.json-unit` dependency so `assertThatJson` was available.


dbmalkovsky:
FWIW, the JDK11 build failed in the IDM Engine module which is not the area that I changed.

PR #2315, opened 2020-05-05
fixing case exit sentry with completion semantics

mkiener:
Fixing case exit sentry with completion semantics, triggered by a milestone to run a completion evaluation on the spot and not relying on the completion flag

#### Check List:
* Unit tests: YES
* Documentation: NA


modules/flowable-cmmn-engine/src/test/java/org/flowable/cmmn/test/runtime/CompleteWithExitSentryByMilestoneTest.java
```diff
@@ -0,0 +1,38 @@
+package org.flowable.cmmn.test.runtime;
```
dbmalkovsky, line 1: @mkiener this file needs the license header added.  Thanks.

mkiener: good point, thanks! already updated it

Issue #2340, opened 2020-05-15
signalEventReceivedAsync with variables


gro-mar:
**Problem description:**
RutimeService does not provide methods to receive signals asynchronously with the payload.

**Describe the solution you'd like**
https://forum.flowable.org/t/add-payload-to-scheduleeventasync/6023
Implementation proposal:

PR #2341, opened 2020-05-15
#2340 payload as job variable implementation

gro-mar:
A simple proposal to start discussion about signalEventReceivedAsync with variables


tijsrademakers:
Hi Martin,
Thanks for the PR. A couple of comments:

- in another usage of this temporary variable storage logic the payload variables are stored separately, not as one blob. I think it would be good to also store the payload variables separately for the signal event triggering.
- could you change the "JOB" type value to "job"?
- why do you explicitly check for execution id being empty:

**if (StringUtils.isEmpty(executionId) && payload != null) {**

When execution id is not empty should the payload logic then not be used?

Thanks

filiphr:
I have some input for this as well. The cleanup of the variables is not in the correct place. Currently the variables are deleted when a job is deleted. However, a job can be deleted by it being moved to the deadletter, suspended or timer table. We need to think a bit more about this. The variables have to be able to live between these moves.

I am currently working on adding a correlation id to the jobs and perhaps we can use it as the scope or subScope id for the job. 

I just realized that the currently added external worker jobs have a similar issue as well, once they are moved from the external worker table in the executable table.

modules/flowable-engine/src/main/java/org/flowable/engine/impl/util/EventSubscriptionUtil.java
```diff
@@ -134,5 +139,16 @@ protected static void scheduleEventAsync(EventSubscriptionEntity eventSubscripti
         }
 
         jobService.scheduleAsyncJob(message);
+
+        if (StringUtils.isEmpty(executionId) && payload != null) {
```
martin-grofcik, line 143: @tijsrademakers : The payload is stored in variables only for start events (intermediate events variables are handled in the different way). From my point of view we could use the same approach for the both of them.

martin-grofcik:
@tijsrademakers 
> in another usage of this temporary variable storage logic the payload variables are stored separately, not as one blob. I think it would be good to also store the payload variables separately for the signal event triggering.

Could you point me to the place where it is used pls?
When the payload is an Object I do no know how to split it into several variables. It could be task for the Serializer to split the object storage.

martin-grofcik:
Hi @filiphr.

Yes - deleting variables is a problem.
Am I right, that the jobId is copied between deadletter/suspended jobs?
Have you any proposal how to delete job variables?
The correlation id - could you describe it little bit more pls?

Thanks for the answers
Martin

martin-grofcik:
If we have jobs for async signals, we should be also able to suspend job execution. May be should provide management service methods to provide it.

modules/flowable-engine/src/test/java/org/flowable/engine/test/externalworker/ExternalWorkerServiceTaskTest.java
```diff
@@ -253,6 +253,53 @@ void testExternalWorkerJobDeadLetterWithVariables() {
                 );
     }
 
+    @Test
+    @Deployment(resources = "org/flowable/engine/test/externalworker/ExternalWorkerServiceTaskTest.testSimple.bpmn20.xml")
+    void testSimpleWithVariablesLeft() {
```
filiphr, line 258: What is this test testing?

modules/flowable-rest/src/test/java/org/flowable/rest/service/api/runtime/SignalsResourceTest.java
```diff
@@ -141,8 +141,7 @@ public void testSignalEventReceivedAsync() throws Exception {
             var.put("value", "test");
 
             httpPost.setEntity(new StringEntity(requestNode.toString()));
-            closeResponse(executeRequest(httpPost, HttpStatus.SC_BAD_REQUEST));
-
+            closeResponse(executeRequest(httpPost, HttpStatus.SC_ACCEPTED));
```
filiphr, line 144: Add assertions that the variables were correctly saved and that continuing will use them properly

filiphr:
The job id is not copied between the different tables. That is why we added a correlation id. Have a look at https://github.com/flowable/flowable-engine/commit/9ed2bf03d803175a5b2344a7abef69e39c50f0dc.

martin-grofcik:
> The job id is not copied between the different tables. That is why we added a correlation id. Have a look at [9ed2bf0](https://github.com/flowable/flowable-engine/commit/9ed2bf03d803175a5b2344a7abef69e39c50f0dc).

When the job is moved to dead letter queue (or is suspended) its Id is the same, right? So the job.id is the same for the dead letter queue job, suspended job and active job too.

filiphr:
Yes seems like you are right about that. However, there are times where we explicitly set the job id to `null`. Look at `DefaultJobManager#unacquire`

martin-grofcik:
@tijsrademakers : What is the status of this pull request?
What is still missing?

tijsrademakers:
@martin-grofcik is this still something you want to look into? I think a flag on the job entity would be good to have to indicate if any variables are added, to prevent doing redundant delete queries for every job for example.

filiphr:
@martin-grofcik we are going to close this since it hasn't had any activity since the last comment from @tijsrademakers. If you want to have this functionality, I would suggest opening a new PR having in mind the comments from Tijs and do it from main to avoid the conflicts.

PR #2350, opened 2020-05-19

dbmalkovsky:
I have started to work on files that are a mix of assertion styles in the same file. There were 5 such files in the `org.flowable.engine.test.runtime` package.

### DIFF
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/runtime/ProcessInstanceIdentityLinksTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/runtime/ProcessInstanceIdentityLinksTest.java
@@ -87,7 +87,7 @@ public class ProcessInstanceIdentityLinksTest extends PluggableFlowableTestCase 
             List<Event> processInstanceEvents = runtimeService.getProcessInstanceEvents(processInstanceId);
             assertThat(processInstanceEvents)
                     .extracting(Event::getAction)
-                    .contains(Event.ACTION_DELETE_GROUP_LINK);
+                    .containsExactlyInAnyOrder(Event.ACTION_DELETE_GROUP_LINK, Event.ACTION_ADD_GROUP_LINK);
             List<String> processInstanceEventMessageParts = processInstanceEvents.get(0).getMessageParts();
             assertThat(processInstanceEventMessageParts)
                     .containsOnly("muppets", IdentityLinkType.PARTICIPANT);
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/runtime/ProcessInstanceIdentityLinksTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/runtime/ProcessInstanceIdentityLinksTest.java
@@ -45,16 +45,16 @@ public class ProcessInstanceIdentityLinksTest extends PluggableFlowableTestCase 
         List<IdentityLink> identityLinks = runtimeService.getIdentityLinksForProcessInstance(processInstanceId);
         IdentityLink identityLink = identityLinks.get(0);
 
-        assertNull(identityLink.getGroupId());
-        assertEquals("kermit", identityLink.getUserId());
-        assertEquals(IdentityLinkType.PARTICIPANT, identityLink.getType());
-        assertEquals(processInstanceId, identityLink.getProcessInstanceId());
+        assertThat(identityLink.getGroupId()).isNull();
+        assertThat(identityLink.getUserId()).isEqualTo("kermit");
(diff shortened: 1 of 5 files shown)


modules/flowable-engine/src/test/java/org/flowable/engine/test/api/runtime/ProcessInstanceIdentityLinksTest.java
```diff
@@ -67,49 +67,33 @@ public void testCandidateGroupLink() {
...
-            Event processIsntanceEvent = findProcessInstanceEvent(processInstanceEvents, Event.ACTION_DELETE_GROUP_LINK);
-            assertEquals(Event.ACTION_DELETE_GROUP_LINK, processIsntanceEvent.getAction());
-            List<String> processInstanceEventMessageParts = processIsntanceEvent.getMessageParts();
-            assertEquals("muppets", processInstanceEventMessageParts.get(0));
-            assertEquals(IdentityLinkType.PARTICIPANT, processInstanceEventMessageParts.get(1));
-            assertEquals(2, processInstanceEventMessageParts.size());
-            assertEquals(2, processInstanceEvents.size());
+            assertThat(processInstanceEvents)
+                    .extracting(Event::getAction)
+                    .contains(Event.ACTION_DELETE_GROUP_LINK);
```
filiphr, line 90: Why not `containsExactly(Event.ACTION_DELETE_GROUP_LINK, Event.ACTION_DELETE_GROUP_LINK)` this way the size of 2 is also verified. Or add `hasSize(2)`

modules/flowable-engine/src/test/java/org/flowable/engine/test/api/runtime/ProcessInstanceSuspensionTest.java
```diff
@@ -413,172 +343,99 @@ public void testTaskOperationsFailAfterProcessInstanceSuspend() {
...
         // to create separate tests for each of them.
 
         // Completing the task should fail
-        try {
-            taskService.complete(task.getId());
-            fail("It is not allowed to complete a task of a suspended process instance");
-        } catch (FlowableException e) {
-            // This is good
-        }
+        assertThatThrownBy(() -> taskService.complete(task.getId()))
```
filiphr, line 355: On this and the rest of the `assertThatThrownBy` add `as(description)` to keep the description

modules/flowable-engine/src/test/java/org/flowable/engine/test/api/runtime/ProcessInstanceSuspensionTest.java
```diff
@@ -629,23 +486,20 @@ public void testJobActivationAfterProcessInstanceSuspend() {
...
 
         // Activation of the suspended job instance should throw exception because parent is suspended
-        try {
-            managementService.moveSuspendedJobToExecutableJob(suspendedJob.getId());
-            fail("FlowableIllegalArgumentException expected. Cannot activate job with suspended parent");
-        } catch (FlowableIllegalArgumentException e) {
-            assertThat(e.getMessage(), containsString("Can not activate job "+ suspendedJob.getId() + ". Parent is suspended."));
-        }
+        assertThatThrownBy(() -> managementService.moveSuspendedJobToExecutableJob(suspendedJob.getId()))
+                .isExactlyInstanceOf(FlowableIllegalArgumentException.class)
```
filiphr, line 501: add `as(description)` to keep the one from the `fail`

modules/flowable-engine/src/test/java/org/flowable/engine/test/api/runtime/RuntimeActivityInstanceTest.java
```diff
@@ -420,44 +415,47 @@ public void testLoop() {
...
         List<ActivityInstance> taskActivityInstances = runtimeService.createActivityInstanceQuery().activityType("userTask").list();
-        assertEquals(10, taskActivityInstances.size());
-        for (ActivityInstance activityInstance : taskActivityInstances) {
-            assertNotNull(activityInstance.getStartTime());
-            assertNotNull(activityInstance.getEndTime());
-        }
+        assertThat(taskActivityInstances).hasSize(10);
+        assertThat(taskActivityInstances)
+                .extracting(ActivityInstance::getEndTime, ActivityInstance::getEndTime)
+                .doesNotContainNull();
```
filiphr, line 430: Using `doesNotContainNull` here is incorrect. It will always be not `null` since the assertion is on the object which is a `tuple` of not null values. You can use `flatExtracting` or keep the loop.

Same comment applies to the other usages of `doesNotContainNull` that extract more than one value

dbmalkovsky:
@filiphr Thanks for the detailed review (again).   I think I've addressed all the current issues.

filiphr:
Thanks for doing the changes. I'll merge this once the build is done

PR #2355, opened 2020-05-21
add multiple results property from decision tables to the DecisionTableVariableManager

vzickner:
#### Check List:
* Unit tests: YES
* Documentation: NA


yvoswillens (commented):
@vzickner shouldn't the change also be applied to the `setDecisionServiceVariablesOnExecution`?

vzickner:
Thanks @yvoswillens, I think you are right. I will do the change.

vzickner:
Hi @yvoswillens, I added the same for decision services as well. Also, multiple results are now set for decision services in case there are multiple results (either through multiple decision tables or one decision table with multiple results).

yvoswillens:
Thanks for the PR @vzickner !

PR #2356, opened 2020-05-22

dbmalkovsky:
I have started to work on files that are a mix of assertion styles in the same file. There were 4 such files in the `org.flowable.engine.test.api.task` package.

I need to find files that are not 3600+ lines long :)


filiphr:
Thanks a lot @dbmalkovsky.

>I need to find files that are not 3600+ lines long :)

Please do, would make my life way easier 😉.

PR #2358, opened 2020-05-25
Use assertj fluent style in flowable-engine module and add check to prevent NPE.

dbmalkovsky:
While converting a test file that contained various forms of assertions to `assertj` I found an example of a cut-n-paste that didn't test the method that the author thought it did.  So when I changed the method name to the correct method name and reran the test a NPE was generated instead of the expected `FlowableIllegalArgumentException`.   Looking in the source file I noticed that all other "ignore case" methods checked for `null` before trying `value.toLowerCase()` except in the case being tested.  This fixes that error.

#### Check List:
* Unit tests: Yes.
* Documentation: NA


filiphr:
Thanks for the PR @dbmalkovsky. This was a big one. I have one request for the future. If possible try to avoid doing formatting changes, even if the lines is too long leave it. It makes it much easier to review. Thanks a lot

dbmalkovsky:
OK I will try to avoid some of the formatting.

PR #2362, opened 2020-05-26
Use assertj fluent style in flowable-form-spring-configurator module.

dbmalkovsky:
Just a single test file that had a mix of assertion styles.


dbmalkovsky:
The JDK 14 test failed in the Flowable Engine module which is unchanged.

PR #2363, opened 2020-05-27
Use assertj fluent style in flowable-http module.

dbmalkovsky:
There was a mixture of assertion styles that now all use assertj.  The conversion was complicated by a local function named `assertEquals()`.

### DIFF
--- a/modules/flowable-http/src/test/java/org/flowable/http/bpmn/HttpServiceTaskTest.java
+++ b/modules/flowable-http/src/test/java/org/flowable/http/bpmn/HttpServiceTaskTest.java
@@ -13,10 +13,12 @@
 package org.flowable.http.bpmn;
 
 import static net.javacrumbs.jsonunit.assertj.JsonAssertions.assertThatJson;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.assertj.core.api.Assertions.entry;
+import static org.assertj.core.api.Assertions.tuple;
 
 import java.io.IOException;
-import java.net.SocketException;
-import java.net.SocketTimeoutException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


modules/flowable-http/src/test/java/org/flowable/http/bpmn/HttpServiceTaskTest.java
```diff
@@ -165,30 +158,21 @@ public void testHttpsSelfSigned() {
...
-            runtimeService.startProcessInstanceByKey("requestTimeout");
-            fail("FlowableException expected");
-        } catch (final Exception e) {
-            assertTrue(e instanceof FlowableException);
-            assertTrue(e.getCause() instanceof SocketTimeoutException
-                    || e.getCause() instanceof SocketException);
-        }
+        assertThatThrownBy(() -> runtimeService.startProcessInstanceByKey("requestTimeout"))
+                .isExactlyInstanceOf(FlowableException.class)
+                .hasCauseInstanceOf(IOException.class);
```
filiphr, line 171: The cause here should be `SocketTimeoutException` or `SocketException`

modules/flowable-http/src/test/java/org/flowable/http/bpmn/HttpServiceTaskTest.java
```diff
@@ -238,27 +218,19 @@ public void testRequestTimeoutFromProcessModelHasPrecedenceSuccess() {
...
-            fail("FlowableException expected");
-        } catch (Exception e) {
-            assertTrue(e instanceof FlowableException);
-            assertEquals("HTTP400", e.getMessage());
-        }
-        assertNull("Process instance was not started.", process);
+        assertThatThrownBy(() -> runtimeService.startProcessInstanceByKey("failStatusCodes"))
+                .isExactlyInstanceOf(FlowableException.class)
+                .hasMessage("HTTP400");
+        assertThat(process).as("Process instance was not started.").isNull();
```
filiphr, line 233: This is always null since `process` is never re-assigned

modules/flowable-http/src/test/java/org/flowable/http/bpmn/HttpServiceTaskTest.java
dbmalkovsky: I coded it this way because I couldn't figure out how to do an `OR` and both `SocketTImeoutException` and `SocketException` extend from `IOException`.

filiphr: You can do `getCause().isInstanceOfAny(SocketTimeoutException.class, SocketException.class)`.

`getCause()` returns a new `AbstractThrowableAssert` and that has `isInstanceOfAny`.

dbmalkovsky: Got it; thanks much.  My one new thing I've learned today :)

PR #2370, opened 2020-05-30

dbmalkovsky:
This is for the `org.flowable.engine.test.impl.calendar` package.  There is a single file in this package that had a mixture of assertion styles.


dbmalkovsky:
The JDK 14 test failed in the CDI module which is well beyond the code this PR touches.

PR #2373, opened 2020-06-01
Set SSL SMTP port when SSL is enabled for email server connections (fixes #2343)

### LINKED ISSUE #2343: MailActivityBehavior with SSL not setting port correctly
**Describe the bug**
When SSL is enabled for email sending, `org.flowable.engine.impl.bpmn.behavior.MailActivityBehavior` is not setting the SMTP port properly. This causes the default SSL port of `465` to be used regardless of other settings.

NOTE: same problem needs to be fixed in `org.activiti.engine.impl.bpmn.behavior.MailActivityBehavior`

**Expected behavior**
In `MailActivityBehavior`'s `setMailServerProperties` function, when the call to `isMailServerUseSSL()` returns `true`, the code needs to set the port using the `setSslSmtpPort` function on `org.apache.commons.mail.Email`. Currently it incorrectly uses `setSmtpPort`.

Question remains whether an additional configuration property should be exposed to set this explicitly.

**Additional context**
- Flowable 6.5.0
- `org.apache.commons:commons-email` version `1.5`

travishaagen:
Set SSL SMTP port when SSL is enabled for email server connections (fixes #2343)

I was able to manually test this fix as it applies to a Flowable 6.5.x BPMN XML use-case with an Email Task. The SSL SMTP port could be customized as needed.

#### Check List:
* Unit tests: NO
* Documentation: YES

### DIFF
--- a/docs/docusaurus/docs/bpmn/ch05a-Spring-Boot.md
+++ b/docs/docusaurus/docs/bpmn/ch05a-Spring-Boot.md
@@ -702,6 +702,7 @@ Here is a list of configuration properties that the Flowable Spring Boot support
     flowable.mail.server.host=localhost # The host of the mail server.
     flowable.mail.server.password= # The password for the mail server authentication.
     flowable.mail.server.port=1025 # The port of the mail server.
+    flowable.mail.server.ssl-port=1465 # The SSL port of the mail server.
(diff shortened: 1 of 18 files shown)


tijsrademakers:
Thanks for PR

PR #2387, opened 2020-06-06

dbmalkovsky:
This is for the `org.flowable.engine.test.bpmn.deployment` package.  Some files had a mix of styles so I converted the whole package of 6 files.

### DIFF
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/deployment/ParsedDeploymentTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/deployment/ParsedDeploymentTest.java
@@ -81,22 +81,22 @@ public class ParsedDeploymentTest extends PluggableFlowableTestCase {
 
         List<ProcessDefinitionEntity> processDefinitions = parsedDeployment.getAllProcessDefinitions();
 
-        assertThat(parsedDeployment.getDeployment()).isSameAs(entity);
+        assertThat(parsedDeployment.getDeployment()).isEqualTo(entity);
         assertThat(processDefinitions).hasSize(4);
 
         ProcessDefinitionEntity id1 = getProcessDefinitionEntityFromList(processDefinitions, ID1_ID);
         ProcessDefinitionEntity id2 = getProcessDefinitionEntityFromList(processDefinitions, ID2_ID);
         assertThat(parsedDeployment.getBpmnParseForProcessDefinition(id1))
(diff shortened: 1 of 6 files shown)


modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/deployment/ParsedDeploymentTest.java
```diff
@@ -84,25 +81,25 @@ public void testCreateAndQuery() throws UnsupportedEncodingException {
...
 
         ProcessDefinitionEntity id1 = getProcessDefinitionEntityFromList(processDefinitions, ID1_ID);
         ProcessDefinitionEntity id2 = getProcessDefinitionEntityFromList(processDefinitions, ID2_ID);
-        assertThat(parsedDeployment.getBpmnParseForProcessDefinition(id1),
-                sameInstance(parsedDeployment.getBpmnParseForProcessDefinition(id2)));
-        assertThat(parsedDeployment.getBpmnModelForProcessDefinition(id1), sameInstance(parsedDeployment.getBpmnParseForProcessDefinition(id1).getBpmnModel()));
-        assertThat(parsedDeployment.getProcessModelForProcessDefinition(id1), sameInstance(parsedDeployment.getBpmnParseForProcessDefinition(id1).getBpmnModel().getProcessById(id1.getKey())));
-        assertThat(parsedDeployment.getResourceForProcessDefinition(id1).getName(), equalTo(IDR_XML_NAME));
-        assertThat(parsedDeployment.getResourceForProcessDefinition(id2).getName(), equalTo(IDR_XML_NAME));
+        assertThat(parsedDeployment.getBpmnParseForProcessDefinition(id1))
```
filiphr, line 89: There is a difference between `isEqualTo` and `isSameInstanceAs`. Let's keep the assertions same as before and use `isSameInstanceAs`

dbmalkovsky: @filiphr  I used `isSameAs` as I couldn't find `isSameInstanceAs` except in the Google Truth package that is their version of assertj.  The [source code is here](https://github.com/google/truth/blob/6dcfce77dea780d6e2dd572244be8bf872452af5/core/src/main/java/com/google/common/truth/Subject.java#L267).

filiphr: I was talking from the top of my head, I didn't check the exact name of the method 😄. Thanks for changing it

PR #2388, opened 2020-06-06
Fix test to ignore order of entries.

dbmalkovsky:
This was found in the validation steps of another PR that I submitted. 

#### Check List:
* Unit tests: YES
* Documentation: NA


modules/flowable-http/src/test/java/org/flowable/http/cmmn/CmmnHttpTaskTest.java
```diff
@@ -79,11 +79,11 @@ public void testGetWithResponseHandler() {
...
         Map<String, Object> variables = cmmnRule.getCmmnRuntimeService().getVariables(caseInstance.getId());
+        Map<String, String> names = new HashMap<>();
+        names.put("firstName", "John");
+        names.put("lastName", "Doe");
         assertThat(variables)
-                .containsExactly(
-                        entry("firstName", "John"),
-                        entry("lastName", "Doe")
-                );
+                .containsExactlyInAnyOrderEntriesOf(names);
```
filiphr, line 86: Thanks for looking into this @dbmalkovsky. 

Just FYI, this could have been done with `containsOnly` as well. Which ignores the order of the entries, but no big deal

PR #2389, opened 2020-06-09
Add missing license headers.

dbmalkovsky:


dbmalkovsky:
The failure in the JDK 14 check is fixed in this pending [PR](https://github.com/flowable/flowable-engine/pull/2388)

PR #2405, opened 2020-06-14

dbmalkovsky:
In the `org.flowable.engine.test.bpmn.event.message` package there were 2 files with mixed assertion styles so they were converted to a single format.  At the same time the remaining 4 files in the package were updated.

### DIFF
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/event/message/MessageBoundaryEventTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/event/message/MessageBoundaryEventTest.java
@@ -73,7 +73,6 @@ public class MessageBoundaryEventTest extends PluggableFlowableTestCase {
         assertThat(userTask.getTaskDefinitionKey()).isEqualTo("taskAfterTask");
         taskService.complete(userTask.getId());
         assertThat(runtimeService.createProcessInstanceQuery().count()).isZero();
-
     }
 
     @Test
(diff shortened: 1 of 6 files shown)


filiphr:
@dbmalkovsky seems like there are some whitespace changes in `MessageBoundaryEventTest`, which are causing the entire file to be changed. Can you please have a look at it and make sure that only relevant changes are committed?

dbmalkovsky:
@filphr  I have no clue as to what has happened here; I did this package exactly like all the other ones.   I found the easiest way to see the diffs is `Hide whitespace changes` in the drop down menu.



image



In the meantime I'll try to figure out what is going on.

filiphr:
Yes that will work, but the git history will still have it look like the entire file changed. Sometimes this happens via the IDE. When it happens to me, I try to do the commit from scratch, by reverting it completely and apply the changes one by one again (copy pasting sections).

I'll have a look at it with the hidden changes tomorrow

dbmalkovsky:
@filiphr I think I know what the problem is; it has to do with line endings.  This is the message that I get when I do a ` git add MessageIntermediateEventTest.java`

<pre>
warning: LF will be replaced by CRLF in modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/event/message/MessageIntermediateEventTest.java.

The file will have its original line endings in your working directory
</pre>

dbmalkovsky:
Closing this PR so another attempt can be made where the whole file doesn't show up as a diff.

PR #2420, opened 2020-06-18

dbmalkovsky:
In the `org.flowable.engine.test.jobexecutor` package this PR updates the remaining files that were not using `assertj`.

### DIFF
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/jobexecutor/AsyncExecutorTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/jobexecutor/AsyncExecutorTest.java
@@ -12,6 +12,9 @@
  */
 package org.flowable.engine.test.jobexecutor;
 
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
 import java.util.Date;
 import java.util.concurrent.Callable;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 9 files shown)


modules/flowable-engine/src/test/java/org/flowable/engine/test/jobexecutor/AsyncExecutorTest.java
```diff
@@ -54,28 +56,25 @@ public void testRegularAsyncExecution() {
...
-                waitForAllJobsBeingExecuted(processEngine, 500L);
-                Assert.fail();
-            } catch (FlowableException e) {
-                // Expected
-            }
-            Assert.assertEquals(1, processEngine.getTaskService().createTaskQuery().taskName("The Task").count());
-            Assert.assertEquals(0, processEngine.getTaskService().createTaskQuery().taskName("Task after timer").count());
-            Assert.assertEquals(1, processEngine.getManagementService().createTimerJobQuery().count());
-            Assert.assertEquals(0, getAsyncExecutorJobCount(processEngine));
+            ProcessEngine processEngine1 = processEngine;
```
tijsrademakers, line 59: Hi @dbmalkovsky , can you elaborate a bit why the assignment to processEngine1 was added?

dbmalkovsky: @tijsrademakers the reason is that the `assertThatThrownBy` requires that the variables be effectively final in the clause and this was the easiest way to get that without changing a lot of code.

PR #2426, opened 2020-06-20

dbmalkovsky:
In the continuing quest to convert all of the `flowable-engine` module to assertj this PR converts the test files in the `org.flowable.engine.test.cfg` package.


dbmalkovsky:
The failures appear to be because of `FlowableOptimisticLocking` issues.

PR #2427, opened 2020-06-20

dbmalkovsky:
Continuing the quest to convert all of `flowable-engine` to `assertj` this PR does all files in the `org.flowable.examples` directory except for the `org.flowable.examples.bpmn` subdirectory in order to keep the size of the PR down.

FWIW, in `StandaloneTaskTest` test file in lines 66-74 I reverted my changes using the `assertThat/extracting/contains` form because when it was there every time I ran the test it generated a concurrent modification exception and test failed.


dbmalkovsky:
The JDK 13 build failed because the OSGi module could not resolve a dependency.

PR #2428, opened 2020-06-21

dbmalkovsky:
This is part 1 for the `org.flowable.examples.bpmn` package.  The files converted are up to and include those in the `gateway` subpackage.   More to come.....


filiphr:
Thanks for all the PRs @dbmalkovsky. We've been going one by one through them

PR #2429, opened 2020-06-22

dbmalkovsky:
This is part 2 for the `org.flowable.examples.bpmn` package. The files converted are after the `gateway` subpackage through and including the `subprocess` package. There is one more part to come.....

### DIFF
--- a/modules/flowable-engine/src/test/java/org/flowable/examples/bpmn/scripttask/ScriptTaskTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/examples/bpmn/scripttask/ScriptTaskTest.java
@@ -28,6 +28,8 @@ import org.junit.jupiter.api.Test;
 
 import com.fasterxml.jackson.databind.node.ObjectNode;
 
+import groovy.lang.MissingPropertyException;
+
 /**
  * @author Joram Barrez
  * @author Christian Stettler
@@ -51,14 +53,14 @@ public class ScriptTaskTest extends PluggableFlowableTestCase {
     @Deployment
     public void testFailingScript() {
         assertThatThrownBy(() -> runtimeService.startProcessInstanceByKey("failingScript"))
-                .hasStackTraceContaining("MissingPropertyException");
+                .hasRootCauseInstanceOf(MissingPropertyException.class);
     }
 
     @Test
     @Deployment
     public void testExceptionThrownInScript() {
         assertThatThrownBy(() -> runtimeService.startProcessInstanceByKey("failingScript"))
-                .hasStackTraceContaining("IllegalStateException");
+                .hasRootCauseInstanceOf(IllegalStateException.class);
     }
 
     @Test
--- a/modules/flowable-engine/src/test/java/org/flowable/examples/bpmn/scripttask/ScriptTaskTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/examples/bpmn/scripttask/ScriptTaskTest.java
@@ -12,8 +12,13 @@
  */
 package org.flowable.examples.bpmn.scripttask;
 
-import com.fasterxml.jackson.databind.node.ObjectNode;
-import groovy.lang.MissingPropertyException;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.util.HashMap;
+import java.util.Map;
+
 import org.flowable.common.engine.api.FlowableException;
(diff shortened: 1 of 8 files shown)


dbmalkovsky:
The JDK 11 job failed because `flowable-mule` could not resolved dependencies.

modules/flowable-engine/src/test/java/org/flowable/examples/bpmn/scripttask/ScriptTaskTest.java
```diff
@@ -39,84 +43,63 @@ public void testSetScriptResultToProcessVariable() {
...
-        try {
-            runtimeService.startProcessInstanceByKey("failingScript");
-        } catch (Exception e) {
-            expectedException = e;
-        }
-
-        // Check if correct exception is found in the stacktrace
-        verifyExceptionInStacktrace(expectedException, MissingPropertyException.class);
+        assertThatThrownBy(() -> runtimeService.startProcessInstanceByKey("failingScript"))
+                .hasStackTraceContaining("MissingPropertyException");
```
filiphr, line 54: Although this is correct. I would prefer to keep the same assertion as before. Verify that the `MissingPropertyException` is in the cause is one of the causes in the stacktrace and not just containing the string

modules/flowable-engine/src/test/java/org/flowable/examples/bpmn/scripttask/ScriptTaskTest.java
```diff
@@ -39,84 +43,63 @@ public void testSetScriptResultToProcessVariable() {
...
-        Exception expectedException = null;
-        try {
-            runtimeService.startProcessInstanceByKey("failingScript");
-        } catch (Exception e) {
-            expectedException = e;
-        }
-
-        verifyExceptionInStacktrace(expectedException, IllegalStateException.class);
+        assertThatThrownBy(() -> runtimeService.startProcessInstanceByKey("failingScript"))
+                .hasStackTraceContaining("IllegalStateException");
```
filiphr, line 61: Some here

modules/flowable-engine/src/test/java/org/flowable/examples/bpmn/servicetask/JavaServiceTaskTest.java
```diff
@@ -132,36 +135,29 @@ public void testExpressionFieldInjectionWithSkipExpression() {
...
         Execution execution2 = runtimeService.createExecutionQuery().processInstanceId(pi2.getId()).activityId("waitState").singleResult();
 
-        assertNotNull(execution2);
+        assertThat(execution2).isNotNull();
 
         Map<String, Object> pi2VarMap = runtimeService.getVariables(pi2.getProcessInstanceId());
-        assertFalse(pi2VarMap.containsKey("var1"));
-        assertFalse(pi2VarMap.containsKey("var2"));
+        assertThat(pi2VarMap)
+                .doesNotContainKey("var1")
```
filiphr, line 142: Shouldn't this be `containsKey`?

modules/flowable-engine/src/test/java/org/flowable/examples/bpmn/scripttask/ScriptTaskTest.java
dbmalkovsky: @filiphr and fixed here.

modules/flowable-engine/src/test/java/org/flowable/examples/bpmn/scripttask/ScriptTaskTest.java
dbmalkovsky: @filiphr OK; fixed.

modules/flowable-engine/src/test/java/org/flowable/examples/bpmn/servicetask/JavaServiceTaskTest.java
dbmalkovsky: @filiphr No, note that the original assertion was `assertFalse`.

filiphr: Yes you are 100% right. I only read the `containsKey`. My mistake sorry.

modules/flowable-engine/src/test/java/org/flowable/examples/bpmn/scripttask/ScriptTaskTest.java
filiphr: Thanks for both fixes

dbmalkovsky:
The JDK11 build fails with timing issues.

PR #2432, opened 2020-06-23
Update assertj fluent style in flowable-bpmn-converter module.

dbmalkovsky:
Updates to the `org.flowable.editor.language.xml` package to use more natural assertj style coding and remove unnecessary duplicate tests (e.g., testing for null by `isInstanceOf` or null before `hasSize`).
Also found one file that was added in April (`org.flowable.editor.language.xml.ExternalWorkerServiceTaskConverterTest`) that needed to be converted to assertj.

Originally thought there would be multiple PRs but the second half of the module was in much better shape so I just added the 4 additional files to this PR.

### DIFF
--- a/modules/flowable-bpmn-converter/src/test/java/org/flowable/editor/language/xml/AsyncEndEventConverterTest.java
+++ b/modules/flowable-bpmn-converter/src/test/java/org/flowable/editor/language/xml/AsyncEndEventConverterTest.java
@@ -15,8 +15,6 @@ package org.flowable.editor.language.xml;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.tuple;
 
-import java.util.List;
-
 import org.flowable.bpmn.model.BpmnModel;
 import org.flowable.bpmn.model.EndEvent;
 import org.flowable.bpmn.model.FlowElement;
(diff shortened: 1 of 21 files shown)


dbmalkovsky:
The failures are due to optimistic locking issues in one case and unresolved dependencies in another.

dbmalkovsky:
@filiphr I found that this construct was used in other of the test suites and I found that it made more sense instead of having to test the class type, create the local variable with a cast, and then test the local variable.

I'll stop making this type of change if that is what you want.   I'd rather not undo these though :)

filiphr:
> I found that this construct was used in other of the test suites and I found that it made more sense instead of having to test the class type, create the local variable with a cast, and then test the local variable.

I think it is a matter of taste. For me personally sometimes it makes sense to have it, but not always. I personally prefer not having the nesting because I can read it easier line by line.

dbmalkovsky:
@filiphr so what is the final verdict?  Leave it as is or undo the changes?

filiphr:
>so what is the final verdict? Leave it as is or undo the changes?

I'll manually merge this and omit the `isInstanceOfSatisfying`. You already did a lot of work on this. For the future let's try to use it less if possible. 

Thanks for the understanding.

filiphr:
Merged it manually in https://github.com/flowable/flowable-engine/commit/07f0ddcc36ceba112ad729a8658fc41cbf99f649

filiphr:
I left some if the `isInstanceOf` checks in `ValuedDataObjectConverterTest` because they were integer, long, double.

PR #2436, opened 2020-06-24
Update assertj fluent style in flowable-idm-engine module.

dbmalkovsky:
A very few updates/consistency changes for the `org.flowable.idm.engine.test.api.identity` package in assertj usage.


dbmalkovsky:
The JDK 14 build failed because of an unresolved dependency in javax.mail during the install.

PR #2443, opened 2020-06-25
Remove test @Ignore because Spring Boot is updated

dbmalkovsky:
Remove the `@Ignore` annotation from a test because Spring Boot is sufficiently past the release that contained the bug that required ignoring the test.


dbmalkovsky:
The JDK 11 build failed with the optimistic locking issue in the Engine module.

PR #2446, opened 2020-06-26
Use assertj fluent style in flowable-camel module.

dbmalkovsky:
Straightforward conversion of the module to assertj style once the jar was added to the POM.

### DIFF
--- a/modules/flowable-camel/src/test/java/org/flowable/camel/exception/CamelExceptionTest.java
+++ b/modules/flowable-camel/src/test/java/org/flowable/camel/exception/CamelExceptionTest.java
@@ -95,19 +95,12 @@ public class CamelExceptionTest extends SpringFlowableTestCase {
     public void testNonBpmnExceptionInCamel() {
         // Signal ThrowBpmnExceptionBean to throw a non BPMN Exception
         ThrowBpmnExceptionBean.setExceptionType(ThrowBpmnExceptionBean.ExceptionType.NON_BPMN_EXCEPTION);
-
-        try {
-            runtimeService.startProcessInstanceByKey("exceptionInRouteSynchron");
-        } catch (FlowableException e) {
-            assertThat(e.getCause().getClass()).isEqualTo(Exception.class);
-            assertThat(e.getCause().getMessage()).isEqualTo("arbitrary non bpmn exception");
-
-            assertThat(ExceptionServiceMock.isCalled()).isFalse();
-            assertThat(NoExceptionServiceMock.isCalled()).isFalse();
-
-            return;
-        }
-        fail("Activiti exception expected");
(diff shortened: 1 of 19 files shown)


modules/flowable-camel/src/test/java/org/flowable/camel/exception/CamelExceptionTest.java
```diff
@@ -95,11 +99,11 @@ public void testNonBpmnExceptionInCamel() {
         try {
             runtimeService.startProcessInstanceByKey("exceptionInRouteSynchron");
         } catch (FlowableException e) {
-            assertEquals(Exception.class, e.getCause().getClass());
```
filiphr, line 98: Why not do `assertThatThrownBy`? The `ExceptionServiceMock`  and `NoExceptionServiceMock` can be done after the `assertThatThrownBy`

modules/flowable-camel/src/test/java/org/flowable/camel/variables/CamelVariableTransferTest.java
```diff
@@ -156,15 +165,18 @@ public void testCamelPropertiesFiltered() throws Exception {
...
         Map<String, Object> variables = runtimeService.getVariables(task.getExecutionId());
-        assertEquals("sampleValueForProperty1", variables.get("property1"));
-        assertEquals("sampleValueForProperty2", variables.get("property2"));
-        assertNull(variables.get("property3"));
+        assertThat(variables)
+                .contains(
+                        entry("property1", "sampleValueForProperty1"),
+                        entry("property2", "sampleValueForProperty2")
+                )
+                .doesNotContain(entry("property3", null));
```
filiphr, line 179: This is not exactly the same assertion as before. This will pass if there is a variable named `property3` with some other value. I think that we should use `containsOnly` instead. This way we are validating all the properties which are in the process. 

Let's do this for all the places where we assert the variables in the `CamelVariableTransferTest`

modules/flowable-camel/src/test/java/org/flowable/camel/exception/CamelExceptionTest.java
dbmalkovsky: For some reason my eyes just jumped over that code block :(
Fixed.

modules/flowable-camel/src/test/java/org/flowable/camel/variables/CamelVariableTransferTest.java
dbmalkovsky: If it is OK with you I used `doesNotContainKeys()` instead to better match the original file.

dbmalkovsky:
The JDK11 failure is the typical `LongServiceTaskLoop` failure in the engine.

PR #2449, opened 2020-06-27
Use assertj fluent style in flowable-cdi module.

dbmalkovsky:
Straightforward conversion of the module to assertj style once the jar was added to the POM.


modules/flowable-cdi/src/test/java/org/flowable/cdi/test/impl/util/ProgrammaticBeanLookupTest.java
```diff
@@ -67,23 +64,23 @@ public static JavaArchive createDeploymentWithSpecialization() {
     public void testLookupBean() {
         deployer.deploy("normal");
         Object lookup = ProgrammaticBeanLookup.lookup("testOnly");
-        assertTrue(lookup.getClass().isAssignableFrom(TestBean.class));
+        assertThat(lookup.getClass().isAssignableFrom(TestBean.class));
```
filiphr, line 67: There is an assertion missing here.

dbmalkovsky: Sharp eyes; good catch; fixed.

PR #2456, opened 2020-06-28
Update assertj fluent style in flowable-content-rest module.

dbmalkovsky:
Update a file that was over looked when the module was originally converted to assertj.


dbmalkovsky:
The JDK11 failure is a dependency issue.

PR #2463, opened 2020-06-29
Update assertj fluent style in flowable-engine module.

dbmalkovsky:
In the process of making consistency changes in the flowable-engine module I found files that needed conversion to assertj in the `org.flowable.engine.test.bpmn` package.  

There are additional files for this module pending but in an attempt to keep the review manageable only this package is in this PR.


dbmalkovsky:
The JDK 14 test fails with the usual `Flowable6Test.testLongServiceTaskLoop` test failing.

PR #2476, opened 2020-07-01
Correct variable name

dbmalkovsky:
In the `if` test for `boolean` the first clause tests the ***value*** for `true`; the second clause tests the ***type*** for `false`.  As `expressionType` is the returned value and not the input this is obviously wrong.  The PR corrects the variable name in the second clause.


dbmalkovsky:
Once again the culprit is `Flowable6Test.testLongServiceTaskLoop` test this time in the JDK 14 build.

PR #2487, opened 2020-07-06
Replace nested "if" statements with "switch"

dbmalkovsky:
Since the code is post Java 7, replaced nested `if` statements with `switch` which may be faster and typically easier to read.


tijsrademakers:
@dbmalkovsky , we don't see this as an improvement. Of course this is a bit subjective, but we would like to stick with the if statements.

PR #2495, opened 2020-07-09

dbmalkovsky:
These 13 files finish the conversion of the `org.flowable.engine.test.bpmn` package to assertj fluent style.


dbmalkovsky:
The JDK8 failure is in the `flowable-spring-boot-sample-ldap` module well after the area of code that I modified.

PR #2499, opened 2020-07-10
Use assertj fluent style in flowable-ldap-configurator module.

dbmalkovsky:
Simple conversion of the `flowable-ldap-configurator` module once the `org.assertj` dependency was added to the `pom.xml` file.


tijsrademakers:
@dbmalkovsky can you have a look at the conflict?

dbmalkovsky:
@tijsrademakers I believe I resolved the conflict but the verification builds are queued and not running yet....

PR #2507, opened 2020-07-17
Value is already assigned

dbmalkovsky:
There is no need to assign the result to `decisionTableModel` as the method `DecisionTableModelConversionUtil.convertModelToV3` internally resets the parameter to that value.   It is all the more confusing as the method does set the variable and does return the variable.

The only other use of the method (found [here](https://github.com/flowable/flowable-engine/blob/master/modules/flowable-ui/flowable-ui-modeler-logic/src/main/java/org/flowable/ui/modeler/service/AppDefinitionImportService.java#L314))  does not use the returned variable value.

Found two other similar method calls this time for the `DecisionTableModelConversionUtil.convertModel` method; made the same type of change.


dbmalkovsky:
The JDK11 failure is a false negative as the issue is `Could not resolve dependencies for project org.flowable:flowable-http:jar:6.5.1-SNAPSHOT`.

PR #2509, opened 2020-07-17
japanese-translation

yumioka:
The very first version of Japanese messages

#### Check List:
* Unit tests:  NA
* Documentation: NA


tijsrademakers:
Thanks for the contribution!

PR #2522, opened 2020-07-25
Use helper method completeTask()

dbmalkovsky:
In `AbstractFlowableTestCase` which is the base class for many unit tests there is a helper method `completeTask(Task task)` that was used in some but not all situations in the tests so for consistency this PR uses the method were appropriate.  

In another PR I'll add a `completeTasks(List<Task> tasks)` method that will replace multiple sequential calls to `completeTask()`.


tijsrademakers:
Hi @dbmalkovsky , I understand why you added the helper method, but it also makes the unit tests a bit less readable because you would have to check which logic is executed in the completeTask method. Because it's only a one liner to complete the task, I'm going to close this PR.

dbmalkovsky:
HI @tijsrademakers, I didn't add the helper method, Dennis Federico did on July 5, 2018 into `org.flowable.engine.impl.test.AbstractFlowableTestCase`.  I saw that it was used 126 times so this PR just expanded it's usage.

PR #2532, opened 2020-07-28
Remove useless check

dbmalkovsky:
Remove useless check:

- `errors == null` can never be try because it is allocated above as `new ArrayList<>()`
- `numberOfProcesses == 0` can never be true as the value of `numberOfProcesses` must be greater than zero to get into the `while` loop
- if `i != 3` the only other value of `i` *is* 3

### DIFF
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/concurrency/ConcurrentEngineUsageTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/concurrency/ConcurrentEngineUsageTest.java
@@ -147,11 +147,9 @@ public class ConcurrentEngineUsageTest extends PluggableFlowableTestCase {
                     retryStartProcess(drivingUser);
                     finishTask = true;
 
-                    if (numberOfProcesses == 0) {
-                        // Make sure while-loop doesn't stop when processes are
-                        // all started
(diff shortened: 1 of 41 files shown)


modules/flowable-engine/src/test/java/org/flowable/engine/test/concurrency/ConcurrentEngineUsageTest.java
```diff
@@ -147,11 +147,9 @@ public void run() {
                     retryStartProcess(drivingUser);
                     finishTask = true;
 
-                    if (numberOfProcesses == 0) {
-                        // Make sure while-loop doesn't stop when processes are
-                        // all started
-                        tasksAvailable = taskService.createTaskQuery().taskAssignee(drivingUser).count() > 0;
-                    }
+                    // Make sure while-loop doesn't stop when processes are all started
```
tijsrademakers, line 150: Hi @dbmalkovsky , I don't think this change is correct. Now we are always testing if tasks are available and before it was only done when numberOfProcesses == 0. Maybe this whole if clause can be removed because numberOfProcesses is always > 0 in this part of the code?

dbmalkovsky:
Closing this request; I'll redo it without the change noted.

PR #2542, opened 2020-08-06
Various unit tests to increase coverage

dbmalkovsky:
In the process fixed some comments and updated an exception message that was too specific given its usage.


dbmalkovsky:
The JDK8 compile error makes no sense as the same construct that is failing is used just lines about the failures.

dbmalkovsky:
It appears that the recent changes for ConcurrentHashMap may have introduced NPEs.

PR #2556, opened 2020-08-24
Execute shutdown() to stop taskExecutor instead of start() in DefaultAsyncJobExecutor

vzickner:
#### Check List:
* Unit tests: NO
* Documentation: NA


tijsrademakers:
Thanks, nice catch

PR #2557, opened 2020-08-24
initialize bulk insert entities for multi-engine usage

vzickner:
Behavior before: in a multi-engine setup the bulk settings for the process engine
  was never configured since sessionFactories was not null and the
  initDbSqlSessionFactory() was never called during the ProcessEngine setup.

Behavior after: initDbSqlSessionFactoryEntitySettings() is always executed, as long
  as the process engine is configured with a relational database. With this it's
  ensured that the bulk insert is also happening for a multi-engine setup with
  pre-configured sessionFactories. This behavior is consistent with the behavior of
  other engines

#### Check List:
* Unit tests: NO
* Documentation: NA


tijsrademakers:
Hi Valentin, thanks for the PR. I've applied your change in a slightly different way and made sure it's also done for the other engines. Therefore closing this PR.

PR #2562, opened 2020-08-26
Feature/start process instance definition caching

vzickner:
Change process instance start for starting by process definition id to a cache based implementation. This is how it was before commit 6188d61eec3e190ae26245989398fa8d9f41c364.
Behavior for process definition key is not changed, since the `DeploymentManager` is not using the cache for key lookup.

#### Check List:
* Unit tests: NO
* Documentation: NO


tijsrademakers:
Thanks Valentin

PR #2590, opened 2020-09-24
Docs: includeCaseVariables instead of includecaseVariables

dbmalkovsky:
Fix for issue raised in this forum [post](https://forum.flowable.org/t/includecasevariables-in-docs-should-be-includecasevariables/6829): `includeCaseVariables` instead of `includecaseVariables`  (upper case the `C`).


filiphr:
THanks David

PR #2591, opened 2020-09-25
code checker

qiudaoke:
#### Check List:
* Unit tests: YES / NO / NA

### DIFF
--- a/modules/flowable-bpmn-model/src/main/java/org/flowable/bpmn/model/GraphicInfo.java
+++ b/modules/flowable-bpmn-model/src/main/java/org/flowable/bpmn/model/GraphicInfo.java
@@ -128,7 +128,7 @@ public class GraphicInfo {
     	}
 
     	// only check for elements that support this value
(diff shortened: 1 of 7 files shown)


modules/flowable-bpmn-model/src/main/java/org/flowable/bpmn/model/GraphicInfo.java
```diff
@@ -128,7 +128,7 @@ public boolean equals(GraphicInfo ginfo) {
     	}
 
     	// only check for elements that support this value
-    	if (null != this.getExpanded() && null != ginfo.getExpanded() && this.getExpanded() != ginfo.getExpanded()) {
+    	if (null != this.getExpanded() && null != ginfo.getExpanded() && !this.getExpanded().equals(ginfo.getExpanded())) {
```
tijsrademakers, line 131: Hi @qiudaoke , can you remove this change? Because the null checks done first ensure that a safe boolean != check can be done.

qiudaoke: ok, i will do it

PR #2618, opened 2020-09-27
Use uppercase l for long literal values

dbmalkovsky:
Instead of doing pieces here and there I grepped the source and changed all lower case `l` literal to uppercase `L` literal as needed.

In some cases it isn't even needed and as side benefit I improved one of the `assertThat()` tests.

As an added bonus I removed a single occurrence of a duplicated semi-colon.


dbmalkovsky:
The JDK8 failure was not in any of the code I changed.

PR #2620, opened 2020-09-29
Simple space related formatting changes

dbmalkovsky:
Simple add or remove spaces for formatting consistency.


filiphr:
Thanks for the PR @dbmalkovsky. It looks OK to me, but lets remove the change in the `MavenWrapperDownloader`. That class comes from the maven wrapper and I think it is good to stay in the format that it is in order to more easily review changes when it gets updated from the maven wrapper itself.

dbmalkovsky:
@filiphr Good catch; undid the changes.   My oversight in not noticing that it was part of the maven wrapper.  My bad ☹️

PR #2626, opened 2020-09-29
Github Actions: Replace Java 14 build with Java 15 build

PascalSchumacher:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [1.8, 11, 13, 14]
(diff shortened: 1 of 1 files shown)


PascalSchumacher:
Can not be merged, because Java 15 removed the java script implementation and some some flowable tests require it:

>org.flowable.common.engine.api.FlowableException: Can't find scripting engine for 'JavaScript'

filiphr:
Thanks for trying this out @PascalSchumacher. We need to find an alternative for Java 15. 

Perhaps we need to look at js-scriptengine from the GraalVM team. I found [this](https://golb.hplar.ch/2020/04/java-javascript-engine.html) blog that shows a possible migration path

PR #2629, opened 2020-10-05
Fix case of LOGGER and typos

dbmalkovsky:
Changed the case of the `logger` variable in  `private static final Logger logger` to all upper case to match all other uses. 

Also stumbled upon a few spelling/grammar mistakes and took the occasion to fix them.

### DIFF
--- a/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/interceptor/CmmnCommandInvoker.java
+++ b/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/interceptor/CmmnCommandInvoker.java
@@ -33,7 +33,7 @@ import org.slf4j.LoggerFactory;
  */
 public class CmmnCommandInvoker extends AbstractCommandInterceptor {
 
-    private static final Logger logger = LoggerFactory.getLogger(CmmnCommandInvoker.class);
+    private static final Logger LOGGER = LoggerFactory.getLogger(CmmnCommandInvoker.class);
 
     protected AgendaOperationRunner agendaOperationRunner;
 
@@ -76,9 +76,7 @@ public class CmmnCommandInvoker extends AbstractCommandInterceptor {
         if (runnable instanceof CmmnOperation) {
             CmmnOperation operation = (CmmnOperation) runnable;
 
-            if (logger.isDebugEnabled()) {
-                logger.debug("Executing agenda operation {}", runnable);
-            }
+            LOGGER.debug("Executing agenda operation {}", runnable);
 
             agendaOperationRunner.executeOperation(commandContext, runnable);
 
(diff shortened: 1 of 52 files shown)


modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/interceptor/CmmnCommandInvoker.java
```diff
@@ -76,9 +76,7 @@ protected void executeOperation(CommandContext commandContext, boolean isStoreCa
         if (runnable instanceof CmmnOperation) {
             CmmnOperation operation = (CmmnOperation) runnable;
 
-            if (logger.isDebugEnabled()) {
```
filiphr, line 79: Let's leave the if check here. When doing a call to a method with varArgs an array is instantiate on every call. This is really bad for performance and memory efficiency. Doing an if check on the other side is much better. 

This places is a hot path that would be affected by the unnecessary varArgs array creation.

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/parser/CmmnParserImpl.java
```diff
@@ -37,7 +37,7 @@
  */
 public class CmmnParserImpl implements CmmnParser {
 
-    private final Logger logger = LoggerFactory.getLogger(CmmnParserImpl.class);
+    private final Logger LOGGER = LoggerFactory.getLogger(CmmnParserImpl.class);
```
filiphr, line 40: This is an instance field, not a `static final` one. The logger should stay lowercase here.

modules/flowable-dmn-engine/src/main/java/org/flowable/dmn/engine/impl/interceptor/DmnCommandInvoker.java
```diff
@@ -56,9 +56,7 @@ protected void executeOperations(CommandContext commandContext, boolean isStoreC
         DmnEngineAgenda agenda = CommandContextUtil.getAgenda(commandContext);
         while (!agenda.isEmpty()) {
             Runnable runnable = agenda.getNextOperation();
-            if (logger.isDebugEnabled()) {
-                logger.debug("Executing agenda operation {}", runnable);
-            }
+            LOGGER.debug("Executing agenda operation {}", runnable);
```
filiphr, line 59: Same comment as for the `CmmnCommandInvoker`

filiphr (changes requested):
Thanks for the PR @dbmalkovsky. I've left some comments regarding some of the loggers.

modules/flowable-dmn-engine/src/main/java/org/flowable/dmn/engine/impl/interceptor/DmnCommandInvoker.java
dbmalkovsky: Done

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/parser/CmmnParserImpl.java
dbmalkovsky: Done

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/interceptor/CmmnCommandInvoker.java
dbmalkovsky: Done

dbmalkovsky:
The JDK 13 failure is an issue with fetching dependencies.

PR #2630, opened 2020-10-06
add defalut in switch

qiudaoke:
#### Check List:
* Unit tests: YES / NO / NA


modules/flowable-bpmn-converter/src/main/java/org/flowable/bpmn/converter/XMLStreamReaderUtil.java
```diff
@@ -55,6 +57,8 @@ public static boolean moveToEndOfElement(XMLStreamReader xtr, String elementName
                     if (xtr.getLocalName().equals(elementName))
                         return true;
                     break;
+                default:
```
dbmalkovsky, line 60: These 2 lines of code are already at master; you need to rebase your master branch before adding new code.

PR #2631, opened 2020-10-06
More typos

dbmalkovsky:
Fixed more typos.   There are many remaining issues but they are used either too often, are in model code, method names, or are file names.   There was even one case where I couldn't figure out what was meant; the word **cuby** in `org.flowable.engine.RuntimeService` where it reads (line 1061):
<pre>
.....  which are included in the cuby SQL directly.
</pre>

Anytime a correction was made to a message (e.g., exception message) I checked that it wasn't used anywhere else or if it was for example in a test I made the corresponding change.

The most **risky** changes were in the files `HistoryJobCollectionResource` which can be found in both the `flowable-rest` and `flowable-cmmn-rest` modules where the code reads (line 62):
<pre>
 @ApiImplicitParam(name = "scopeTyoe", dataType = "string", value = ...
</pre>
and I changed the **name** from **scopeTyoe** to **scopeType**
<pre>
 @ApiImplicitParam(name = "scopeType", dataType = "string", value = ...
</pre>

### DIFF
--- a/docs/docusaurus/docs/cmmn/ch08-REST.md
+++ b/docs/docusaurus/docs/cmmn/ch08-REST.md
@@ -1408,7 +1408,7 @@ The request should be of type +multipart/form-data+. There should be a single fi
 | - | - | - | - |
 |caseInstanceId|No|String|The case instance id for which the stage overview is retrieved.|    
 
-*Success reponse body* {
+*Success response body* {
 
 ```
 [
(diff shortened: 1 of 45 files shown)


filiphr (approved):
Thanks for doing the changes @dbmalkovsky.

The changes you mentioned as "risky" are OK as well. Those are only documentation changes which were not correct anyways. 

Will merge once the CI is done

Issue #2661, opened 2020-10-19
release 6.6.0 flowable-ui-modeler ExclusiveGateWay 


weweibuy:
ExclusiveGateWay missed a default folw checkbox.
There is the message content 'Exclusive gateway has at least one outgoing sequence flow without a condition (which isn't the default one)' when we check the model.
![image](https://user-images.githubusercontent.com/43676605/96471280-bdc71280-1261-11eb-8b35-077c990aac40.png)

douglascrp:
I am facing the same problem with the 6.6.0 version.

There is no way to set the value, as the checkbox is not present.

![image](https://user-images.githubusercontent.com/5930258/102806310-66e1d480-439b-11eb-8889-82359f3beb97.png)

filiphr:
The 6.6.0 version does not have the fix from PR #2662. I believe that this PR will fix this problem as well. Therefore, I am closing this issue

PR #2662, opened 2020-10-19
Fixing a modeler problem: property checkboxes are invisible

jaam:
Due to a wrong configuration boolean properties couldn't be modified in modeler because the checkbox component couldn't be rendered

#### Check List:
* Unit tests: NA
* Documentation: NA


douglascrp:
This is still a problem with the Gateway's Default Condition property, as described at #2661

filiphr:
@douglascrp this fix is not released. Did you try a snapshot version or are you talking about 6.6.0

douglascrp:
@filiphr Right... got it, so that is it, as I'm using the 6.6.0 version.

I am going to build my own snapshot version and give it a try.

Thank you.

douglascrp:
@filiphr indeed, the snapshot version fixes it.

Thank you

PR #2663, opened 2020-10-21
Use the EventRegistryEngineConfigurator produced by a method in ProcessEngineConfiguration

JacekMiszczak:
Currently ProcessEngineConfiguration always creates EventRegistryEngineConfigurator as the default event registry, even when using the SpringProcessEngineConfiguration, which overrides the createDefaultEventRegistryEngineConfigurator() method.

#### Check List:
* Unit tests: NA
* Documentation: NA


filiphr:
Good spot @JacekMiszczak, thanks for the PR. The `CmmnEngineConfiguration` had the same problem and I fixed it in https://github.com/flowable/flowable-engine/commit/2694a87ab4e51991e2276bc3a1715b32eee489e4

PR #2672, opened 2020-10-30
support of event payload field type 'json' for a xml event

amporsim:


tijsrademakers:
Closing this PR, let's look more into the use case etc.

PR #2665, opened 2020-10-27
Update to Spring 5.2.10.RELEASE

dbmalkovsky:

PR #2675, opened 2020-11-02
Use isAutoComplete variable

dbmalkovsky:
The code reads:
```
CompletionEvaluationResult completionEvaluationResult = PlanItemInstanceContainerUtil
            .shouldPlanItemContainerComplete(commandContext, caseInstance, true);
```

where the last parameter to the method is whether the container is autocomplete or not.

Note that the previous line of code reads:
```
boolean isAutoComplete = getPlanModel(caseInstance).isAutoComplete();
```
but the variable `isAutoComplete` is never used and thus a dead store.  I assumed that instead of always passing `true` the variable should be used in its place.

I didn't find a test that exercised the code directly.


dbmalkovsky:
The JDK 8 failure seems like the same error loop handling error that was in [PR 2665](https://github.com/flowable/flowable-engine/pull/2665)

jbarrez:
Good catch, thanks!

PR #2676, opened 2020-11-02
Fix exception message text

dbmalkovsky:
The exception message is attempting to output the id but is using the object that wasn't found and thus null.


dbmalkovsky:
Looks like yet another failure (the same?) that is not the result of a message change.

dbmalkovsky:
I retriggered the CI build and now another JDK is failing ... oh well.

PR #2684, opened 2020-11-04
Update AssertJ to version 3.18.0.

PascalSchumacher:
I had to update some assertions in order to make test pass with the new version.


PascalSchumacher:
Need some work. I will update it in the next few days.

PR #2694, opened 2020-11-08
Remove the xercesImpl dependency and use the SAX/StAX implementation …

PascalSchumacher:
…supplied by Java instead.

Just an idea.

Feel free to close this, if you think it is a bad idea.


PascalSchumacher:
I am closing this, because accessing the "JDK’s internal copy of the Xerces XML processor" is discouraged and will probably no longer possible on Java 16+ (see: https://openjdk.java.net/jeps/396).

PR #2698, opened 2020-11-09
Use processDefinitionId to avoid NPE

dbmalkovsky:
See [comment in commit](https://github.com/flowable/flowable-engine/commit/e43101f1aaa2b463d6a71a21419c8ada69b70663#r44002698) which reads in part:

       This is going to throw a NPE as processDefinition is null (line 1391); I believe instead of processDefinition.getId() it should just be processDefinitionId


tijsrademakers:
Thanks David for the quick fix

PR #2704, opened 2020-11-13
Fix MultiInstanceActivityBehavior wrt renamed variable names

BorisKlinkerSAP:
We were trying to rename default loop variable names by an own extended implementation of MultiInstanceActivityBehavior which overwrites the methods setLoopVariable(), getLoopVariable() and getLocalLoopVariable(), renaming the variables and delegating to their overwritten implementations.

This works well except in case of MultiInstanceActivityBehavior.buildCompletedEvent() because it invokes directly DelegateExecution.getVariable() utilizing final constant values. There is no chance to intercept there smoothly.

This Pull Request changes this part of code to also use (as already all other places) getLoopVariable() which can be easily overwritten by extending classes.

The added test verifies that the created events carry the expected variable values.

#### Check List:
* Unit tests: YES
* Documentation: NA


tijsrademakers:
@BorisKlinkerSAP , merged the PR in a separate commit.

PR #2705, opened 2020-11-13
Upgrade to Spring Boot 2.3.6.RELEASE

dbmalkovsky:


PascalSchumacher:
Hi David,

did you heed the comment 

`<!-- When updating one spring version, make sure that all of them are updated to their latest compatible versions -->`

and did check if any spring dependencies should be updated to match the version used by spring boot?

dbmalkovsky:
It appears that only 2 dependent jars are needed for this update.  One is the Spring FrameWork which is updated in another of my PRs and Kafka.   The second commit here updates Kafka.

PR #2717, opened 2020-11-21
Add includeProcessVariables in CMMN rest api

sina-al:
#### Check List:
* Unit tests: YES / NO / NA

### DIFF
--- a/modules/flowable-cmmn-rest/src/main/java/org/flowable/cmmn/rest/service/api/history/task/HistoricTaskInstanceBaseResource.java
+++ b/modules/flowable-cmmn-rest/src/main/java/org/flowable/cmmn/rest/service/api/history/task/HistoricTaskInstanceBaseResource.java
@@ -218,9 +218,9 @@ public class HistoricTaskInstanceBaseResource {
             }
         }
 
-        if (request.getIncludeProcessVariables() != null) {
-            if (request.getIncludeProcessVariables()) {
-                taskQuery.includeProcessVariables();
+        if (queryRequest.getIncludeProcessVariables() != null) {
+            if (queryRequest.getIncludeProcessVariables()) {
+                query.includeProcessVariables();
             }
         }
 
--- a/modules/flowable-cmmn-rest/src/main/java/org/flowable/cmmn/rest/service/api/history/task/HistoricTaskInstanceBaseResource.java
+++ b/modules/flowable-cmmn-rest/src/main/java/org/flowable/cmmn/rest/service/api/history/task/HistoricTaskInstanceBaseResource.java
@@ -218,6 +218,12 @@ public class HistoricTaskInstanceBaseResource {
             }
         }
 
+        if (request.getIncludeProcessVariables() != null) {
+            if (request.getIncludeProcessVariables()) {
+                taskQuery.includeProcessVariables();
+            }
+        }
+
         if (queryRequest.getTaskVariables() != null) {
             addTaskVariables(query, queryRequest.getTaskVariables());
         }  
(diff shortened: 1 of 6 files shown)


modules/flowable-cmmn-rest/src/main/java/org/flowable/cmmn/rest/service/api/history/task/HistoricTaskInstanceBaseResource.java
```diff
@@ -218,6 +218,12 @@
             }
         }
 
+        if (request.getIncludeProcessVariables() != null) {
```
dbmalkovsky, line 221: I think `request` should be `queryRequest` instead.

modules/flowable-cmmn-rest/src/main/java/org/flowable/cmmn/rest/service/api/history/task/HistoricTaskInstanceCollectionResource.java
```diff
@@ -225,6 +226,10 @@
             queryRequest.setIncludeTaskLocalVariables(Boolean.valueOf(allRequestParams.get("includeTaskLocalVariables")));
         }
 
+        if (requestParams.containsKey("includeProcessVariables")) {
```
dbmalkovsky, line 229: Shouldn't `requestParams` be `allRequestParams` instead?

sina-al: Indeed 2230030567c3621503a19ff5cfe6eb5f30b56876

modules/flowable-cmmn-rest/src/main/java/org/flowable/cmmn/rest/service/api/history/task/HistoricTaskInstanceBaseResource.java
sina-al: Yes it should 2230030

tijsrademakers:
We don't want to expose this in the CMMN REST API, this is available in the BPMN REST API already.

sina-al:
Hey @tijsrademakers 

What would be the ideal way, via the BPMN REST API, to query for runtime tasks where

- `caseInstanceIdWithChildren` is specified
- `includeProcesVariables` is true

I believe `caseInstanceIdWithChildren` filter is available in on `GET /cmmn-api/cmmn-runtime/tasks`  but on in `GET /bpmn-api/runtime/tasks`

Likewise, I believe `includeProcesVariables` is available in on `GET /bpmn-api/runtime/tasks`  but on in `GET /cmmn-api/cmmn-runtime/tasks`

(at the time this PR was opened)


My current workaround is to do N+1 requests to fit my usecase.

Is there a better approach? Or do you would you say there's something off with this use case in general?

tijsrademakers:
Hi @sina-al,

Thanks for your feedback.
Is it correct that you are looking for a way to get all tasks of a specific case instance where the process tasks from a sub process instance have the process variables included?

Best regards

sina-al:
Hey  @tijsrademakers thanks for the response.

> Is it correct that you are looking for a way to get all tasks of a specific case instance where the process tasks from a sub process instance have the process variables included?

Yeah that's a spot on description of what I'm looking for.

I've also explored inheriting variables on the sub processes instances so that I can filter by them, but when I saw `caseInstanceIdWithChildren` on the `GET /cmmn-api/cmmn-runtime/tasks` endpoint it seemed like the more natural route to go with.

Best regards

tijsrademakers:
Thanks for the additional feedback, I've merged the PR.

PR #2726, opened 2020-11-28
remove IDM from exception inside the AbstractEngineConfigurator

vzickner:


modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/AbstractEngineConfigurator.java
```diff
@@ -254,7 +254,7 @@ protected void initDataSource(AbstractEngineConfiguration engineConfiguration, A
         if (engineConfiguration.getDataSource() != null) {
             targetEngineConfiguration.setDataSource(engineConfiguration.getDataSource());
         } else {
-            throw new FlowableException("A datasource is required for initializing the IDM engine ");
+            throw new FlowableException("A datasource is required for initializing the engine ");
```
dbmalkovsky, line 257: How about this instead; it will identify the engine:
````
throw new FlowableException("A datasource is required for initializing the " + engineConfiguration.getEngineName() + " engine");
````

filiphr: If we are doing something like this then engine type or getting the current class is better. Engine name is the same for all engines by default

vzickner: The one time I have had that exception it was easy to identify the correct engine through the stack trace. The only issue was that the message was confusing to the user.

PR #2730, opened 2020-12-03
Introduce flag to control error messages being sent back with 500

anatoliy-balakirev:
Today Flowable by default returns detailed error messages for all errors
For errors 500, however, responses contain information about software
exceptions with detailed error description if an error occurs. Such
information is revealed at multiple endpoints in case of invalid input
parameters.
Technical information is not supposed to be visible to clients as it
does not provide any added value. To an attacker however, this
information might give valuable hints as to which frameworks and
libraries are used by the web application. Apart from that, technical
details can help in the analysis of errors and simplify the search for
security issues.

So to avoid this, the default behavior was changed not to return
detailed messages for all 500 errors. That default can also be
overridden by the newly introduced flag:
`flowable.common.rest.exception.return-messages-for-500-errors`

#### Check List:
* Unit tests: NA
* Documentation: NO


filiphr:
Closing this since this has been done in some way in 80b0ff21fc615a676c14da99853e26a980397bdf

PR #2734, opened 2020-12-04
Use newly created variable in place of null value

dbmalkovsky:
The variable, `historicTaskInstance` is null that is being sent to `setEndProperties`.  Use the newly created `historicTaskInstanceEntity` instead.


dbmalkovsky:
This PR is the result of an email exchange with @jbarrez

PR #2749, opened 2020-12-18
Minor cleanup of CallActivityBehavior

AlexanderWirthSAP:
* Declare "variables" HashMap inline to avoid it being accidentally used further down
* Use already retrieved eventDispatcher instead of fetching it again from process engine configuration

#### Check List:
* Unit tests: NA
* Documentation: NA


dbmalkovsky:
Both failures are the result of issues with Mule jar dependencies.

AlexanderWirthSAP:
Can I retrigger the builds somehow?

filiphr:
You don't have to do that @AlexanderWirthSAP. We are always checking the builds and if something is not OK we let people know

PascalSchumacher:
>Can I retrigger the builds somehow?

You can retrigger the builds by amending the commit and then force pushing.

PR #2750, opened 2020-12-21
Implement exception logging logic from CommandContext for async jobs as well

reva:
There is this interesting feature on FlowableException not to log the exception which works for sync cases - this is an attempt to get the same default behaviour for the async cases.

### DIFF
--- a/modules/flowable-job-service/src/main/java/org/flowable/job/service/impl/asyncexecutor/DefaultAsyncRunnableExecutionExceptionHandler.java
+++ b/modules/flowable-job-service/src/main/java/org/flowable/job/service/impl/asyncexecutor/DefaultAsyncRunnableExecutionExceptionHandler.java
@@ -12,6 +12,8 @@
  */
 package org.flowable.job.service.impl.asyncexecutor;
 
+import org.flowable.common.engine.api.FlowableException;
+import org.flowable.common.engine.api.FlowableOptimisticLockingException;
 import org.flowable.common.engine.api.delegate.event.FlowableEngineEventType;
 import org.flowable.common.engine.api.delegate.event.FlowableEventDispatcher;
 import org.flowable.common.engine.impl.interceptor.Command;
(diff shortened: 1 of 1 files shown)


reva:
Nevermind, the default of this flag is `false` so that would not be a good idea to have it as a default...

reva:
Alternative if you don't want specific exceptions logged in async jobs, you could use a logging filter e.g. for logback:

```
import ch.qos.logback.classic.spi.ILoggingEvent;
import ch.qos.logback.classic.spi.IThrowableProxy;
import ch.qos.logback.classic.spi.ThrowableProxy;
import ch.qos.logback.core.filter.Filter;
import ch.qos.logback.core.spi.FilterReply;

public class FlowableExceptionFilter extends Filter<ILoggingEvent> {

    @Override
    public FilterReply decide(ILoggingEvent iLoggingEvent) {
        IThrowableProxy throwableProxy = iLoggingEvent.getThrowableProxy();
        if (!(throwableProxy instanceof ThrowableProxy)) {
            return FilterReply.NEUTRAL;
        }

        Throwable throwable = ((ThrowableProxy) throwableProxy).getThrowable();

        if ((throwable instanceof FlowableExceptionNotLogged)) {
            return FilterReply.DENY;
        }

        return FilterReply.NEUTRAL;
    }

}
```

Otherwise using the default spring boot logback settings, the configuration could look like this:
```
<?xml version="1.0" encoding="UTF-8"?>
<configuration>
    <include resource="org/springframework/boot/logging/logback/defaults.xml"/>
    <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
        <encoder>
            <pattern>${CONSOLE_LOG_PATTERN}</pattern>
        </encoder>
        <filter class="your.package.FlowableExceptionFilter"/>
    </appender>
    <root level="INFO">
        <appender-ref ref="CONSOLE" />
    </root>
    <logger name="org.springframework.web" level="DEBUG"/>
</configuration>
```

PR #2753, opened 2020-12-23
Added new filter to HistoricTaskInstanceQuery

parth-lunagariya:
Added new filter taskAssigned(filtering task where assignee is not null) and taskUnassigned(filtering task where assignee is null) for HistoricTaskInstanceQuery

### DIFF
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/task/TaskQueryTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/task/TaskQueryTest.java
@@ -1279,8 +1279,8 @@ public class TaskQueryTest extends PluggableFlowableTestCase {
     @Test
     public void testQueryByAssignedOr() {
         TaskQuery query = taskService.createTaskQuery().or().taskId("invalid").taskAssigned();
-        assertThat(query.count()).isEqualTo(0);
-        assertThat(query.list()).hasSize(0);
+        assertThat(query.count()).isEqualTo(1);
+        assertThat(query.list()).hasSize(1);
     }
 
     @Test
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/task/TaskQueryTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/task/TaskQueryTest.java
@@ -1268,6 +1268,20 @@ public class TaskQueryTest extends PluggableFlowableTestCase {
         assertThat(query.count()).isEqualTo(11);
         assertThat(query.list()).hasSize(11);
     }
+    
+    @Test
+    public void testQueryByAssigned() {
+        TaskQuery query = taskService.createTaskQuery().taskAssigned();
+        assertThat(query.count()).isEqualTo(1);
+        assertThat(query.list()).hasSize(1);
+    }
+
+    @Test
+    public void testQueryByAssignedOr() {
+        TaskQuery query = taskService.createTaskQuery().or().taskId("invalid").taskAssigned();
+        assertThat(query.count()).isEqualTo(0);
+        assertThat(query.list()).hasSize(0);
+    }
 
     @Test
     public void testQueryByCandidateUser() {
--- a/modules/flowable-task-service/src/main/resources/org/flowable/task/service/db/mapping/entity/Task.xml
+++ b/modules/flowable-task-service/src/main/resources/org/flowable/task/service/db/mapping/entity/Task.xml
@@ -660,6 +660,9 @@
       <if test="unassigned">
         and RES.ASSIGNEE_ IS NULL
       </if>
+      <if test="withAssignee">
+        and RES.ASSIGNEE_ IS NOT NULL
+      </if>
       <if test="noDelegationState">
         and RES.DELEGATION_ IS NULL
       </if>
@@ -1123,6 +1126,9 @@
             <if test="orQueryObject.unassigned">
               or RES.ASSIGNEE_ IS NULL
             </if>
+            <if test="orQueryObject.withAssignee">
+              and RES.ASSIGNEE_ IS NOT NULL
+      		</if>
             <if test="orQueryObject.noDelegationState">
               or RES.DELEGATION_ IS NULL
             </if>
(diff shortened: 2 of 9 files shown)


filiphr:
Thanks a lot for the PR @parth-lunagariya. I think that this is a useful addition.

However, I think that it makes more sense to add this to the `TaskInfoQuery` so it is available for runtime tasks as well. Additionally, would you mind adding some tests for this as well. You can add them to [`TaskInfoQueryTest`](https://github.com/flowable/flowable-engine/blob/master/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/task/TaskInfoQueryTest.java) and [`HistoricTaskInstanceTest`](https://github.com/flowable/flowable-engine/blob/master/modules/flowable-engine/src/test/java/org/flowable/engine/test/history/HistoricTaskInstanceTest.java).

Finally, if you want to make this available through the REST API you can have a look the the usages of [`TaskBaseResource#getTasksFromQueryRequest`](https://github.com/flowable/flowable-engine/blob/9a102f654d8280e93b871d3e21400645707741fc/modules/flowable-rest/src/main/java/org/flowable/rest/service/api/runtime/task/TaskBaseResource.java#L127) and [`HistoricTaskInstanceBaseResource#getQueryResponse`](https://github.com/flowable/flowable-engine/blob/9a102f654d8280e93b871d3e21400645707741fc/modules/flowable-rest/src/main/java/org/flowable/rest/service/api/history/HistoricTaskInstanceBaseResource.java#L74) for the BPMN engine and [`TaskBaseResource#getTasksFromQueryRequest`](https://github.com/flowable/flowable-engine/blob/9a102f654d8280e93b871d3e21400645707741fc/modules/flowable-cmmn-rest/src/main/java/org/flowable/cmmn/rest/service/api/runtime/task/TaskBaseResource.java#L127) and [`HistoricTaskInstanceBaseResource#getQueryResponse`](https://github.com/flowable/flowable-engine/blob/9a102f654d8280e93b871d3e21400645707741fc/modules/flowable-cmmn-rest/src/main/java/org/flowable/cmmn/rest/service/api/history/task/HistoricTaskInstanceBaseResource.java#L74) for the CMMN engine.

parth-lunagariya:
Hi @filiphr ,

I haven't added those filter in TaskInfoQuery because runtime tasks already have similar filter "taskUnassigned" in TaskQuery.java file which is i think specific for runtime task and now I want that for historic task also for my project. I haven't added in REST API as we are not using REST API and i might break something in REST API.

I will add required test cases for HistoricTaskInstanceTest.

filiphr:
I forgot about the `TaskQuery#taskUnassigned` thanks for reminding me. In that case I would suggest moving method in the `TaskInfoQuery` as the implementation logic is identical. The method should instead return `T` (the generic argument of the `TaskInfoQuery`).

Let's also move `taskAssigned` to the `TaskInfoQuery`, because people will surely want to use that as well 😄.

No problem for the REST API, we'll have a look at it once you are done.

parth-lunagariya:
ok @filiphr, I will move both methods to TaskInfoQuery as you said and I will also change necessary test cases.

parth-lunagariya:
I have moved both methods to TaskInfoQuery. I also changed required test cases. I haven't change TaskInfoQueryTest as it looks like more generic test case. so I added test case in test file of implementation of that query(TaskQueryTest and HistoricTaskInstanceTest).

@filiphr  please review this code changes.

parth-lunagariya:
> I have moved both methods to TaskInfoQuery. I also changed required test cases. I haven't change TaskInfoQueryTest as it looks like more generic test case. so I added test case in test file of implementation of that query(TaskQueryTest and HistoricTaskInstanceTest).
> 
> @filiphr please review this code changes.

Hi @filiphr , I did required code changes. Let me know any more changes required.

modules/flowable-engine/src/test/java/org/flowable/engine/test/api/task/TaskQueryTest.java
```diff
@@ -1268,6 +1268,20 @@ public void testQueryByUnassignedOr() {
...
+    public void testQueryByAssigned() {
+        TaskQuery query = taskService.createTaskQuery().taskAssigned();
+        assertThat(query.count()).isEqualTo(1);
+        assertThat(query.list()).hasSize(1);
+    }
+
+    @Test
+    public void testQueryByAssignedOr() {
+        TaskQuery query = taskService.createTaskQuery().or().taskId("invalid").taskAssigned();
+        assertThat(query.count()).isEqualTo(0);
```
filiphr, line 1282: It seems like the `or` is not working correctly. This should have a count of `1` since there is 1 assigned task.

modules/flowable-task-service/src/main/resources/org/flowable/task/service/db/mapping/entity/Task.xml
```diff
@@ -1123,6 +1126,9 @@
             <if test="orQueryObject.unassigned">
               or RES.ASSIGNEE_ IS NULL
             </if>
+            <if test="orQueryObject.withAssignee">
+              and RES.ASSIGNEE_ IS NOT NULL
```
filiphr, line 1130: This should be an `or` and not an `and`.

filiphr:
@parth-lunagariya, I was on holidays, that's why the slow response.

Anyways, I've left 2 comments, can you please do those fixes and then we can merge this.

modules/flowable-task-service/src/main/resources/org/flowable/task/service/db/mapping/entity/Task.xml
parth-lunagariya: ohhh. i will fix this issue and let you know @filiphr.

https://github.com/flowable/flowable-engine/pull/2756 this PR is also created by me. can you please look at that?

modules/flowable-engine/src/test/java/org/flowable/engine/test/api/task/TaskQueryTest.java
parth-lunagariya: Fixed above issue.

modules/flowable-task-service/src/main/resources/org/flowable/task/service/db/mapping/entity/Task.xml
parth-lunagariya: Fixed above issue.

parth-lunagariya:
@filiphr , I made requested code changes. Please review.

filiphr:
Thanks for the changes @parth-lunagariya. I've merged this manually in https://github.com/flowable/flowable-engine/commit/0847d9dd8f5739bc2503a7bd80598597411a1190

PR #2756, opened 2020-12-30
Fixed : Validation error is thrown when "to" is not given but "bcc" or "cc" is given in email task

parthlunagariya:
Fixed below issue : 
"No recipient is defined on the mail activity" validation message is shown when "to" field is not provided and "cc" or "bcc" is provided in email task.

Normally, user can send email by just providing value in "cc" and "bcc".

so i have changed validation code for email task and now validation error won't be given if any one of them is provided.

### DIFF
--- a/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/behavior/impl/MailActivityBehavior.java
+++ b/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/behavior/impl/MailActivityBehavior.java
@@ -97,6 +97,7 @@ public class MailActivityBehavior extends CoreCmmnActivityBehavior {
 
             email = createEmail(textStr, htmlStr, attachmentsExist(files, dataSources));
             addHeader(email, headersStr);
+            validateToCcBcc(commandContext, toStr, ccStr, bccStr, planItemInstanceEntity.getTenantId());
             addTo(commandContext, email, toStr, planItemInstanceEntity.getTenantId());
             setFrom(commandContext, email, fromStr, planItemInstanceEntity.getTenantId());
             addCc(commandContext, email, ccStr, planItemInstanceEntity.getTenantId());
@@ -183,11 +184,38 @@ public class MailActivityBehavior extends CoreCmmnActivityBehavior {
             throw new FlowableException("Could not create text-only email", e);
         }
     }
+    
+    protected void validateToCcBcc(CommandContext commandContext, String to, String cc, String bcc, String tenantId) {
+        
+        if(to == null && cc == null && bcc == null) {
+            throw new FlowableException("No recipient could be found for sending email");
+        }
+        
+        String newTo, newCc, newBcc;
+        newTo = newCc = newBcc = getForceTo(commandContext, tenantId);
+        if (newTo == null) {
+            newTo = to;
+        }
+        
+        String[] tos = splitAndTrim(newTo);
+        if (newCc == null) {
+            newCc = cc;
+        }
+        
+        String[] ccs = splitAndTrim(newCc);
+        if (newBcc == null) {
+            newBcc = bcc;
+        }
+        
+        String[] bccs = splitAndTrim(newBcc);
+        if(tos == null && ccs == null && bccs == null) {
+            throw new FlowableException("No recipient could be found for sending email");
+        }
+    }
 
     protected void addTo(CommandContext commandContext, Email email, String to, String tenantId) {
         if (to == null) {
-            // To has to be set, otherwise it can fallback to the forced To and then it won't be noticed early on
-            throw new FlowableException("No recipient could be found for sending email");
+           return;
         }
         String newTo = getForceTo(commandContext, tenantId);
         if (newTo == null) {
@@ -202,8 +230,6 @@ public class MailActivityBehavior extends CoreCmmnActivityBehavior {
                     throw new FlowableException("Could not add " + t + " as recipient", e);
                 }
             }
-        } else {
-            throw new FlowableException("No recipient could be found for sending email");
         }
     }
 
--- a/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/behavior/impl/MailActivityBehavior.java
+++ b/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/behavior/impl/MailActivityBehavior.java
@@ -187,7 +187,7 @@ public class MailActivityBehavior extends CoreCmmnActivityBehavior {
     
     protected void validateToCcBcc(CommandContext commandContext, String to, String cc, String bcc, String tenantId) {
         
-        if(to == null && cc == null && bcc == null) {
+        if (to == null && cc == null && bcc == null) {
             throw new FlowableException("No recipient could be found for sending email");
         }
         
@@ -208,7 +208,7 @@ public class MailActivityBehavior extends CoreCmmnActivityBehavior {
         }
         
         String[] bccs = splitAndTrim(newBcc);
-        if(tos == null && ccs == null && bccs == null) {
+        if (tos == null && ccs == null && bccs == null) {
             throw new FlowableException("No recipient could be found for sending email");
         }
     }
--- a/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/behavior/impl/MailActivityBehavior.java
+++ b/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/behavior/impl/MailActivityBehavior.java
@@ -97,7 +97,7 @@ public class MailActivityBehavior extends CoreCmmnActivityBehavior {
 
             email = createEmail(textStr, htmlStr, attachmentsExist(files, dataSources));
             addHeader(email, headersStr);
-            validateToCcBcc(commandContext, toStr, ccStr, bccStr, planItemInstanceEntity.getTenantId());
+            validateToCcBcc(toStr, ccStr, bccStr);
             addTo(commandContext, email, toStr, planItemInstanceEntity.getTenantId());
             setFrom(commandContext, email, fromStr, planItemInstanceEntity.getTenantId());
             addCc(commandContext, email, ccStr, planItemInstanceEntity.getTenantId());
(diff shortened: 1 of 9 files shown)


filiphr (changes requested):
Thanks for the PR @parthlunagariya.

Unfortunately, this is not enough as it will lead to a runtime exception in the [`MailActivityBehavor`](https://github.com/flowable/flowable-engine/blob/master/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/MailActivityBehavior.java). More precisely [here](https://github.com/flowable/flowable-engine/blob/9a102f654d8280e93b871d3e21400645707741fc/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/MailActivityBehavior.java#L200).

Can you please address that (the same is valid for the CMMN activity).

Additionally can you also add a test for the validation in the [`DefaultProcessValidatorTest`](https://github.com/flowable/flowable-engine/blob/master/modules/flowable-engine/src/test/java/org/flowable/standalone/validation/DefaultProcessValidatorTest.java). A new BPMN XML with 4 mail tasks should be enough (one without anything, one with only `to`, one with only `cc` and one with only `bcc`).

We will need similar tests for the runtime execution as well.

parthlunagariya:
Hi @filiphr ,

I looked into MailActivityBehavor code. There is validation only for To field and for CC and BCC, no validation error is thrown. All three methods are separate. So, I want to confirm approach with you first. I will change addTo() method and will make it similar to addCc and addBcc method and I will add separate method for validation before these 3 methods which will check for all 3 fields. Let me know what you think about this.

If i do similar change in CMMN activity, do i have to write separate test cases? if yes, then please give me name of file for the same.

In DefaultProcessValidatorTest, you want me to add new test method with new BPMN XML with 4 mail tasks. right??

filiphr:
>I looked into MailActivityBehavor code. There is validation only for To field and for CC and BCC, no validation error is thrown. All three methods are separate. So, I want to confirm approach with you first. I will change addTo() method and will make it similar to addCc and addBcc method and I will add separate method for validation before these 3 methods which will check for all 3 fields. Let me know what you think about this.


Yes adding a new method for validating the input is the approach that I would take as well. The method should take the `toStr`, `ccStr` and `bccStr` input parameters. For the `addTo` method make sure that the `forceTo` is applied before returning in that method.

>If i do similar change in CMMN activity, do i have to write separate test cases? if yes, then please give me name of file for the same.

Yes we need separate tests for the CMMN activity as well. Have a look at [`EmailServiceTaskTest`](https://github.com/flowable/flowable-engine/blob/master/modules/flowable-engine/src/test/java/org/flowable/examples/bpmn/mail/EmailServiceTaskTest.java) and [`EmailSendTaskTest`](https://github.com/flowable/flowable-engine/blob/master/modules/flowable-engine/src/test/java/org/flowable/examples/bpmn/mail/EmailSendTaskTest.java) for the BPMN Engine and the [`CmmnMailTaskTest`](https://github.com/flowable/flowable-engine/blob/master/modules/flowable-cmmn-engine/src/test/java/org/flowable/cmmn/test/task/CmmnMailTaskTest.java) for the CMMN engine.

>In DefaultProcessValidatorTest, you want me to add new test method with new BPMN XML with 4 mail tasks. right??

Yes the idea was to add a new test method with new BPMN XML with 4 mail tasks. Similar to how it is done for the `testExternalWorkerTaskWithoutTopic`

parthlunagariya:
ok. Thanks @filiphr . I will do required changes.

parthlunagariya:
> Hi @filiphr ,
> 
> I looked into MailActivityBehavor code. There is validation only for To field and for CC and BCC, no validation error is thrown. All three methods are separate. So, I want to confirm approach with you first. I will change addTo() method and will make it similar to addCc and addBcc method and I will add separate method for validation before these 3 methods which will check for all 3 fields. Let me know what you think about this.
> 
> If i do similar change in CMMN activity, do i have to write separate test cases? if yes, then please give me name of file for the same.
> 
> In DefaultProcessValidatorTest, you want me to add new test method with new BPMN XML with 4 mail tasks. right??

I did requested code changes and I also added test case in DefaultProcessValidatorTest with xml containing 4 mail task as mentioned in comment above.

I looked at look at EmailServiceTaskTest, EmailSendTaskTest and CmmnMailTaskTest and executed them. No need to change anything in these tests as they are getting executed successfully and it also covers new validation method because validation was already there but it was in separate method for each field (to,cc and bcc) but now we have added a common method. So I haven't changed these test cases.

@filiphr , Please review this changes.

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/behavior/impl/MailActivityBehavior.java
```diff
@@ -183,11 +184,38 @@ protected MultiPartEmail createMultiPartEmail(String text) {
...
+            newCc = cc;
+        }
+        
+        String[] ccs = splitAndTrim(newCc);
+        if (newBcc == null) {
+            newBcc = bcc;
+        }
+        
+        String[] bccs = splitAndTrim(newBcc);
+        if(tos == null && ccs == null && bccs == null) {
```
dbmalkovsky, line 211: This is only a style comment: please use `if (` instead of `if(`.  You have both versions in this PR.   Thanks

parthlunagariya: @dbmalkovsky , I did above styling changes in both MailActivityBehavior files. Thanks.

parthlunagariya:
@filiphr  , Please review and let me know if any more changes required.

filiphr (changes requested):
Thanks for the changes @parthlunagariya.

I've left some comments (they are also valid for the Cmmn `MailActivityBehavour`).

>I looked at EmailServiceTaskTest, EmailSendTaskTest and CmmnMailTaskTest and executed them. No need to change anything in these tests as they are getting executed successfully and it also covers new validation method because validation was already there but it was in separate method for each field (to,cc and bcc) but now we have added a common method. So I haven't changed these test cases.

I think we need to add new tests cases there that would test sending out emails only with `cc` and only with `bcc` defined, since this is new. It wasn't tested before (it would have failed), but it is always good to have more tests 😉

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/behavior/impl/MailActivityBehavior.java
```diff
@@ -183,11 +184,38 @@ protected MultiPartEmail createMultiPartEmail(String text) {
...
+    
+    protected void validateToCcBcc(CommandContext commandContext, String to, String cc, String bcc, String tenantId) {
+        
+        if (to == null && cc == null && bcc == null) {
+            throw new FlowableException("No recipient could be found for sending email");
+        }
+        
+        String newTo, newCc, newBcc;
+        newTo = newCc = newBcc = getForceTo(commandContext, tenantId);
+        if (newTo == null) {
```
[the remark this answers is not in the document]
parth-lunagariya: Fixed this.

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/behavior/impl/MailActivityBehavior.java
```diff
@@ -183,11 +184,38 @@ protected MultiPartEmail createMultiPartEmail(String text) {
...
+        if (tos == null && ccs == null && bccs == null) {
+            throw new FlowableException("No recipient could be found for sending email");
+        }
+    }
 
     protected void addTo(CommandContext commandContext, Email email, String to, String tenantId) {
         if (to == null) {
-            // To has to be set, otherwise it can fallback to the forced To and then it won't be noticed early on
-            throw new FlowableException("No recipient could be found for sending email");
+           return;
```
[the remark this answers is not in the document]
parth-lunagariya: Made requested code changes.

parth-lunagariya:
@filiphr , I did requested code changes and also added test cases having only cc and only bcc.

modules/flowable-process-validation/src/main/java/org/flowable/validation/validator/impl/ExternalInvocationTaskValidator.java
```diff
@@ -32,7 +32,7 @@ protected void validateFieldDeclarationsForEmail(org.flowable.bpmn.model.Process
         boolean textOrHtmlDefined = false;
 
         for (FieldExtension fieldExtension : fieldExtensions) {
-            if ("to".equals(fieldExtension.getFieldName())) {
+            if ("to".equals(fieldExtension.getFieldName()) || "cc".equals(fieldExtension.getFieldName()) || "bcc".equals(fieldExtension.getFieldName())) {
```
dbmalkovsky, line 35: Maybe this is just my coding preferences but it seems that `fieldExtension.getFieldName()` is executed 7 times in this loop; wouldn't something like this be cleaner?
````
        for (FieldExtension fieldExtension : fieldExtensions) {
            switch (fieldExtension.getFieldName()) {
                case "to":
                case "cc":
                case "bcc": {
                    toDefined = true;
                    break;
                }
                case "html":
                case "htmlVar":
                case "text":
                case "textVar": {
                    textOrHtmlDefined = true;
                }
            }
        }
````

parthlunagariya: @dbmalkovsky  I did above changes.

parthlunagariya:
I did requested code changes, @filiphr  and @dbmalkovsky , please review.

parthlunagariya:
After recent check in, checks have failed due to some flowable dependency but i have not made any such changes. @filiphr , can you please look at the changes? let me know if anything is required from my side. I have made all changes requested.

filiphr:
thanks for the changes @parth-lunagariya. We will finish the review next week.

filiphr:
Thanks for this @parth-lunagariya. I have manually merged it in https://github.com/flowable/flowable-engine/commit/da3dcc219a0f7ed10fe7dcbfab55b67125ccc3c6. I took only the tests, since we already added this functionality in https://github.com/flowable/flowable-engine/pull/2938

PR #2760, opened 2021-01-05
ProcessInstanceCollectionResource: Honor 'returnVariables' parameter …

reva:
…instead of always returning them (default according to docs)

According to the documentation, a parameter `returnVariables` can be set to return variables when starting a process. The default is supposed to be `false`. At the moment the variables are always returned. There was a test which was "fixed" (?) with d9ebf1e5dcafcfae2e5c1f11c2ec6caf18159930.

As this behaviour is present for quite a while, maybe we should just set the default to true and change the documentation as well?

### DIFF
--- a/modules/flowable-rest/src/main/java/org/flowable/rest/service/api/RestResponseFactory.java
+++ b/modules/flowable-rest/src/main/java/org/flowable/rest/service/api/RestResponseFactory.java
@@ -571,41 +571,23 @@ public class RestResponseFactory {
     }
 
     public ProcessInstanceResponse createProcessInstanceResponse(ProcessInstance processInstance, RestUrlBuilder urlBuilder) {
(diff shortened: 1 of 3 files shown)


filiphr:
The change that led to the variables being exposed is https://github.com/flowable/flowable-engine/commit/376cbfe23fc7dd8c864d6f25549bd5d07aae9983#diff-73b6c409052c3dc84aef67752c4b5e23d7484d829b56e284c1f526442da1c254R1087-R1093

tijsrademakers:
Closing the PR, now that we have changed the docs to state that the returnVariables is by default true. Sorry for the confusion. (https://github.com/flowable/flowable-engine/commit/22e852028b623d30981b0529cff8b93a8f83f75b)

reva:
@tijsrademakers Okay, but the behavior is still broken, as 'returnVariables=false' still returns the variables. Should I open another PR for that?

tijsrademakers:
@reva that would be good indeed

PR #2777, opened 2021-01-20
Remove unused import in recent DMN commit

dbmalkovsky:


yvoswillens:
Hi @dbmalkovsky 

I had a commit which wasn't pushed and had, amongst some other minor changes, this change in it.
I just pushed that.
Thanks for pointing it out though.

PR #2787, opened 2021-02-03
support for hsql in bulk delete operations

ikaakkola:
Fixes various bulk delete operations to work with 'hsql' (http://www.hsqldb.org) database

#### Check List:
* Unit tests:  NA
* Documentation:  NA


tijsrademakers:
Thanks for the PR

PR #2791, opened 2021-02-04
Update AssertJ to version 3.19.0.

PascalSchumacher:


modules/flowable-engine/src/test/java/org/flowable/examples/test/OneTaskProcessTest.java
```diff
@@ -98,7 +98,7 @@ public void testProcessModelFailure() {
         // deploy different process - test should fail
         assertThatThrownBy(() -> testProcessModelByAnotherProcess(createTestProcessBpmnModel("twoTasksProcess")))
                 .isInstanceOf(AssertionError.class)
-                .hasMessageContaining("\nExpecting:\n <1L>\nto be equal to:\n <0L>");
+                .hasMessageContainingAll("expected: 0L", "but was : 1L");
```
PascalSchumacher, line 101: This change also make the test pass on windows (where the assertion message contains `\r\n` instead of `\n`).

filiphr:
Thanks @PascalSchumacher

Issue #2744, opened 2020-12-15
Event listeners are not dispatched from the CMMN Task Service


DavidPLamas:
**Is your feature request related to a problem? Please describe.**
We're currently using event listeners configured on the engine to apply permission validations when any task is assigned to a user. We use this to check whether a user belongs to a group any time a task is assigned to someone.
We've recently started using CMMN and noticed that in the CMMN Engine you can still define "typedEventListeners" and "eventListeners" but those aren't being used when tasks are assigned. Looks like some code is missing on the CMMN engine :(

**Describe the solution you'd like**
"eventListeners" and "typedEventListeners" should be used when a task is completed in the CMMN engine. 
This means changing the method org.flowable.cmmn.engine.impl.task.TaskHelper#changeTaskAssignee and add code to dispatch the event, the same way it's done in org.flowable.engine.impl.util.TaskHelper#changeTaskAssignee

**Additional context**
We're currently using Flowable 6.5.0 but I have also checked and this feature is still not available in the 6.6.0 version. 

I don't mind making a pull request with this feature. Is this something you feel worth contributing?

DavidPLamas:
Hi, I've just added a new pull request to implement this feature.

PR #2794, opened 2021-02-05
Dispatch cmmn engine events when task is assigned #2744

DavidPLamas:
#### Check List:
* Unit tests: YES
* Documentation: NO


modules/flowable-cmmn-api/src/main/java/org/flowable/cmmn/api/event/FlowableTaskAssignedEvent.java
```diff
@@ -0,0 +1,13 @@
+package org.flowable.cmmn.api.event;
```
dbmalkovsky, line 1: Please add the standard license header to all new files; thanks.

filiphr (changes requested):
Thanks for the PR @DavidPLamas. 

It looks great, I have only one small remark. I don't think that we need a new `FlowableTaskAssignedEvent` class since there are a lot of other places where we throw the task assigned event with a `FlowableEngineEntityEvent` instead. 

Can you please adapt your changes to match the way it is done currently?

DavidPLamas:
Hi @filiphr,

I've removed the FlowableTaskAssignedEvent and started using FlowableEntityEvent instead. I was gonna copy the code from ```org.flowable.engine.delegate.event.impl.FlowableEventBuilder#createEntityEvent``` but I thought it was too much.
Let me know if there is anything else that needs changing.

DavidPLamas:
Hi,
It's been a while, I just noticed the build has failed. From looking at the error log, it doesn't seem to be because of my changes.
Is there any way to retrigger the build to check if it was just a temporary problem?

filiphr:
Thanks for the ping @DavidPLamas. No need to retrigger. The Windows build is a bit unstable. The other builds look good

filiphr:
I manually merged this in https://github.com/flowable/flowable-engine/commit/86f818f37eca2f2c06ea2e66f1340ba51e0308db

PR #2795, opened 2021-02-09
Multi instance behavior complete condition on child execution

tinabandaloemruli:
#### Check List:
* Unit tests:  NO
* Documentation: YES

added new abstract method getCompletionConditionExecution in MultiInstanceActivityBehavior to make it possible to easy override it and enable checking completionConditionSatisified on child execution


modules/flowable-engine/src/test/java/org/flowable/engine/test/cfg/multitenant/SharedExecutorServiceAsyncExecutorTest.java
```diff
@@ -0,0 +1,71 @@
+package org.flowable.engine.test.cfg.multitenant;
```
dbmalkovsky, line 1: Please add the common file license header to any new files; thanks.

modules/flowable-engine/src/test/java/org/flowable/engine/test/cfg/multitenant/SharedExecutorServiceAsyncExecutorTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+import org.flowable.common.engine.impl.cfg.multitenant.TenantInfoHolder;
```
dbmalkovsky, line 3: The project has a established a standard order for imports.   You can find default settings for Eclipse and IntelliJ in the directory `.ide-settings` in the root of the project.  

For example, import ordering is 
* static imports
* java
* javax
* org
* com
* everything else

Please order the imports accordingly; thanks.

modules/flowable-engine/src/test/java/org/flowable/engine/test/cfg/multitenant/SharedExecutorServiceAsyncExecutorTest.java
tinabandaloemruli: thanks for the hint I added it

modules/flowable-engine/src/test/java/org/flowable/engine/test/cfg/multitenant/SharedExecutorServiceAsyncExecutorTest.java
tinabandaloemruli: imported code style format from flowable guess this is enough

dbmalkovsky: Once you load the code style file to rearrange the imports you need to use `Code`->`Optimize Imports` from the menu or on Linux, `Ctrl+Alt+O`.

filiphr (changes requested):
Thanks for the PR @tinabandaloemruli.

I've left a comment about the current implementation.

Additionally, why are there changes to the  `SharedExecutorServiceAsyncExecutor.java` file? I don't see how this is related to the goal of the PR.

PR #2798, opened 2021-02-12
Support method overloading in JUEL and cache ELResolvers

filiphr:
This PR adds support for method overloading in JUEL for Flowable
It also adds caching for ELResolvers by refactoring the DefaultExpressionManager and making the VariableContainerELResolver stateless.

This is done in 5 commits instead of one big commit in order to display the simplification steps towards the method overload support and the default expression manager refactoring.


modules/flowable-dmn-spring-configurator/src/main/java/org/flowable/dmn/spring/configurator/SpringDmnEngineConfigurator.java
```diff
@@ -19,7 +19,6 @@
 import org.flowable.dmn.engine.DmnEngine;
```
dbmalkovsky, line 19: Just above this at line 17 the import, `org.flowable.common.engine.impl.el.ExpressionManager;` is no longer used.

modules/flowable-dmn-spring-configurator/src/main/java/org/flowable/dmn/spring/configurator/SpringDmnEngineConfigurator.java
```diff
@@ -39,20 +38,9 @@ public void configure(AbstractEngineConfiguration engineConfiguration) {
         
         SpringEngineConfiguration springEngineConfiguration = (SpringEngineConfiguration) engineConfiguration;
         ((SpringDmnEngineConfiguration) dmnEngineConfiguration).setTransactionManager(springEngineConfiguration.getTransactionManager());
-        ExpressionManager configuredExpressionManager = dmnEngineConfiguration.getExpressionManager();
```
dbmalkovsky, line 42: Removing this line is the reason that the import noted above is no longer used.

modules/flowable-dmn-spring-configurator/src/main/java/org/flowable/dmn/spring/configurator/SpringDmnEngineConfigurator.java
filiphr: Thanks for the check David

PR #2812, opened 2021-02-13
Update Jackson to version 2.12.1.

PascalSchumacher:
Spring Boot 2.4 supports Jackson 2.12, so this update should be possible.

### DIFF
--- a/distro/src/notice.txt
+++ b/distro/src/notice.txt
@@ -96,10 +96,10 @@ Quartz
 This software includes unchanged copies of the following libraries:
 
 aopalliance                     aopalliance                 1.0             Public Domain
-com.fasterxml.jackson.core      jackson-annotations         2.11.4          The Apache Software License, Version 2.0
-com.fasterxml.jackson.core      jackson-core                2.11.4          The Apache Software License, Version 2.0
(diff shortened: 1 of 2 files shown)


dbmalkovsky:
Some of the failures are:
````
java.lang.IllegalArgumentException: Java 8 date/time type `java.time.Instant` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling
````
which indicates adding this `dependency` may help:
````
<dependency>
    <groupId>com.fasterxml.jackson.datatype</groupId>
    <artifactId>jackson-datatype-jsr310</artifactId>
    <version>${jackson.version}</version>
</dependency>
````

dbmalkovsky:
@PascalSchumacher when I pulled your PR and added the jar as noted above I got a clean build and test.

PascalSchumacher:
@dbmalkovsky Thanks, but the question is: Why is jackson-datatype-jsr310 required when using Jackson 2.12?

On further thinking I believe it is best to wait with the update until Spring Boot uses Jackson 2.12 too.

PascalSchumacher:
I guess this is the reason for the test failures: https://github.com/FasterXML/jackson-databind/issues/2683

PR #2819, opened 2021-02-17
Chore: remove unused imports

dbmalkovsky:
This is embarrassing as I'm the reason for the unused imports because of push removing duplicated catch clauses.


dbmalkovsky:
@filiphr Added another unused import to this PR from the code you just merged for David Lamas

PR #2821, opened 2021-02-18
Upgrade Spring Boot 2.4.3

dbmalkovsky:
See release notes: https://github.com/spring-projects/spring-boot/releases/tag/v2.4.3

I have included the update to Spring Security in this PR and attempted to find and upgrade all dependencies noted in the Release Notes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,11 +14,11 @@
 		<distributionManagementSnapshotsUrl>https://oss.sonatype.org/content/repositories/snapshots/</distributionManagementSnapshotsUrl>
 		<jdk.version>1.8</jdk.version>
 		<!-- When updating one spring version, make sure that all of them are updated to their latest compatible versions -->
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -222,7 +222,7 @@
 			<dependency>
 				<groupId>mysql</groupId>
 				<artifactId>mysql-connector-java</artifactId>
-				<version>8.0.19</version>
+				<version>8.0.23</version>
```
filiphr, line 225: Let's don't upgrade the MySQL version. There is a breaking change in the MySQL JDBC driver which no longer works with Liquibase,

filiphr (changes requested):
Thanks for the update @dbmalkovsky. I've left a comment which needs to be addressed before merging it.

pom.xml
dbmalkovsky: @filiphr Good catch; I've reverted the change.

PR #2824, opened 2021-02-19
Basic CDI support for CMMN engine

andyverberne:
Implementation of:
- CmmnEngineLookup.java (like ProcesEngineLookup.java)
- FlowableCmmnServices.java (like FlowableServices.java)

#### Check List:
* Unit tests: NO
* Documentation: NO


modules/flowable-cmmn-cdi/src/main/java/org/flowable/cdi/impl/FlowableCmmnExtension.java
```diff
@@ -0,0 +1,111 @@
...
+                LOGGER.debug("CmmnEngineLookup service {} returned cmmn engine.", cmmnEngineLookup.getClass());
+                break;
+            } else {
+                LOGGER.debug("CmmnEngineLookup service {} returned 'null' value.", cmmnEngineLookup.getClass());
+            }
+        }
+
+        if (cmmnEngineLookup == null) {
+            throw new FlowableException(
+                    "Could not find an implementation of the org.flowable.cdi.spi.CmmnEngineLookup service " + "returning a non-null cmmnEngine. Giving up.");
```
dbmalkovsky, line 90: Is there some reason that the message is formed using concatenation of strings?

andyverberne: This class is a copy of org.flowable.cdi.impl.FlowableExtension.
I deliberately modified the code as less as possible so that these two classes would look similar. 
This helps when these classes need to be changed in the future.
The original class already had this concatenation of strings.

dbmalkovsky: My inclination would be to change both not to use concatenation; but it is your call and the owners' of the project.

andyverberne: Oh Sorry, Off course you are right. Id did not look close enough :-)
These two strings should have been concatenated off course

andyverberne: The original code uses conactenation of strings and variables while I always use the String.format() functionality.
I was inclined to use that in the new classes I introduced but then I decided not so that the classes would look similar.
So when you talk about concatenation I presumed it was a concatenation of strings and variables.

tijsrademakers:
Thanks for the PR!

PR #2829, opened 2021-02-22
Update Jackson to version 2.12.3.

PascalSchumacher:
New version of https://github.com/flowable/flowable-engine/pull/2812 as I can not reopen the pull request because of force pushing.

Not sure if this should be merged or if you like to wait until Spring Boot uses Jackson 2.12.

### DIFF
--- a/distro/src/notice.txt
+++ b/distro/src/notice.txt
@@ -96,10 +96,10 @@ Quartz
 This software includes unchanged copies of the following libraries:
 
 aopalliance                     aopalliance                 1.0             Public Domain
-com.fasterxml.jackson.core      jackson-annotations         2.11.4          The Apache Software License, Version 2.0
-com.fasterxml.jackson.core      jackson-core                2.11.4          The Apache Software License, Version 2.0
-com.fasterxml.jackson.core      jackson-databind            2.11.4          The Apache Software License, Version 2.0
-com.fasterxml.jackson.datatype  jackson-datatype-joda       2.11.4          The Apache Software License, Version 2.0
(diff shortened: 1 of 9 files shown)


filiphr:
Thanks for figuring this out @PascalSchumacher. We'll have a look at it and try to resolve it. 

It is failing now because the `JavaTimeModule` was not registered, I think that it isn't done automatically. Looking at the failure it happens in testing, I think we can do better there, we aren't using java time in the production code.

PascalSchumacher:
Hi Filip,

not sure what you mean. Should these tests be removed?

filiphr:
>not sure what you mean. Should these tests be deleted?

No we don't need to delete them. We need to analyse why this error is happening.

PascalSchumacher:
o.k., build is fixed.

PascalSchumacher:
Hi Filip,

any feedback on this pull request?

filiphr:
Thanks for the ping @PascalSchumacher. I am not sure that the custom Jackson object provider is the way to go for JsonUnit. We should review the tests.

In addition to that Spring Boot 2.4.5 still uses Jackson 2.11. I think that it is better to wait and upgrade the Jackson Dependency in the same time as the Spring Boot dependency. 

Let's leave this open until we do that.

filiphr:
Thanks a lot for this @PascalSchumacher. I have manually merged it in https://github.com/flowable/flowable-engine/commit/074a82d6b7c3e88a552d5c55157373f193450a24.

I've fixed the tests that were causing the problem with Jackson and removed the `JavaTimeObjectMapperProvider`(s). Your PR also inspired me and I did https://github.com/flowable/flowable-engine/commit/52255d018795a62c94efae4efda17f18b4163aef

PR #2830, opened 2021-02-23
Import a process model as a new version

t-d-marsen:
This PR is going to improve three topics (refer to https://forum.flowable.org/t/import-a-process-model-as-a-new-version/7522) regarding the "Import Process" behavior of the "Modeler App":

1. AS IS: It is currently possible to import a new workflow with key "X" even if there is already a workflow existing with that key "X" resulting in a duplicate. Subsequent save operations will then fail. TO BE: Importing a workflow with existing key will no longer be possible.

2. AS IS: Error Messages on Import Process are showing as part of the footer, right aligned without proper layout. TO BE: Align error message handling with the "Create Process" dialog.

3. AS IS: It is currently not possible to import a new version into an existing process. TO BE: Provide a new Option "Create new version if the key of that workflow already exists." as part of the import dialog. If importing a workflow with an already existing key and the option is checked then continue to import the workflow as a new version into the existing one.

Out of Scope
- Change of documentation
- Change of CMMN, DMN and Form

### DIFF
--- a/modules/flowable-ui/flowable-ui-modeler-frontend/src/main/resources/static/modeler/views/popup/process-import.html
+++ b/modules/flowable-ui/flowable-ui-modeler-frontend/src/main/resources/static/modeler/views/popup/process-import.html
@@ -16,6 +16,13 @@
                     <i class="glyphicon glyphicon-info-sign"></i>
                     {{'PROCESS.POPUP.IMPORT-DESCRIPTION' | translate}}
                 </p>
+
+                <div ng-if="model.error && model.errorMessage && model.errorMessage.length > 0" class="alert error" style="font-size: 14px; margin-top:20px">
+                    <div class="popup-error" style="font-size: 14px">
+                        <span class="glyphicon glyphicon-remove-circle"></span>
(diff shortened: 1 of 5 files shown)


modules/flowable-ui/flowable-ui-modeler-frontend/src/main/resources/static/modeler/views/popup/process-import.html
```diff
@@ -65,14 +72,18 @@ <h2>{{'PROCESS.POPUP.IMPORT-TITLE' | translate}}</h2>
                 </div>
 
                 <!-- <![endif]-->
+
+                <div class="form-check">
+                    <input class="form-check-input" type="checkbox" ng-model="model.importNewVersionIndicator" id="importNewVersionIndicator">
+                    <label class="form-check-label" for="importNewVersionIndicator">
+                        Import into existing process model as new version if keys are matching.
```
dbmalkovsky, line 79: Shouldn't this be something like:
````
 {{'PROCESS.POPUP.IMPORT-NEW-VERSION' | translate}}
````
and then the string added to `modules/flowable-ui/flowable-ui-modeler-frontend/src/main/resources/static/modeler/i18n/en.json` some where around line 2153 so it can be localized?

tijsrademakers:

PR #2831, opened 2021-02-24
Fix wait state of asynchronous multi-instance activities

AlexanderWirthSAP:
Without this change a flow node with multi-instance characteristics marked as asynchronous is still initially executed synchronously by the ContinueProcessOperation. If a failure happens during the initialization of the multi-instance, e.g. the referenced collection does not exist, then a rollback occurs. However, this rollback disregards that the asynchronous flow node should be a wait state and instead rolls back to the beginning of any previous asynchronous activity.

Example: Start Event --> Async Script Task --> Async Multi-Instance Call Activity --> End Event
If the multi-instance call activity fails because the collection variable cannot be resolved, then the execution is rolled back all the way to before the script task, whereas the expectation would be that the execution is rolled back only to the beginning of the call activity, because it is marked as asynchronous.

This change fixes this by removing the special case for multi-instance activities in ContinueProcessOperation, so that any asynchronous flow node again first commits the transaction and creates a new async-continuation job and is then actually being executed when it is later picked up by the job executor.

#### Check List:
* Unit tests: YES (MultiInstanceTest::testAsync**Rollback)
* Documentation: NA


tijsrademakers:
Thanks for the PR, we are closing this PR because it's outdated, but please feel free to reopen a new PR and we can discuss it again.

PR #2832, opened 2021-02-24
Multi instance behavior complete condition on child execution

tinabandaloemruli:
leave method changed to enable checking completionConditionSatisified on child execution


tinabandaloemruli:
@filiphr as discussed today

PR #2833, opened 2021-02-25
JsonNodeELResolver_extension_for_BigInteger

tinabandaloemruli:


tijsrademakers:
Hi @tinabandaloemruli, for BigDecimals the value asDouble() is used, so for BigInteger it would be more consistent to use asLong.

tinabandaloemruli:
@tijsrademakers yes I saw it for double and float etc. but anyway there are different limits for those datatypes or don't we want to support that. if so it would make sense to reject it or not?

tijsrademakers:
Closing PR based on earlier remark.

PR #2850, opened 2021-03-13
Fix infinite recursion

dbmalkovsky:
As the code currently is written a call to `getJobServiceConfiguration()` results in infinite recursion.  I think the intent was to call `getServiceConfiguration` which is part of `AbstractServiceEngineEntityManager`.

FWIW, `JobInfoEntityManagerImpl` extends `AbstractJobServiceEngineEntityManager` which extends `AbstractServiceEngineEntityManager`.


jbarrez:
Thanks @dbmalkovsky. The method doesn't seem to be used anywhere else, but good that you caught it!

PR #2859, opened 2021-03-24
Handle candiate groups and users user task activity

tinabandaloemruli:
added validation for non supported types of candidates user and group
added variables for a limitation of candidateGroups and candidateUsers and per default no limit
added normalization of candidateGroups and candidateUsers as duplicate values are possible and also empty values see Tests

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/UserTaskActivityBehavior.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/UserTaskActivityBehavior.java
@@ -314,7 +314,7 @@ public class UserTaskActivityBehavior extends TaskActivityBehavior implements Ac
                     Expression idExpression = expressionManager.createExpression(userIdentityLink);
                     Object value = idExpression.getValue(execution);
 
-                    Collection<String> userIds = extractCandidates(value);
+                    Collection<String> userIds = extractCandidates(value, idExpression.getExpressionText());
                     for (String userId : userIds) {
                         IdentityLinkEntity identityLinkEntity = processEngineConfiguration.getIdentityLinkServiceConfiguration()
                                 .getIdentityLinkService().createTaskIdentityLink(task.getId(), userId, null, customUserIdentityLinkType);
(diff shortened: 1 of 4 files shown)


modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/UserTaskActivityBehavior.java
```diff
@@ -18,6 +18,8 @@
 import java.util.Collections;
 import java.util.Date;
 import java.util.List;
+import static java.util.stream.Collectors.toList;
```
dbmalkovsky, line 21: Thanks for the PR.

This is a *very* small issue given the work done here but the project standards call for all `static` imports to be first in the file.  Obviously this doesn't change the execution of the code but is just the chosen style.

Thanks again.

tijsrademakers:
Hi @tinabandaloemruli, adding a limit on the number of candidate users and groups means that the process execution will basically fail when the number of candidate users and groups is exceeded for a specific user task. I'm not seeing the general benefit of adding this to Flowable, and certainly not on an engine configuration level, which means it's for every user task that will be created in every process definition.

tinabandaloemruli:
> Hi @tinabandaloemruli, adding a limit on the number of candidate users and groups means that the process execution will basically fail when the number of candidate users and groups is exceeded for a specific user task. I'm not seeing the general benefit of adding this to Flowable, and certainly not on an engine configuration level, which means it's for every user task that will be created in every process definition.

@tijsrademakers Yes but what would be the reason to have an unlimited possibility to add them? Couldn't cause an overload of database at some point? And what about users having a huge amount of processes and data?

tijsrademakers:
We don't want to impose limits for this. If there's a reason to limit this then it can be done with for example task listeners etc.

tinabandaloemruli:
> We don't want to impose limits for this. If there's a reason to limit this then it can be done with for example task listeners etc.

@tijsrademakers I have removed it

modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/UserTaskActivityBehavior.java
tinabandaloemruli: @dbmalkovsky thanks for the review I have done the change.

PR #2864, opened 2021-04-01
CMMN case reactivation, part 1

mkiener:
Part 1 consists of copying the case instance, plan items and variables from history back to runtime, reactivate the event listener and trigger it to reactivate the case.

#### Check List:
* Unit tests: YES
* Documentation: NO

### DIFF
--- a/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/agenda/operation/ReactivateCaseInstanceOperation.java
+++ b/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/agenda/operation/ReactivateCaseInstanceOperation.java
@@ -1,0 +1,113 @@
+/* Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.flowable.cmmn.engine.impl.agenda.operation;
+
+import java.util.List;
+
+import org.flowable.cmmn.api.runtime.PlanItemInstanceState;
+import org.flowable.cmmn.engine.impl.persistence.entity.CaseInstanceEntity;
+import org.flowable.cmmn.engine.impl.persistence.entity.PlanItemInstanceEntity;
+import org.flowable.cmmn.engine.impl.repository.CaseDefinitionUtil;
+import org.flowable.cmmn.engine.impl.util.CommandContextUtil;
+import org.flowable.cmmn.model.PlanItem;
+import org.flowable.cmmn.model.ReactivateEventListener;
+import org.flowable.common.engine.api.FlowableIllegalArgumentException;
+import org.flowable.common.engine.impl.interceptor.CommandContext;
+
+/**
+ * This operation reactivates a case model by first setting its reactivation event listener to active and triggering it, then in a second phase step through
+ * all plan items having a reactivation sentry and reactivate them before the model is completely re-evaluated again for reactivation of the case.
+ *
+ * @author Micha Kiener
+ */
+public class ReactivateCaseInstanceOperation extends AbstractCaseInstanceOperation {
+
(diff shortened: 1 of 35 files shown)


modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/runtime/CaseInstanceHelperImpl.java
```diff
@@ -421,6 +472,178 @@ protected CaseInstanceEntity createCaseInstanceEntityFromDefinition(CommandConte
...
+        if (caseInstanceEntity instanceof CaseInstanceEntityImpl) {
+            ((CaseInstanceEntityImpl) caseInstanceEntity)
+                .setVariableInstancesToCache(createCaseVariablesFromHistoricCaseInstance(historicCaseInstance, caseInstanceEntity));
+        } else {
+            throw new FlowableIllegalArgumentException("Cannot set variables on reactivated case instance as it is not from expected implementation class.");
+        }
+
+        // create runtime plan items from the history and set them as the new child plan item list
+        caseInstanceEntity.setChildPlanItemInstances(createCasePlanItemsFromHistoricCaseInstance(historicCaseInstance, caseInstanceEntity));
+
```
mkiener, line 524: trigger history manager reactivate method here -> and move the logic below there

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/runtime/CaseInstanceHelperImpl.java
```diff
@@ -421,6 +472,178 @@ protected CaseInstanceEntity createCaseInstanceEntityFromDefinition(CommandConte
...
+     * Creates a copy of the provided historic plan item back to the runtime with exactly the same data. Nothing is changed from a data perspective, not even
+     * the state or any timestamps, this might be changed later on according the reactivation case model.
+     *
+     * @param caseInstanceEntity the copied case instance from the runtime where the plan item is attached to
+     * @param historicPlanItemInstance the historic plan item to be copied back to the runtime
+     * @return the newly created runtime plan item instance for further processing
+     */
+    protected PlanItemInstanceEntity createPlanItemEntityFromHistoricPlanItemInstance(CaseInstanceEntity caseInstanceEntity, HistoricPlanItemInstance historicPlanItemInstance) {
+        PlanItemInstanceEntityManager planItemInstanceEntityManager = cmmnEngineConfiguration.getPlanItemInstanceEntityManager();
+        PlanItemInstanceEntity planItemInstanceEntity = planItemInstanceEntityManager.create();
```
mkiener, line 573: use a new create method as well taking in a historic plan item instance

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/runtime/CaseInstanceHelperImpl.java
```diff
@@ -421,6 +472,178 @@ protected CaseInstanceEntity createCaseInstanceEntityFromDefinition(CommandConte
...
+        if (variables != null) {
+            Map<String, VariableInstanceEntity> newVars = new HashMap<>(variables.size());
+            for (HistoricVariableInstance variable : variables) {
+                VariableInstanceEntity newVariable = variableService.createVariableInstance(
+                    variable.getVariableName(), variableTypes.getVariableType(variable.getVariableTypeName()), variable.getValue());
+
+                newVariable.setId(variable.getId());
+                newVariable.setScopeId(newCaseInstance.getId());
+                newVariable.setScopeType(variable.getScopeType());
+                newVariable.setSubScopeId(variable.getSubScopeId());
```
mkiener, line 637: don't copy, if there is a sub-scope set as it would mean to copy local vars like from stages, etc which we don't need

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/agenda/operation/ReactivateCaseInstanceOperation.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * Reactivates the given plan item into the provided new state. It will also reset all necessary data like ending times to prepare the plan item
+     * being active again once the listener is triggered.
+     *
+     * @param planItem the plan item to be reactivated
+     * @param newState the new state to set
+     */
+    protected void reactivatePlanItem(PlanItemInstanceEntity planItem, String newState) {
```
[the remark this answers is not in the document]
mkiener: Will be a bigger change as it involves a new operation on the agenda and even new transition and listeners

PR #2884, opened 2021-04-09
injecting error code into execution on boundary / start error event

borispran:
Motive for this feature is to enable generic BPMN error handling without the need to know in workflow design-time which error code will be thrown. 

ErrorEventDefinition gets three new properties:
- String errorVariableName, name of the variable containing error code, default null
- Boolean errorVariableTransient, whether variable should be transient, default true
- Boolean errorVariableLocalScope, whether variable should be stored in local scope, default true

If errorVariableName is not set nothing gets injected, this way backward compatibility is maintained, else, error code is stored into named variable. Scope and transience depends on flags.

Unit testing and documentation is missing since these are just proposed changes.

### DIFF
--- a/modules/flowable-ui/flowable-ui-modeler-frontend/src/main/resources/static/modeler/i18n/en.json
+++ b/modules/flowable-ui/flowable-ui-modeler-frontend/src/main/resources/static/modeler/i18n/en.json
@@ -730,6 +730,20 @@
             "DESCRIPTION" : "Define the error name."
           }
         },
+        "ERRORVARIABLEPACKAGE" : {
+          "ERRORVARIABLENAME" : {
+            "TITLE" : "Error variable name",
+            "DESCRIPTION" : "Define the error variable name."
+          },          
+          "ERRORVARIABLETRANSIENT" : {
+            "TITLE" : "Store as transient variable",
+            "DESCRIPTION" : "Flag defines whether a variable will be stored transiently."
+          },
+          "ERRORVARIABLELOCALSCOPE" : {
+            "TITLE" : "Store to a local scope",
+            "DESCRIPTION" : "Flag defines whether variable will be stored to a local scope."
+          }          
+        },
         "ESCALATIONREFPACKAGE" : {
           "ESCALATIONREF" : {
             "TITLE" : "Escalation reference",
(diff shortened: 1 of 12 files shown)


modules/flowable-ui/flowable-ui-modeler-frontend/src/main/resources/static/modeler/i18n/en.json
```diff
@@ -730,6 +730,20 @@
...
           }
         },
+        "ERRORVARIABLEPACKAGE" : {
+          "ERRORVARIABLENAME" : {
+            "TITLE" : "Error variable name",
+            "DESCRIPTION" : "Define the error variable name."
+          },          
+          "ERRORVARIABLETRANSIENT" : {
+            "TITLE" : "Store as transient variable",
+            "DESCRIPTION" : "Flag defines weather a variable will be stored transiently."
```
dbmalkovsky, line 740: Spelling mistake:    weather -> whether

modules/flowable-ui/flowable-ui-modeler-frontend/src/main/resources/static/modeler/i18n/en.json
```diff
@@ -730,6 +730,20 @@
...
+          },
+          "ERRORVARIABLELOCALSCOPE" : {
+            "TITLE" : "Store to a local scope",
+            "DESCRIPTION" : "Flag defines weather variable will be stored to a local scope."
```
dbmalkovsky, line 744: Spelling mistake: weather -> whether

dbmalkovsky:
Thanks for the PR; found two small spelling mistakes (the words are pronounced the same but have very different meanings).

modules/flowable-ui/flowable-ui-modeler-frontend/src/main/resources/static/modeler/i18n/en.json
borispran: Ups! Embarrassing ... will fix it at once :)

tijsrademakers:
Thanks for the PR!

arthware:
Forum discussion: https://forum.flowable.org/t/bpmnerror-context-in-errorboundaryevent/762/13

PR #2892, opened 2021-04-14
Rework how we do paging and including data from other tables

filiphr:
When checking data from other tables e.g. Variable values, process definition etc.
we should always do it in a sub select instead of doing it in a join.
This makes sure that we can correctly sort and order without using distinct.

In case we need to include data from other tables, we would do the querying and ordering in a sub select
and then join on the other tables.

Doing this makes the different query limits and querying in memory obsolete


gro-mar:
Hi Filip.
This change is really interesting.
Is there any performance impact?
Do we have any benchmarks?

Regards
Martin

PR #2898, opened 2021-04-16
Upgrade to Spring Boot 2.4.5

dbmalkovsky:
The Spring Boot Release Notes are [here](https://github.com/spring-projects/spring-boot/releases/tag/v2.4.5).

I made the dependency updates that I found in the release notes and ran "mvn clean test" with no errors. 

The only change I did *not* make is the update to HSQLDB (from 2.3.2 to 2.5.2) as it seemed like a big jump and I could not find any references to *org.hsqldb* in any file (other than pom.xml) in the project.    I suspect that it might be a candidate for removal from the project.

Note: I could not update to the version for spring-data-jpa as noted in the release notes because it caused compile errors in some tests.   Given that it was just in test files I left the version as it was.    There is a similar issue with the tests in the HTTP module and upgrading Jetty (not compile but runtime).

### DIFF
--- a/modules/flowable-app-engine-rest/pom.xml
+++ b/modules/flowable-app-engine-rest/pom.xml
@@ -17,7 +17,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <jetty.version>9.4.38.v20210224</jetty.version>
+        <jetty.version>${jetty.version}</jetty.version>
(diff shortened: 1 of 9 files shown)


filiphr:
Thanks for the PR @dbmalkovsky. However, by now Spring Boot 2.5.0 has been released already. I would say we should upgrade to that version directly. Are you interested in doing that, or you want us to do that upgrade?

PR #2906, opened 2021-04-21
Minor Fix to Modeler - Create Event Model Logic

maisalragheb:
Create Event Model should add the payload params whether the send event task is triggable or not


filiphr:
Hey @maisalragheb,

Thanks for the PR. If the task is not triggerable the transaction will not stop and there it will continue immediately after sending the event. Can you please explain why you need the IN variables for such a task?

maisalragheb:
Hello, if the event model is not having any payload definition, the task is sending an event without any payload. The in params are needed in order to to createEventPayloadInstances. So we need to send an event containing a payload values without having to wait for another event to come.

The attached screenshot has a kafka topic for the event sent without having the event params

At offset 16 the issue was solved

Screenshot 2021-04-21 at 17 12 40

maisalragheb:
@filiphr in short, we need it to send event payload that is already existing in the process variables.

filiphr:
You are right @maisalragheb. Thanks for the detailed explanation. Looking at the code the `eventModel` is indeed for the outgoing event and the `triggerModel` is for the incoming.

P.S. I had this already typed out 10 minutes after your first reply, but it seems I didn't press the send button 😄

PR #2912, opened 2021-04-22
flowable-cxf: Fix maven surefire plugin deprecation warning.

PascalSchumacher:


filiphr:
Thanks for the fix @PascalSchumacher

PR #2914, opened 2021-04-26
jmx ldap describe

qiudaoke:
#### Check List:
* Unit tests: YES / NO / NA


filiphr:
Thanks @qiudaoke. Recently we have moved away from asciidoc for the docs and are using docusaurus (https://github.com/flowable/flowable-engine/tree/master/docs/docusaurus). We only moved the english docs, perhaps we need to look into moving the Chinese one as well

PR #2916, opened 2021-04-27
Fix for Boundary Events on Asynchronous Multi-Instance Activities

vasil-tsimashchuk:
This contribution is a fix for the issue described in [the following forum discussion](https://forum.flowable.org/t/differences-in-boundary-event-behavior-on-sync-and-async-multi-instance-activities/7784): when a boundary event is attached to an asynchronous multi-instance activity, the corresponding boundary event is created `n+1` times upon activity execution, where `n` is the number of instances in the multi-instance activity.

In this contribution, an additional check is introduced for boundary event creation which assures that the events are only created for the multi-instance root execution. As a result, there will be only 1 boundary event created independently of the number of instances in the multi-instance activity.

#### Check List:
* Unit tests: YES
* Documentation: NA

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/agenda/ContinueProcessOperation.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/agenda/ContinueProcessOperation.java
@@ -206,7 +206,7 @@ public class ContinueProcessOperation extends AbstractOperation {
         if (activityBehavior != null) {
             executeActivityBehavior(activityBehavior, flowNode);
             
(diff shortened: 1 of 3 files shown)


jbarrez:
Manually merged it here: https://github.com/flowable/flowable-engine/commit/973706915e1065ee008922ef5e073bd67fd1dcd7

And added another test: https://github.com/flowable/flowable-engine/commit/b5adf403912c2fc27bb38d41b08ee0d3dfbd2742

Thanks!

vasil-tsimashchuk:
Thank you Joram!

PR #2917, opened 2021-04-27
Throw exception when a property is evaluated of a null base

vasil-tsimashchuk:
In this contribution, the issue is fixed described in [the following forum discussion](https://forum.flowable.org/t/astproperty-missing-exception-by-evaluation-of-a-property-of-a-null-base/7766/2): when a property is evaluated of a `null` base the corresponding method short circuits to `null`.

With the fix, stricter check is introduced which results in an exception when a property of a `null` base is requested.

#### Check List:
* Unit tests: NO
* Documentation: NA


vasil-tsimashchuk:
Closing for reevaluation

PR #2926, opened 2021-05-13
Improved conditional test execution

sando1998:
Problem:
Conditions within the test method may produce a false test outcome. Once a condition is not met, and the code block inside the condition is not executed, a false 'passed' outcome can be produced without the test ever being executed.

Solution:
Instead of using a conditional to control the test execution, we propose using JUnit Assumptions — a failed assumption results in a test being aborted. Assumptions are typically used whenever it does not make sense to continue executing a given test method.


filiphr:
@sando1998 there is a reason for us using the if conditions. This is not the only place where this is done. Some tests test a mixture of things and we have different profiles where the tests are run differently, i.e. without history, with different level of history etc.

Therefore, we are going to reject this PR. Hope you understand. Feel free to provide other contributions if you are interested.

PR #2940, opened 2021-06-02
flowable entrypoint script and permission changes

im-pratham:
#### Check List:
* Unit tests:  NA
* Documentation:  NA

The changes are to allow image to run on Openshift container platform

### DIFF
--- a/docker/base-image/Dockerfile
+++ b/docker/base-image/Dockerfile
@@ -1,11 +1,15 @@
 FROM adoptopenjdk/openjdk11:jre-11.0.8_10-alpine
 
-MAINTAINER Flowable <info@flowable.org>
-RUN apk add --no-cache ttf-dejavu su-exec && rm -rf /var/cache/apk/*
+LABEL maintainer="Flowable <info@flowable.org>"
+RUN apk add --no-cache ttf-dejavu && rm -rf /var/cache/apk/*
 
 RUN addgroup -S flowable && adduser -S flowable -G flowable
 
-RUN mkdir /data && chown flowable:flowable /data
+RUN mkdir /data && chown flowable:flowable /data && \
+    chgrp -R 0 /data && \
+    chmod -R g=u /data
 
 ADD wait-for-something.sh .
 RUN chmod +x wait-for-something.sh
+
+USER flowable:flowable
--- a/modules/flowable-ui/flowable-ui-app/pom.xml
+++ b/modules/flowable-ui/flowable-ui-app/pom.xml
@@ -173,7 +173,7 @@
                             <permissions>
                                 <permission>
                                     <file>/flowable-entrypoint.sh</file>
-                                    <mode>544</mode> <!-- Read/execute for owner, read for group/other -->
+                                    <mode>554</mode> <!-- Read/execute for owner/group, read for other -->
                                 </permission>
                             </permissions>
                         </extraDirectories>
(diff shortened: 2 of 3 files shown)


yvoswillens (commented):
Hi @im-pratham ,

Will the `554` mode fix the issue that Openshift will use a random user?
Will that user be part of the 'root group'?

Could you explain why switch with executing the command?
Will starting the `FlowableUiApplication` as the `flowable` user not suffice?

Regards,

Yvo

im-pratham:
@yvoswillens refer https://docs.openshift.com/container-platform/4.3/openshift_images/create-images.html section *Support arbitrary user ids*, there it's mentioned as "the container user is always a member of the root group." 
I added double check for ensure the condition of root user where both user id and group id is 0

im-pratham:
> 
> 
> Hi @im-pratham ,
> 
> Will the `554` mode fix the issue that Openshift will use a random user?
> Will that user be part of the 'root group'?
> 
> Could you explain why switch with executing the command?
> Will starting the `FlowableUiApplication` as the `flowable` user not suffice?
> 
> Regards,
> 
> Yvo

For starting as flowable user, su-exec is not permitted. I am getting error error as 
`su-exec: setgroups: Operation not permitted` even though I added 554 permission

im-pratham:
@yvoswillens I am using openshift developer sandbox to test the changes - https://developers.redhat.com/developer-sandbox/get-started

yvoswillens:
The image does not start without errors (outside open shift)

```
docker run --rm  -p8080:8080 impratham/flowable-ui
standard_init_linux.go:219: exec user process caused: no such file or directory
```

im-pratham:
That's mistake from my side, latest points to malfunctioning image. It had CRLF line ending. Refer this image `docker run --rm  -p8080:8080 impratham/flowable-ui:6.7.0-SNAPSHOT`.

Until then I will fix the tags

yvoswillens:
I'm happy with the first part of the change.
Less with the second.
I've to investigate a bit further how to support the use case of starting the container as root group member; but keep being able to start the main process as flowable user before merging this to master.

Will post progress here in this thread.

im-pratham:
@yvoswillens for 2nd part I am having some ideas, let me know your thoughts.
We can USER directives in docker file to cater the non-openshift environments. ( Openshift ignores the USER directives, hence giving 554 permission will suffice). So we can remove su-exec dependency altogether, but I need to investigate further how it will work with base-image and jib-maven plugin

yvoswillens:
I would prefer keep using su-exec. Let's see what is needed in order to execute that as root group member.

im-pratham:
@yvoswillens any specific reason to go for su-exec than user directive from dockerfile?

yvoswillens:
the way su-exec steps down from root has some benefits. But; rehashing it I see potential issues with OpenShift.
Could you prepare the PR with the USER alternative so that we can review it?

im-pratham:
> the way su-exec steps down from root has some benefits. But; rehashing it I see potential issues with OpenShift.
> Could you prepare the PR with the USER alternative so that we can review it?

Sure

im-pratham:
@yvoswillens few changes made after our discussion - 
1. Changed MAINTAINER to LABEL
2. Removed su-exec 
3. Granted permission to root group to app folder as per openshift guidelines
4. Granted 555 permission to flowable-entrypoint. (the files added via jib are owned by root:root. So to allow flowable user to execute entrypoint script I have to give exec permission to others)

Let me know your thoughts. I tested both images on openshift and local (Docker desktop for windows with WSL2 backend). To test you can use this image `impratham/flowable-ui:6.7.0-SNAPSHOT`

yvoswillens (approved):

yvoswillens:
First checks looks good! I'll run some more from master.
Thanks for the PR @im-pratham !

Regards,

Yvo

PR #2942, opened 2021-06-03
CMMN validation

4gsim:
#### Check List:
* Unit tests: YES
* Documentation: NO


modules/flowable-case-validation/src/test/java/org/flowable/cmmn/validation/validator/impl/HumanTaskValidatorTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+    private CmmnModel readXMLFile(String resource) {
+        return new CmmnXmlConverter().convertToCmmnModel(() -> HumanTaskValidatorTest.class.getResourceAsStream(resource), true, false);
+    }
+
+    @Test
+    public void testValidateNoErrors() {
+        CmmnModel cmmnModel = readXMLFile("humanTaskNoErrors.cmmn");
+        List<ValidationEntry> validationEntries = new HumanTaskValidator().validate(cmmnModel);
+        Assert.assertTrue(validationEntries.isEmpty());
```
dbmalkovsky, line 39: Please change and use Assertj (see https://assertj.github.io/doc/) instead of junit.Assert as this is what the project uses as its standard.  Thanks.

modules/flowable-case-validation/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>org.flowable</groupId>
+            <artifactId>flowable-cmmn-model</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+        </dependency>
```
filiphr, line 39: Testing frameworks should be in the test scope

filiphr (commented):
As a first step this looks really good @4gsim.

I've left one comment that we should address and also @dbmalkovsky has left a comment that needs to be addressed.

Apart from those two comments, let's also use the `CaseValidator` from the `CmmnEngineConfiguration` in the `CmmnParserImpl` similar to how it is done in the `BpmnParse` i.e. perform validation before we call `processCmmnElements`.

modules/flowable-case-validation/src/test/java/org/flowable/cmmn/validation/validator/impl/HumanTaskValidatorTest.java
4gsim: Done

modules/flowable-case-validation/pom.xml
4gsim: Done

filiphr:
Thanks for this PR @4gsim. I have manually integrated it in https://github.com/flowable/flowable-engine/commit/f87b3beb75374dca761b5a48c26f33f78e64a4f7 and did some polishing in https://github.com/flowable/flowable-engine/commit/17996f7949cffd2994bc2dff621033cf1c51e6c7

PR #2943, opened 2021-06-03
source expression resolves to Int in call activity

im-pratham:
#### Check List:
* Unit tests:  Yes 
* Documentation: NA

PR is to fix situation where Integer value of source expression is saved as String in Target Expression of call activity

### DIFF
--- a/modules/flowable-engine-common/src/test/java/org/flowable/common/engine/impl/el/JsonNodeELResolverTest.java
+++ b/modules/flowable-engine-common/src/test/java/org/flowable/common/engine/impl/el/JsonNodeELResolverTest.java
@@ -1,3 +1,15 @@
+/* Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.flowable.common.engine.impl.el;
 
 import static org.assertj.core.api.Assertions.assertThat;
--- a/modules/flowable-engine-common/src/test/java/org/flowable/common/engine/impl/el/JsonNodeELResolverTest.java
+++ b/modules/flowable-engine-common/src/test/java/org/flowable/common/engine/impl/el/JsonNodeELResolverTest.java
@@ -6,6 +6,9 @@ import org.flowable.common.engine.impl.javax.el.ELContext;
 import org.flowable.common.engine.impl.javax.el.ELResolver;
 import org.junit.jupiter.api.Test;
 
+import java.math.BigDecimal;
+import java.util.Date;
+
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 /**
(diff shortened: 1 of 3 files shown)


filiphr:
Thanks for the PR @im-pratham. Can you please add a test case for this change? It would be good to see where the problem is and why this fixes it

im-pratham:
@filiphr added test case to cover added functionality let me know if more is required,

modules/flowable-engine-common/src/test/java/org/flowable/common/engine/impl/el/JsonNodeELResolverTest.java
```diff
@@ -0,0 +1,85 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.flowable.common.engine.impl.javax.el.ELContext;
+import org.flowable.common.engine.impl.javax.el.ELResolver;
+import org.junit.jupiter.api.Test;
+
+import java.math.BigDecimal;
+import java.util.Date;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
```
dbmalkovsky, line 12: Per the project ide settings all static imports are listed first after the package name; thanks.

modules/flowable-engine-common/src/test/java/org/flowable/common/engine/impl/el/JsonNodeELResolverTest.java
```diff
@@ -0,0 +1,85 @@
...
+class JsonNodeELResolverTest {
+    private final ELResolver resolver = new JsonNodeELResolver();
+    private final ELContext context = new FlowableElContext(resolver, null);
+    private final ObjectMapper mapper = new ObjectMapper();
+
+    @Test
+    void integerValueShouldBeIntNode() {
+        ObjectNode node = mapper.createObjectNode();
+        resolver.setValue(context, node, "age", 11);
+        assertTrue(node.get("age").isInt());
```
dbmalkovsky, line 26: The project's standard testing framework is assertj; please see https://assertj.github.io/doc/

modules/flowable-engine-common/src/test/java/org/flowable/common/engine/impl/el/JsonNodeELResolverTest.java
```diff
@@ -0,0 +1,85 @@
+package org.flowable.common.engine.impl.el;
```
dbmalkovsky, line 1: Please add the standard license header; thanks.

modules/flowable-engine-common/src/test/java/org/flowable/common/engine/impl/el/JsonNodeELResolverTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper;
```
dbmalkovsky, line 3: The import order is defined in https://github.com/flowable/flowable-engine/tree/master/ide-settings:

- statics
- java
- javax
- org
- com
- liquibase
- everything else

modules/flowable-engine-common/src/test/java/org/flowable/common/engine/impl/el/JsonNodeELResolverTest.java
im-pratham: Done

modules/flowable-engine-common/src/test/java/org/flowable/common/engine/impl/el/JsonNodeELResolverTest.java
im-pratham: Done

modules/flowable-engine-common/src/test/java/org/flowable/common/engine/impl/el/JsonNodeELResolverTest.java
im-pratham: done

filiphr (changes requested):
Thanks for adding the tests @im-pratham. However, we do not want to test on such a low level. For us in Flowable it is important that the functionality works on the API level.

For you it seems to be a problem when you are writing in a nested Json Variable. I would suggest looking into [`JsonTest`](https://github.com/flowable/flowable-engine/blob/master/modules/flowable-engine/src/test/java/org/flowable/engine/test/json/JsonTest.java) and adding some of the missing test cases there.

Some tests that might inspire you are:

* `testSetNestedJsonNodeValue`
* `testSetDateInJsonNode`
* `testSetInstantInJsonNode`
* `testSetArrayNestedJsonNodeValue`
* `testSetNestedArrayNestedJsonNodeValue`

im-pratham:
> Thanks for adding the tests @im-pratham. However, we do not want to test on such a low level. For us in Flowable it is important that the functionality works on the API level.
> 
> For you it seems to be a problem when you are writing in a nested Json Variable. I would suggest looking into [`JsonTest`](https://github.com/flowable/flowable-engine/blob/master/modules/flowable-engine/src/test/java/org/flowable/engine/test/json/JsonTest.java) and adding some of the missing test cases there.
> 
> Some tests that might inspire you are:
> 
>     * `testSetNestedJsonNodeValue`
> 
>     * `testSetDateInJsonNode`
> 
>     * `testSetInstantInJsonNode`
> 
>     * `testSetArrayNestedJsonNodeValue`
> 
>     * `testSetNestedArrayNestedJsonNodeValue`

That is what I was looking for. I will commit on test case fot it

im-pratham:
@filiphr made changes. Have a look

filiphr:
Thanks for doing the changes @im-pratham. I have manually integrated this in https://github.com/flowable/flowable-engine/commit/83f721c794fb9f0629f01df004d1ed8f4bbd9239

PR #2945, opened 2021-06-08
ChangeTenantId Feature - Initial commit

llyorshch:
New method in the management interfaces of the engines to change the tenant id of the main instances (processes, cases, decisions) and the related entities (Timers, tasks, jobs, etc.). It does not change the tenant id of definitions, deployments and models. It's meant for active and historic instances only.

#### Check List:
* Unit tests: YES 
* Documentation: YES (Javadoc)


modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/cmd/ChangeTenantIdCmmnCompleteCmd.java
```diff
@@ -0,0 +1,117 @@
+package org.flowable.cmmn.engine.impl.cmd;
```
dbmalkovsky, line 1: Please add the standard license header to all new files; thanks.

llyorshch: Done. 👍🏻

modules/flowable-content-engine/src/main/java/org/flowable/content/engine/impl/cmd/ChangeTenantIdContentCompleteCmd.java
```diff
@@ -0,0 +1,37 @@
+package org.flowable.content.engine.impl.cmd;
+
+import static org.flowable.common.engine.api.tenant.ChangeTenantIdResult.Key.*;
```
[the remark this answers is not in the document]
llyorshch: Done. 👍🏻

modules/flowable-content-engine/src/main/java/org/flowable/content/engine/impl/cmd/ChangeTenantIdContentCompleteCmd.java
```diff
@@ -0,0 +1,50 @@
...
+import org.flowable.content.engine.impl.util.CommandContextUtil;
+import org.flowable.common.engine.api.tenant.ChangeTenantIdResult;
+import org.flowable.common.engine.impl.interceptor.Command;
+import org.flowable.common.engine.impl.interceptor.CommandContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ChangeTenantIdContentCompleteCmd implements Command<ChangeTenantIdResult> {
+
+        private final static Logger LOGGER = LoggerFactory.getLogger(ChangeTenantIdContentCompleteCmd.class);
```
dbmalkovsky, line 28: Modifier order; is `private static final`.  See for example, https://checkstyle.sourceforge.io/config_modifier.html#ModifierOrder

llyorshch: Done. 👍🏻

filiphr:
@llyorshch I haven't seen anything yet. One small comment, please do not merge from the master branch into the PR branch. It might make the review more difficult and also the merge. Ideally you can do a rebase and then force push.

e.g.

```
git rebase upstream/master
git push -f
```

modules/flowable-event-registry/src/main/resources/org/flowable/eventregistry/db/mapping/mappings.xml
```diff
@@ -19,6 +19,8 @@
     <mapper resource="org/flowable/common/db/mapping/entity/Property.xml" />
     <mapper resource="org/flowable/common/db/mapping/entity/TableData.xml" />
     <mapper resource="org/flowable/common/db/mapping/common.xml" />
+    <!-- <mapper resource="org/flowable/common/db/mapping/changeTenantId.xml"/> -->
```
llyorshch, line 22: Remove dead code

llyorshch (commented):
Last file review before marking the PR as ready

modules/flowable-engine-common-api/src/main/java/org/flowable/common/engine/api/tenant/ChangeTenantIdResult.java
```diff
@@ -0,0 +1,86 @@
+package org.flowable.common.engine.api.tenant;
```
dbmalkovsky, line 1: Looks like you missed one; the license header is missing here.

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/tenant/ChangeTenantIdBuilderCmmnImpl.java
```diff
@@ -0,0 +1,42 @@
+package org.flowable.cmmn.engine.impl.tenant;
```
dbmalkovsky, line 1: Missing license header.

modules/flowable-cmmn-api/src/main/java/org/flowable/cmmn/api/CmmnManagementService.java
```diff
@@ -318,4 +319,31 @@
...
+     *    <li>SuspendedJobs</li>
+     *    <li>DeadLetterJobs</li>
+     *    <li>HistoryJobs</li>
+     *    <li>ExternalWorkerJobs</li>
+     * </ul>
+     * <p>
+     * You must provide the source tenant id and the destination tenant id. All instances from the source tenant id in the CMMN scope
+     * will be changed to the target tenant id.
+     * 
+     */    ChangeTenantIdBuilder createChangeTenantIdBuilder(String fromTenantId, String toTenantId);
```
filiphr, line 347: Please format this correctly. The Java code should be below the comment line. This is meant for all management services

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/persistence/entity/data/impl/MybatisCaseInstanceDataManagerImpl.java
```diff
@@ -224,4 +225,26 @@ protected void setSafeInValueLists(CaseInstanceQueryImpl caseInstanceQuery) {
             }
         }
     }
+
+    @Override
+    public long countChangeTenantIdCmmnCaseInstances(String sourceTenantId, boolean onlyInstancesFromDefaultTenantDefinitions) {
+        String defaultTenantId = getCmmnEngineConfiguration().getDefaultTenantProvider().getDefaultTenant(sourceTenantId, ScopeTypes.CMMN, null);
```
filiphr, line 231: I am not sure that using the default tenant provider in such a low level is OK. The DataManagers are usually should not do any business specific logic. The default tenant provider in my opinion is something more high level.

This comment is valid for all the places the default tenant provider is used in the data managers.

modules/flowable-cmmn-engine/src/main/resources/org/flowable/cmmn/db/mapping/common.xml
```diff
@@ -0,0 +1,28 @@
...
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<!DOCTYPE mapper PUBLIC "-//mybatis.org//DTD Mapper 3.0//EN" "http://mybatis.org/dtd/mybatis-3-mapper.dtd">
+
+<mapper namespace="org.flowable.cmmn.db.mapping">
+    
+    <sql id="changeTenantIdCaseCriteriaSql">
```
filiphr, line 21: The pattern used here is not used in other places of the Flowable codebase. The pattern being, having a common sql in one file which is then reused from other MyBatis mappings. We do have one common.xml, which is for the native queries, but that one is simpler and doesn't use any column names or table names.

Let's move the logic to the respective MyBatis configurations for each entity.

modules/flowable-cmmn-engine/src/test/java/org/flowable/cmmn/test/tenant/ChangeTenantIdCaseInstanceTest.java
```diff
@@ -0,0 +1,264 @@
...
+                .addClasspathResource("org/flowable/cmmn/test/tenant/caseWithMilestone.cmmn").tenantId(TEST_TENANT_C)
+                .deploy().getId();
+        addDeploymentForAutoCleanup(deploymentIdWithTenantC);
+        this.deploymentIdWithoutTenant = cmmnRepositoryService.createDeployment()
+                .addClasspathResource("org/flowable/cmmn/test/tenant/caseWithMilestoneDup.cmmn").deploy().getId();
+        addDeploymentForAutoCleanup(deploymentIdWithoutTenant);
+    }
+
+    @Test
+    public void testDeployments() {
```
filiphr, line 61: Each test needs to be self explanatory. There are 2 tests here, where one test tests the setup for the other test. Let's move those 2 things together.

modules/flowable-cmmn-engine/src/test/java/org/flowable/cmmn/test/tenant/ChangeTenantIdCaseInstanceTest.java
```diff
@@ -0,0 +1,264 @@
...
+        String caseInstanceIdACompleted = startCase(TEST_TENANT_A, "caseWithMilestone", "caseInstanceACompleted", true);
+        String caseInstanceIdBCompleted = startCase(TEST_TENANT_B, "caseWithMilestone", "caseInstanceBCompleted", true);
+        String caseInstanceIdCCompleted = startCase(TEST_TENANT_C, "caseWithMilestone", "caseInstanceCCompleted", true);
+        
+        // Starting case instances that will be kept active
+        String caseInstanceIdAActive = startCase(TEST_TENANT_A, "caseWithMilestone", "caseInstanceAActive", false);
+        String caseInstanceIdBActive = startCase(TEST_TENANT_B, "caseWithMilestone", "caseInstanceBActive", false);
+        String caseInstanceIdCActive = startCase(TEST_TENANT_C, "caseWithMilestone", "caseInstanceCActive", false);
+        
+        Set<String> caseInstanceIdsTenantA = Sets.newSet(caseInstanceIdACompleted, caseInstanceIdAActive);
```
filiphr, line 84: Use `new HashSet<>(Arrays.asList(caseInstanceIdACompleted, caseInstanceIdAActive")` instead of using `Sets` from an internal Mockito API

modules/flowable-cmmn-engine/src/test/java/org/flowable/cmmn/test/tenant/ChangeTenantIdCaseInstanceTest.java
```diff
@@ -0,0 +1,264 @@
...
+
+        // All the instances should now be assigned to the tenant B
+        checkTenantIdForAllInstances(caseInstanceIdsTenantA, TEST_TENANT_B, "after the change to " + TEST_TENANT_B);
+        checkTenantIdForAllInstances(caseInstanceIdsTenantB, TEST_TENANT_B, "after the change to " + TEST_TENANT_B);
+
+        // The instances for tenant C remain untouched
+        checkTenantIdForAllInstances(caseInstanceIdsTenantC, TEST_TENANT_C, "after the change to " + TEST_TENANT_B);
+
+        // The simulation result must match the actual result
+        assertThat(simulationResult).isEqualTo(result).as("The simulation result must match the actual result.");
```
filiphr, line 114: Let's also verify the actual contents of the results. The reason for that is because we want to verify that the API is correct and contains the expected data. This applies to all the tests in this file and the other similar tests for the other engines.

modules/flowable-engine-common-api/src/main/java/org/flowable/common/engine/api/tenant/ChangeTenantIdBuilder.java
```diff
@@ -0,0 +1,47 @@
...
+ * 
+ */
+public interface ChangeTenantIdBuilder {
+
+    /**
+     * Only the instances that were created following a definition from the default tenant will be affected by the change,
+     * that is, if the instance was created with the "fallback to default tenant" feature.
+     *
+     */    
+    ChangeTenantIdBuilder onlyInstancesFromDefaultTenantDefinitions(boolean onlyInstancesFromDefaultTenantDefinitionsEnabled);
```
filiphr, line 31: This method does not follow the patterns we use in our API. For such a method it is enough doing `onlyInstancesFromDefaultTenantDefinitions()`

modules/flowable-engine-common-api/src/main/java/org/flowable/common/engine/api/tenant/ChangeTenantIdResult.java
```diff
@@ -0,0 +1,99 @@
...
+
+ package org.flowable.common.engine.api.tenant;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Container class to return the result of a change in a tenant id operation (produced by a {@link ChangeTenantIdBuilder})
+ */
+public class ChangeTenantIdResult {
```
filiphr, line 22: This class does not follow the patterns for most of our APIs in Flowable. 

1. Let's make this an interface that would be able to provide a method that would return the number of changed instances for a particular key, all types of instances that were changed. Exposing the map might be OKish as well.
1. We shouldn't use an enum for the key as it makes it difficult to add new elements if we need them.
1. The public API doesn't need to have a builder exposed. The implementation class can easily accept a `Map<String, Long>` for the beginning.
1. equals and hashCode are not really needed. The important part is that the counts are correct. Is there a reason for these 2 methods, apart the use of them in the tests for the `isEqualTo` check

modules/flowable-engine-common/src/main/resources/org/flowable/common/db/mapping/changeTenantId.xml
```diff
@@ -0,0 +1,54 @@
...
+<mapper namespace="org.flowable.common.db.mapping">
+
+    <sql id="changeTenantIdSetSql">
+        <set>
+            TENANT_ID_ = #{targetTenantId, jdbcType=VARCHAR}
+        </set>
+    </sql>
+    
+    <sql id="changeTenantIdCriteriaSql">
+        where RES.TENANT_ID_ = #{sourceTenantId, jdbcType=VARCHAR}
```
filiphr, line 28: This will not work on Oracle if the `sourceTenantId` is the default empty tenant. In Oracle an empty tenant is stored as null. Therefore the where clause will always be `false`. This is also applicable for the other places where you use the `TENANT_ID_` in the SQL.

filiphr (changes requested):
Thanks for this PR @llyorshch. 

I've did a first pass and left some comments. Here is also some more general feedback:

There is a lot of repetition for `countChangeTenantIdXXXInstances` and `changeTenantIdXXXInstances` lets add a new interface `ChangeTenantIdEntityManager` in the common package similar like we have the `JobInfoEntityManager`. Perhaps we would need another similar interface that would be applicable for the entities with scope. Doing that could make it simpler to do the count in a generic way. If we do this then we can have a `ChangeTenantIdBuilderImpl` implementation in the common package that will receive a map of key and a `ChangeTenantIdEntityManager` and do the logic in one common place.
Every engine will then instantiate the builder with the appropriate map. You could create a fake `ChangeTenantIdEntityManager` for the scoped ones on the fly then.

Please format every new file correct. Some files have no line ending, some files have 3 or 4 empty lines in the end of the file.

Add some validation for the change of tenant to tenant. Moving between the same tenants doesn't make a lot of sense.

Some of my comments are left only in one engine. However, they are applicable for all engines.

modules/flowable-engine-common-api/src/main/java/org/flowable/common/engine/api/tenant/ChangeTenantIdResult.java
llyorshch: Missing License header added. 👍🏻

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/tenant/ChangeTenantIdBuilderCmmnImpl.java
llyorshch: Missing License header added. 👍🏻

modules/flowable-cmmn-engine/src/test/java/org/flowable/cmmn/test/tenant/ChangeTenantIdCaseInstanceTest.java
llyorshch: Actually, the isEqualTo is checking the contents of the result map. The equalTo function is overriden.

modules/flowable-engine-common-api/src/main/java/org/flowable/common/engine/api/tenant/ChangeTenantIdResult.java
llyorshch: I have implemented all the change suggested. Please, review.

modules/flowable-engine-common-api/src/main/java/org/flowable/common/engine/api/tenant/ChangeTenantIdBuilder.java
llyorshch: I have applied this. Please, review.

modules/flowable-cmmn-api/src/main/java/org/flowable/cmmn/api/CmmnManagementService.java
llyorshch: fixed

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/persistence/entity/data/impl/MybatisCaseInstanceDataManagerImpl.java
llyorshch: I have moved this logic to the abstract class that each data manager expands. I have seen that something similar is done for the getIdGenerator() method.

modules/flowable-cmmn-engine/src/main/resources/org/flowable/cmmn/db/mapping/common.xml
llyorshch: Done

modules/flowable-cmmn-engine/src/test/java/org/flowable/cmmn/test/tenant/ChangeTenantIdCaseInstanceTest.java
llyorshch: Done

modules/flowable-cmmn-engine/src/test/java/org/flowable/cmmn/test/tenant/ChangeTenantIdCaseInstanceTest.java
llyorshch: Done

llyorshch:
@filiphr , please let's discuss your ChangeTenantIdEntityManager interface proposal. Sorry, but I am not sure to follow what you mean and before doing something wrong, I think is better to talk (or chat) a little bit about it. Specially, I don't understand the concept of a "fake" ChangeTenantIdEntityManager.

Looking forward to discuss this matter with you.

modules/flowable-cmmn-engine/src/test/java/org/flowable/cmmn/test/tenant/ChangeTenantIdCaseInstanceTest.java
filiphr: Yes I saw that. But the `simulationResult` is not tested as well. Both can very much be empty and the assertion will be true.

Therefore, I prefer testing the contents more.

In addition to that in `assertThat(...).isEqualTo(...).as(...)` the `as` part is obsolete. If you want it to have an impact it should be before the `isEqualTo` e.g. `assertThat(...).as(...).isEqualTo(...)`

llyorshch: I understand, @filiphr . I will apply the check of the contents in addition to the isEqualTo. Also, I will review the order of the .as() call.

llyorshch:
I close this pull request in favour of https://github.com/flowable/flowable-engine/pull/3048

PR #2947, opened 2021-06-09
Upgrade to Spring Boot 2.5.0

dbmalkovsky:
The Spring Boot Release Notes are [here]( https://github.com/spring-projects/spring-boot/releases/tag/v2.5.0).   

I made the dependency updates that I found in the release notes and ran locally "mvn clean test" with no errors (distro profile).

FWIW, none of the jars that were changed are listed in distro/src/notice.txt. 

I **did **not** do** the following upgrades as they caused compile errors in flowable.http.common (and maybe else where):

- Upgrade to HttpClient5 5.0.4
- Upgrade to HttpCore5 5.1.1

There is a big jump from the version 4 jars currently used in the project and version 5.

### DIFF
--- a/modules/flowable-app-rest/src/test/java/org/flowable/rest/app/FlowableRestApplicationSecurityTest.java
+++ b/modules/flowable-app-rest/src/test/java/org/flowable/rest/app/FlowableRestApplicationSecurityTest.java
@@ -71,7 +71,8 @@ public class FlowableRestApplicationSecurityTest {
             "mappings",
             "caches",
             "caches-cache",
-            "health-path"
+            "health-path",
+            "configprops-prefix"
         )
     );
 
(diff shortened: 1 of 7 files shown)


filiphr:
The test failures are kinda expected. Can you please add the "configprops-prefix" to the tests in the `flowable-ui-app` that failed?

dbmalkovsky:
As you were typing the message, I was adding the link to the test 😄

dbmalkovsky:
Finally; sorry it took 2 tries.

filiphr:
Thanks @dbmalkovsky

filiphr:
>I did not do the following upgrades as they caused compile errors in flowable.http.common (and maybe else where):
>
>Upgrade to HttpClient5 5.0.4
>Upgrade to HttpCore5 5.1.1
>There is a big jump from the version 4 jars currently used in the project and version 5.

Actually Spring Boot and Spring support both HttpClient5 and HttpClient4. Those can be treated as 2 different frameworks as they have different packages and different jars.

Issue #2953, opened 2021-06-15
History level switching failure


martin-grofcik:
**Describe the bug**
OneTaskProcess is started with the `HistoryLevel.NONE` and `DefaultHistoryManager`.
When `HistoryLevel.NONE` is switched to `HistoryLevel.ACTIVITY`, and a task from the process instance is completed, null pointer exception is thrown:
```
nullPointerException is thrown in class org.flowable.engine.impl.history.DefaultHistoryManager (Line 185), the variable "historicActivityInstance " is null.
```
The same problem can occur when process instance is upgraded to the newer version with  higher `HistoryLevel`.

**Expected behavior**
`taskService.completeTask()` passes correctly.

**Solution proposal:**
https://github.com/flowable/flowable-engine/pull/2954

martin-grofcik:
Solved by https://github.com/flowable/flowable-engine/pull/2954

PR #2954, opened 2021-06-15
#2953 History level change NONE -> ACTIVITY causes NPE

martin-grofcik:
#### Check List:
* Unit tests: YES
* Documentation: NA

### DIFF
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/history/ChangeHistoryLevelTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/history/ChangeHistoryLevelTest.java
@@ -1,0 +1,83 @@
+package org.flowable.engine.test.api.history;
+
+import org.flowable.common.engine.impl.history.HistoryLevel;
+import org.flowable.engine.HistoryService;
+import org.flowable.engine.ProcessEngineConfiguration;
+import org.flowable.engine.RuntimeService;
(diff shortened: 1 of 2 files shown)


modules/flowable-engine/src/test/java/org/flowable/engine/test/api/history/ChangeHistoryLevelTest.java
```diff
@@ -0,0 +1,83 @@
+package org.flowable.engine.test.api.history;
```
dbmalkovsky, line 1: Please add the standard license header comment; thanks.

modules/flowable-engine/src/test/java/org/flowable/engine/test/api/history/ChangeHistoryLevelTest.java
```diff
@@ -0,0 +1,83 @@
...
+import org.flowable.engine.test.FlowableTest;
+import org.flowable.task.api.Task;
+import org.flowable.variable.api.history.HistoricVariableInstance;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.Collections;
+
+import static org.assertj.core.api.Assertions.assertThat;
```
dbmalkovsky, line 20: Import order is wrong; statics, java, then org.

PR #2963, opened 2021-06-22
Upgrade to Spring Security 5.5.1

dbmalkovsky:
Release notes: https://github.com/spring-projects/spring-security/releases/tag/5.5.1

Only dependency that needed updating was `org.slf4j`.


dbmalkovsky:
@filiphr I'm in over my head here; there appears to be an ordering issue causing the compile errors but I can't seem to untangle the mess.   This is much more than I expected with a dotdot upgrade.

filiphr:
Seems like it is related to https://github.com/spring-projects/spring-security/issues/9767. Let's wait for the Spring Boot upgrade before moving the Spring Security version. Seems like Spring Boot is excluding it from `spring-security-config` (which we do depend on).

dbmalkovsky:
Replaced with Spring Boot upgrade [PR](https://github.com/flowable/flowable-engine/pull/2967)

Issue #2674, opened 2020-11-01
Flowable UI Modeler - Processes and Apps not displayed when using an Oracle database and no tenant


PascalSchumacher:
Flowable UI Modeler version `6.6.0` (using either `flowable-spring-boot-starter-ui-modeler` or `flowable-ui-app`) does no longer display processes and apps when it is used in combination with an Oracle database and no tenant is configured (the default).

This was working in previous version like `6.5.0` etc. It is still possible to create processes and apps.

The problem is that Oracle database saves empty strings as null.

The `tenant_id` column is `null` and and the SQL statements used by Flowable Modeler (selecting rows where `tenant_id` is `''`):

`select * from ACT_DE_MODEL WHERE model_type = 0 and tenant_id = '' order by last_updated desc`
`select * from ACT_DE_MODEL WHERE model_type = 3 and model_key = 'testapp' and tenant_id = ''`

do not find any entries.

Using `h2` as a database everything works fine.

filiphr:
Fixed in #2965

PR #2965, opened 2021-06-24
Changed the default tenant from empty string to null

4gsim:
#2674 Flowable UI Modeler - Processes and Apps not displayed when using an Oracle database and no tenant

#### Check List:
* Unit tests: NA
* Documentation: NA

### DIFF
--- a/distro/src/notice.txt
+++ b/distro/src/notice.txt
@@ -133,9 +133,9 @@ org.liquibase                   liquibase-core              4.3.5           Apac
 org.mybatis                     mybatis                     3.5.6           The Apache Software License, Version 2.0
 org.mybatis                     mybatis-spring              2.0.6           The Apache Software License, Version 2.0
 org.mvel                        mvel2                       2.2.6.Final     The Apache Software License, Version 2.0
-org.slf4j                       jcl-over-slf4j              1.7.30          MIT License
-org.slf4j                       slf4j-api                   1.7.30          MIT License
-org.slf4j                       slf4j-log4j12               1.7.30          MIT License
+org.slf4j                       jcl-over-slf4j              1.7.31          MIT License
+org.slf4j                       slf4j-api                   1.7.31          MIT License
+org.slf4j                       slf4j-log4j12               1.7.31          MIT License
 org.springframework             spring-beans                5.3.8           The Apache Software License, Version 2.0
 org.springframework             spring-core                 5.3.8           The Apache Software License, Version 2.0
 org.springframework             spring-context              5.3.8           The Apache Software License, Version 2.0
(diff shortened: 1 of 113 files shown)


filiphr:
Thanks for doing this PR @4gsim.

The fix for the problem should be in the MyBatis XML for the Modeler, not in the logic of how the default tenant is set. The default tenant is the empty string on other DBs, it is only null in Oracle (because Oracle stores empty string as null)

4gsim:
How can you fix the problem in the XML only for Oracle?
What is the reason for the default tenant being an empty string and not null?

filiphr:
>How can you fix the problem in the XML only for Oracle?

We have a lot of places in the XMLs where we do `(tenant_id = '' or tenant_id is null)`. In order to be consistent we need to do it like that for the modeler as well.


>What is the reason for the default tenant being an empty string and not null?

This is mostly a legacy reason. The tenant id has been an empty string fro years. We cannot easily change it without big migrations and we want to avoid such migrations.

4gsim:
There are a lot of cases in the xml files where this check is not done. Should we do this only for the modeler xml?

filiphr:
Can you give examples? There are some queries when you pass a tenant id, where we do not do this. However, in those queries there is `withoutTenantId` that does this. We are quite confident that the XMLs for the engines are good 😄 .

4gsim:
For example for the ProcessDefinition.xml
`       <if test="tenantId != null">
        and RES.TENANT_ID_ = #{tenantId}
      </if>
      <if test="tenantIdLike != null">
        and RES.TENANT_ID_ like #{tenantIdLike}${wildcardEscapeClause}
      </if>
      <if test="withoutTenantId">
        and (RES.TENANT_ID_ = '' or RES.TENANT_ID_ is null)
      </if>`
You are right there is a withoutTenantId flag, but it is it automatically linked to the tenantId variable, or you have to set it manually in the query? If you don't set it and set the tenantId to empty string you will get the same behavior.

filiphr:
>You are right there is a withoutTenantId flag, but it is it automatically linked to the tenantId variable, or you have to set it manually in the query? If you don't set it and set the tenantId to empty string you will get the same behavior.
Should there be the same implementation for the model xml?

You have to manually invoke `ProcessDefinitionQuery#withoutQuery`. Usually when querying by tenant id you should decide whether you want to search for entities without a query, or you want to search for entities with a specific tenant id. We don't need to change anything here.

4gsim:
I meant if for the Modeler xml we should do a similar implementation with 'withoutTenantId' flag?

filiphr:
Sorry I misunderstood you. No we don't need to do that for the Modeler. I think that it is enough to do it directly in the XML when the tenantId is the empty string. The Modeler is an end user application, not an API that someone else can use it

PascalSchumacher:
@filiphr Any updates? (It would be great to have this fix in the next flowable version imho.)

filiphr:
@PascalSchumacher, I was just in the process of merging this

PR #2975, opened 2021-06-29
Flowable-LDAP: Add option to configure connection pooling and enable …

PascalSchumacher:
…it by default.

According to my tests this makes flowable-ldap three times faster.


filiphr:
Thanks for adding this @PascalSchumacher. 3 times faster is not bad 😄 . The tests are failing because there is no `connectionPooling` in the Spring Boot `FlowableLdapProperties`. Can we add this there as well?

PascalSchumacher:
Sorry took me a while, but I have updated the pull request.

filiphr:
Thanks @PascalSchumacher

PR #2982, opened 2021-07-14
Upgrade to Spring Framework 5.3.9

dbmalkovsky:
Release Notes:  https://github.com/spring-projects/spring-framework/releases/tag/v5.3.9
No relevant dependency updates

#### Check List:
* Unit tests: NA
* Documentation: YES

PR #2988, opened 2021-07-22
Upgrade to Spring Boot 2.5.3

dbmalkovsky:
Release Notes: https://github.com/spring-projects/spring-boot/releases/tag/v2.5.3

Ran `mvn clean test` locally without error.

This replaces the Spring Framework upgrade (https://github.com/flowable/flowable-engine/pull/2982) as this includes that dependency update.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,18 +14,18 @@
 		<distributionManagementSnapshotsUrl>https://oss.sonatype.org/content/repositories/snapshots/</distributionManagementSnapshotsUrl>
 		<jdk.version>1.8</jdk.version>
 		<!-- When updating one spring version, make sure that all of them are updated to their latest compatible versions -->
-		<spring.boot.version>2.5.2</spring.boot.version>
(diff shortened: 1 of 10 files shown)


pom.xml
```diff
@@ -222,12 +222,12 @@
 			<dependency>
 				<groupId>mysql</groupId>
 				<artifactId>mysql-connector-java</artifactId>
-				<version>8.0.19</version>
+				<version>8.0.26</version>
```
filiphr, line 225:
```suggestion
				<version>8.0.19</version>
```

Lets not update MySQL for now since we know that there were some issues with versions after 8.0.19

filiphr (commented):
Thanks for the PR @dbmalkovsky. I've left a comment about the update of the MySQL JDBC driver, let's not do that for now

pom.xml
dbmalkovsky: Reverted the update.

PR #2991, opened 2021-07-23
Chore: small code cleanups

dbmalkovsky:
- remove unused imports
- correct missorted modifiers
- use parameterized logging instead of concatenation
- remove concatenation of empty string
- remove dup semicolon


dbmalkovsky:
The small changes made should not be causing issues.  It is interesting that the JDK 8 job fails with an error in org.flowable.engine.test.api.history.HistoricProcessInstanceAndVariablesQueryTest and Yvo's previous [PR](https://github.com/flowable/flowable-engine/pull/2990) has the same error.

When I run the individual test locally there are no errors.

But if I run `mvn clean test` locally without this PR the job fails in the Engine tests.

I'm guessing there is a fundamental issue somewhere else.

filiphr:
We are aware of the failing tests. I think that it will be resolved by PR #2990

PR #3000, opened 2021-08-03
Fix for escalation propagation to suspended execution

vasil-tsimashchuk:
In this contribution, an issue is fixed when boundary escalation events on call activities get triggered even if the process instance they belong to is suspended.

With the fix, escalation propagation results in an error when the corresponding catching execution is suspended.

#### Check List:
* Unit tests: YES
* Documentation: NA

### DIFF
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/event/escalation/BoundaryEscalationEventTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/event/escalation/BoundaryEscalationEventTest.java
@@ -14,14 +14,14 @@ package org.flowable.engine.test.bpmn.event.escalation;
 
 import static java.lang.String.format;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.jupiter.api.Assertions.assertThrows;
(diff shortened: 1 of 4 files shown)


modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/event/escalation/BoundaryEscalationEventTest.java
```diff
@@ -12,12 +12,16 @@
  */
 package org.flowable.engine.test.bpmn.event.escalation;
 
+import static java.lang.String.format;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
```
dbmalkovsky, line 17: Where as there are about 6 uses of `assertThrows` in the code base, the primary form used for this type of test is `org.assertj.core.api.Assertions.assertThatThrownBy;` with 1000+ uses.  Please consider changing to the more common format.  Thanks.

vasil-tsimashchuk: Feedback incorporated. Thank you!

PR #3002, opened 2021-08-05
fix unknown type in ServiceTaskParseHandler

vzickner:


tijsrademakers:
Thanks for the PR @vzickner, we've used it to understand the issue and implement a similar fix.

PR #3004, opened 2021-08-05
resolves #3003 form content type

### LINKED ISSUE #3003: Wrong mediaType reported by REST API for forms
**Describe the bug**
The `app-repository/deployments/{id}/resources` REST API returns `mediaType` as `text/xml` for `.form` files.

**Expected behavior**
Return `application/json`, which is the correct media type for forms.

**Additional context**
Found in v6.6.0 and further verified in master branch of source [DefaultContentTypeResolver](https://github.com/flowable/flowable-engine/blob/master/modules/flowable-common-rest/src/main/java/org/flowable/common/rest/resolver/DefaultContentTypeResolver.java)

Additionally found that `.app` files content type is not included in the common class but in [AppDeploymentResourceResource](https://github.com/flowable/flowable-engine/blob/master/modules/flowable-app-engine-rest/src/main/java/org/flowable/app/rest/service/api/repository/AppDeploymentResourceResource.java) so that should probably be centralised too.

tstephen:
#### Check List:
* Unit tests: NO
* Documentation:  NA


filiphr:
Thanks @tstephen

PR #3016, opened 2021-08-17
Improve Script Engine caching for concurrent access (fixes #2993).

### LINKED ISSUE #2993: ScriptingEngines fails to cache Groovy engine instance
**Describe the bug**
org.flowable.common.engine.impl.scripting.ScriptingEngines#getEngineByName is supposed to cache engines by name, but it fails to do so because of an implementation defect.

**Expected behavior**
if the cache flag is enabled an engine used from multiple threads should only instantiate one Groovy scripting engine

ScriptingEngines uses a regular HashMap for the cache; and doesn't lock it:
`cachedEngines = new HashMap<>();`

PascalSchumacher:
Without the `ConcurrentHashMap` it is possible that every thread will have its own copy of the map in its local registers.

### DIFF
--- a/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/scripting/ScriptingEngines.java
+++ b/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/scripting/ScriptingEngines.java
@@ -12,9 +12,9 @@
  */
 package org.flowable.common.engine.impl.scripting;
 
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import javax.script.Bindings;
 import javax.script.ScriptContext;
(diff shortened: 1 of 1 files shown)


modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/scripting/ScriptingEngines.java
```diff
@@ -124,7 +124,7 @@ protected ScriptEngine getEngineByName(String language) {
                     if (threadingParameter != null) {
```
PascalSchumacher, line 124: Without this check, we could just use `computeIfAbsent`.

filiphr:
Thanks for the PR @PascalSchumacher.

I've worked on something similar as well in this area. Your proposed solution is OK (in the manner that it only adds it once). However, it will still create multiple `ScriptEngine`(s). I think that the goal was not to create multiple ones. 

My current work is on [this](https://github.com/filiphr/flowable-engine/tree/script-engine-cache) branch

filiphr:
Closing this since it was resolved in https://github.com/flowable/flowable-engine/commit/2816fcabe1c1b7170b2b00ed282815add45261a0. 

Thanks for the work on this anyways @PascalSchumacher

PR #3028, opened 2021-08-25
Add ability to dinamically search for comments

DavidPLamas:
#### Check List:
* Unit tests: YES 
* Documentation: NO

I've recently encountered the need to make more complex queries for comments, more specifically I needed to be able to paginate the results through SQL queries instead of using Java due to performance reasons. 
I found a way to extend Flowable's Engine in my working context and implement all those features but I felt like this should be a good contribution for the whole community so I'm making this pull request.

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/CommentQueryImpl.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/CommentQueryImpl.java
@@ -1,0 +1,148 @@
+package org.flowable.engine.impl;
+
+import java.io.Serializable;
+import java.util.Collection;
+import java.util.Date;
+import java.util.List;
+
+import org.flowable.common.engine.api.FlowableIllegalArgumentException;
+import org.flowable.common.engine.impl.interceptor.CommandContext;
+import org.flowable.common.engine.impl.interceptor.CommandExecutor;
+import org.flowable.common.engine.impl.query.AbstractQuery;
+import org.flowable.engine.impl.cfg.ProcessEngineConfigurationImpl;
+import org.flowable.engine.task.Comment;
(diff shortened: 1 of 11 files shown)


modules/flowable-engine/src/main/java/org/flowable/engine/impl/CommentQueryImpl.java
```diff
@@ -0,0 +1,148 @@
+package org.flowable.engine.impl;
```
dbmalkovsky, line 1: Please add the standard project license header; thanks.

modules/flowable-engine/src/main/java/org/flowable/engine/impl/CommentQueryProperty.java
```diff
@@ -0,0 +1,40 @@
+package org.flowable.engine.impl;
```
dbmalkovsky, line 1: License is needed here too.

modules/flowable-engine/src/main/java/org/flowable/engine/task/CommentQuery.java
```diff
@@ -0,0 +1,91 @@
+package org.flowable.engine.task;
```
dbmalkovsky, line 1: Another missing license header

modules/flowable-engine/src/test/java/org/flowable/engine/test/api/comment/CommentQueryTest.java
```diff
@@ -0,0 +1,276 @@
+package org.flowable.engine.test.api.comment;
```
dbmalkovsky, line 1: Tests should have the license header too, thanks.

Issue #3031, opened 2021-08-30
DB2/LINUXPPC64LE is not detected correctly


el-hackerino:
**Describe the bug**
Starting Flowable on a PPC64LE system with DB2 on Linux leads to this error:
`Error creating bean with name 'processEngine': [...] couldn't deduct database type from database product name 'DB2/LINUXPPC64LE'`

**Expected behavior**
There should be no issues with this DB2 flavor.

**Additional context**
Flowable 6.5.0 within Springboot.

PR #3032, opened 2021-08-30
Added DB2/LINUXPPC64LE detection

el-hackerino:
Added DB2/LINUXPPC64LE to the database type list to fix the "Couldn't deduct database type from database product name" error found during testing on PPC. Interestingly, the type is present in `modules/flowable5-engine/src/main/java/org/activiti/engine/impl/cfg/ProcessEngineConfigurationImpl.java` but absent in 2 other locations. I'm assuming it's just an oversight.

#### Check List:
* Unit tests: NA
* Documentation: NA

Fixes https://github.com/flowable/flowable-engine/issues/3031


tijsrademakers:
Thank you for the PR

PR #3033, opened 2021-08-30
Update ch07b-BPMN-Constructs.md

im-pratham:
#### Check List:
* Unit tests: NA
* Documentation: NA


docs/docusaurus/docs/bpmn/ch07b-BPMN-Constructs.md
```diff
@@ -3536,7 +3536,7 @@ The External Worker task is implemented as a dedicated [Service Task](bpmn/ch07b
 <serviceTask id="externalWorkerOrder" flowable:type="external-worker">
 ```
 
-The External Worker task is configured by se setting the `topic` (can be an EL expression) which the External Worker uses to query for jobs to execute.
+The External Worker task is configured by setting the `topic` (can be an EL expression) which the External Worker uses to query for jobs to execute.
```
dbmalkovsky, line 3539: FWIW, this problem also exists in `ch06-cmmn.md` at line 533.

PR #3035, opened 2021-08-31
Fixing empty or condition when there are no candidateGroups

ilgrosso:
Background information: https://forum.flowable.org/t/sql-exception-with-task-query-after-upgrade-to-6-7-0/8334


filiphr:
Thanks for the PR @ilgrosso, I was already working on it and added some more tests for it as well

ilgrosso:
@filiphr I see, sounds good.

Do you think such a fix is going to be released soon with - say - 6.7.1?

filiphr:
I am not sure @ilgrosso, we will need to look into this. In any case I have posted a workaround for this in the forum question.

Instead of

```
taskService.createTaskQuery()..taskCandidateOrAssigned(authUser).list();
```

You can do

```
taskService.createTaskQuery.or().taskCandidateUser(authUser).taskAssignee(authUser).endOr().list();
```

It is a bit more verbose, but it is doable.

ilgrosso:
@filiphr yep I did already applied the suggested workaround and I confirm it works as expected.

PR #3047, opened 2021-09-18
Fix stage force complete

### LINKED ISSUE #3046: completeStage force parameter hardcoded to true
**Describe the bug**
cmmnRuntimeService.completeStagePlanItemInstance ignores the _force_ parameter, it is hardcoded to true. So, even if you pass false as the value for the _force_ parameter, it will do a force completion of the stage.

**Expected behavior**
If false is passed as second parameter, and there are active tasks on the stage, an exception should be thrown.

**Code**
```
    @Override
    public void completeStagePlanItemInstance(String planItemInstanceId, boolean force) {
        commandExecutor.execute(new CompleteStagePlanItemInstanceCmd(planItemInstanceId, true));
    }

```
https://github.com/flowable/flowable-engine/blob/main/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/runtime/CmmnRuntimeServiceImpl.java#L135

fraagol:
Fixes #3046 

#### Check List:
* Unit tests: YES
* Documentation: NA


jbarrez:
Thanks @fraagol !

PR #3048, opened 2021-09-20
Change Tenant Id feature

llyorshch:
PR changes.
- ChangeTenantIdRequest carrying all the tenant Id operation information used instead of a chain of parameters.
- ChangeTenantId commands removed from the specific engines and moved to a generic cmd in flowable-engine-common controlled by the ChangeTenantIdRequest plus the series of entity manager calls to perform in a map.
- Tests changed to check the contents of the results and to check if the source and the target tenant id are the same.
- Removed COALESCE use in oracle SQL. Instead, use of conventional IS NULL options.
- Added Process and Case in tests to cause additional entities to be tested (Jobs, event subscriotions).


tijsrademakers:
@llyorshch, there is BPMN and CMMN engine specific logic added to the common engine (for example with changeTenantId.xml and ChangeTenantIdEntityTypes). This feels unnecessary as the change tenant id logic is engine specific and can be added to the BPMN and CMMN engine.

filiphr:
Closing this in favour of PR #3064

PR #3058, opened 2021-09-25
自定义eventMessage

qiudaoke:
#### Check List:
* Unit tests: YES / NO / NA

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/ProcessEngineConfiguration.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/ProcessEngineConfiguration.java
@@ -81,6 +81,7 @@ import org.flowable.task.service.TaskPostProcessor;
  * 
  * @see ProcessEngines
  * @author Tom Baeyens
+ * @author 盘古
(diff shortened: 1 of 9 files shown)


modules/flowable-engine/src/main/java/org/flowable/engine/ProcessEngineConfiguration.java
```diff
@@ -157,6 +158,9 @@
     protected int cleanInstancesEndedAfterNumberOfDays = 365;
     protected HistoryCleaningManager historyCleaningManager;
 
+    // eventMessage
+    protected int eventMessageLength = 163;
+    protected int eventMessageSubLength = 163;
```
dbmalkovsky, line 163: Shouldn't this value be 160 and not 163 (to leave room for `...`)?

qiudaoke: ok i do it

filiphr:
Closing this since most of the logic has been implemented in #3224.

PR #3072, opened 2021-10-12
Remove collections whose contents were updated but never queried.

dbmalkovsky:
Part of the periodic check of the code base.


PascalSchumacher:
+1 imho this should be merged.

PR #3073, opened 2021-10-12
Remove redundant initializers.

dbmalkovsky:
The variable is always assigned a value further down in the code or an exception is thrown.   Initializing the variable is unneeded.

There are still around 600 or more occurrences of similar changes but submitting this to see if this change is acceptable.

Part of the periodic check of the code base.


dbmalkovsky:
Can I assume since this has been open for many months that such changes are not wanted?

tijsrademakers:
Hi David, sorry for the lack of response, yes correct we don't see this as something we want to do throughout the codebase.

PR #3082, opened 2021-10-18
Use batch for deleting data

filiphr:


modules/flowable-cmmn-engine/src/test/java/org/flowable/cmmn/test/history/HistoryDataDeleteTest.java
```diff
@@ -13,27 +13,48 @@
...
 import org.flowable.cmmn.engine.test.CmmnDeployment;
 import org.flowable.cmmn.engine.test.FlowableCmmnTestCase;
 import org.flowable.cmmn.engine.test.impl.CmmnHistoryTestHelper;
 import org.flowable.cmmn.test.itemcontrol.RepetitionVariableAggregationTest;
 import org.flowable.common.engine.impl.history.HistoryLevel;
+import org.flowable.job.api.Job;
 import org.flowable.task.api.Task;
+import org.junit.After;
 import org.junit.Test;
+import org.junit.jupiter.api.AfterEach;
```
dbmalkovsky, line 43: Import is not used.

modules/flowable-engine/src/main/java/org/flowable/engine/impl/delete/ComputeDeleteHistoricProcessInstanceStatusJobHandler.java
```diff
@@ -0,0 +1,126 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.flowable.engine.impl.delete;
+
+import java.util.List;
+
+import org.flowable.batch.api.Batch;
+import org.flowable.batch.api.BatchPart;
+import org.flowable.batch.service.impl.persistence.entity.BatchEntity;
```
dbmalkovsky, line 19: Import is not used.

filiphr: Thanks @dbmalkovsky. I fixed the two imports

Issue #2402, opened 2020-06-13
something is broken with OpenApi definition files


pietrobaricco:
**Describe the bug**
I have used [OpenApi generator](https://openapi-generator.tech/) to generate a PHP client for Flowable. Normal requests work, but When trying to create a deployment (uploadDeployment) I'm greeted by this response:

`{"message":"Internal server error","exception":"Content type 'application/octet-stream' not supported"}`

I think that this stems by this section of the OpenApi definition file:

    BinaryAttachment:
      content:
        application/octet-stream:
          schema:
            type: string
            format: binary
      required: true

Whereas the Swagger version is: (will try if it works)

    - name: "file"
        in: "formData"
        required: true
        type: "file"`

**Expected behavior**
The definition should match the required request format for the application endpoints

**Additional context**
Flowable-rest 6.5.0.6, Tomcat 7, CentOS 7, all stock

tijsrademakers:
Hi @pietrobaricco ,
Are you referring to the uploadDeployment for the BPMN engine?
Best regards

pietrobaricco:
HI @tijsrademakers, yes: the one used to load in .bpmn files. I haven't tried other methods that involve uploads, but I think that the problem is common to them all, as in the OpenApi definition files  there is a common reference to the same chunk of code (BinaryAttachment, I've posted it) wherever there is a file upload.

NicoP-S:
Hi,
I'm using Flowable v6.6.0 and have a similar problem. I also generated php classes via OpenApi generator. And I am missing all the paging and sorting params in all requests. ( list process instances, list process tasks ... ) 
I also checked the swagger ui. There is also no option for paging!

https://flowable.com/open-source/docs/bpmn/ch15-REST/#paging-and-sorting
It would be nice if these parameters could be added  to all corresponding swagger entries .

PR #3090, opened 2021-10-28
Order & sort for listHistoricTaskInstances in swagger.yaml

NicoP-S:
Mentioned in #2402 
Order and sort is missing in most of the swagger requests

### DIFF
--- a/docs/public-api/references/swagger/process/flowable-swagger-process.yaml
+++ b/docs/public-api/references/swagger/process/flowable-swagger-process.yaml
@@ -936,6 +936,36 @@ paths:
           \ set. If false, the withoutTenantId parameter is ignored."
         required: false
         type: "boolean"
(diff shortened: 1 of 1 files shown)


tijsrademakers:
Thanks for the PR, we generate the yaml files automatically, so changing them manually won't be a good solution. The REST services should be annotated so that the generation of the yaml file includes the order & sort information.

NicoP-S:
@tijsrademakers  can you tell me which file i have to edit?

filiphr:
@NicoP-S the files are generated in https://github.com/flowable/flowable-engine/blob/main/docs/public-api/tools/flowable-oas-generator/pom.xml. There are numerous endpoints that need to have the order and sort parameters. Ideally there should be a solution that will automatically add these things without the need to change all the java files

Issue #3094, opened 2021-11-04
Support thrown BPMN error in triggerable Service Task


martin-grofcik:
**Description:**
When triggerable service task throws BpmnError during the trigger execution, the exception is not handled by any event subprocess.

**Expected behavior**
When BpmnError is thrown in the trigger method of the triggerable service task, any event subprocess can handle error.

**Code**
https://github.com/martin-grofcik/flowable-engine/commit/27d6f23d422c0c728546f36c9b86465915baf56f
https://github.com/martin-grofcik/flowable-engine/tree/%233094_triggerableBpmnError

The proposed solution (jUnit test covers only one case) is to add try/catch blocks to all trigger calls.

martin-grofcik:
@tijsrademakers, @jbarrez : What is your opinion?
I can contribute with the  fix for all cases.

tijsrademakers:
@martin-grofcik , if this doesn't work right now for handling the BpmnError in event sub processes then it's a bug and good to fix.

PR #3113, opened 2021-11-21
#3094 triggerable bpmn error

martin-grofcik:
#### Check List:
* Unit tests: YES
* Documentation: NA

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/helper/ClassDelegate.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/helper/ClassDelegate.java
@@ -58,6 +58,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
  * @author Falko Menge
  * @author Saeid Mirzaei
  * @author Yvo Swillens
+ * @author martin.grofcik
  */
 public class ClassDelegate extends AbstractClassDelegate implements TaskListener, ExecutionListener, TransactionDependentExecutionListener, TransactionDependentTaskListener, SubProcessActivityBehavior, CustomPropertiesResolver {
 
@@ -216,7 +217,15 @@ public class ClassDelegate extends AbstractClassDelegate implements TaskListener
         }
 
         if (activityBehaviorInstance instanceof TriggerableActivityBehavior) {
-            ((TriggerableActivityBehavior) activityBehaviorInstance).trigger(execution, signalName, signalData);
+            try {
+                ((TriggerableActivityBehavior) activityBehaviorInstance).trigger(execution, signalName, signalData);
+            } catch (BpmnError error) {
+                ErrorPropagation.propagateError(error, execution);
+            } catch (RuntimeException e) {
+                if (!ErrorPropagation.mapException(e, (ExecutionEntity) execution, mapExceptions)) {
+                    throw e;
+                }
+            }
             if (triggerable) {
                 leave(execution);
             }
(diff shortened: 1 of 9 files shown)


martin-grofcik:
@tijsrademakers @jbarrez @filiphr : Is there any reason why pull request is not merged?

tijsrademakers:
@martin-grofcik this PR is a good improvement, but we are missing the ServiceTaskFutureJavaDelegateActivityBehavior and ServiceTaskJavaDelegateActivityBehavior changes in the same way as you did it now for the expression variant. Can you also make sure unit tests are added for these behaviors as well?

modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/helper/ClassDelegate.java
```diff
@@ -216,7 +217,15 @@ public void trigger(DelegateExecution execution, String signalName, Object signa
...
+            } catch (BpmnError error) {
+                ErrorPropagation.propagateError(error, execution);
+            } catch (RuntimeException e) {
+                if (!ErrorPropagation.mapException(e, (ExecutionEntity) execution, mapExceptions)) {
+                    throw e;
+                }
+            }
             if (triggerable) {
                 leave(execution);
             }
```
filiphr, line 231: This here is a bit strange how it works. If an BPMN error is thrown or the exception is somehow mapped then no exception is thrown and the leave will be reached. That could lead to some strange state I think. Can you please add test cases that cover this as well and make sure that we don't leave when an error happens.

filiphr (changes requested):
@martin-grofcik in addition to the changes commented by Tijs, can you please have a look at my comment as well.

martin-grofcik:
Hi @tijsrademakers,

> ServiceTaskFutureJavaDelegateActivityBehavior

Are already covered by `catchErrorThrownByTriggerableFutureJavaDelegateProvidedByDelegateExpressionOnServiceTask` .
The trigger method invocation is handled by 'ServiceTaskDelegateExpressionActivityBehavior'

> ServiceTaskJavaDelegateActivityBehavior

Test `classDelegateTriggerBpmnException` and exception is caught by `ClassDelegate`.

modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/helper/ClassDelegate.java
martin-grofcik: Hi @filiphr.

Currently `TakeOutgoing...Operation` is planned, but it is not executed because of
```
            // Execute the operation if the operation has no execution (i.e. it's an operation not working on a process instance)
            // or the operation has an execution and it is not ended
            if (operation.getExecution() == null || !operation.getExecution().isEnded()) {
```
In `CommanInvoker`.

I've put `leave(execution)` into try block to do not even plan to leave.

filiphr: That is really tangled @martin-grofcik. The fact that it works might be by chance. We shouldn't plan a leave if we don't need to leave through the flow. I'll have a look at your changes

filiphr: I checked your last changes and you still have not moved this `leave`.

filiphr:
I went through your changes @martin-grofcik, why did you add the `ActivitiAgenda` to the `flowable-engine` module?

martin-grofcik:
Hi @filiphr, yes `ActivitiAgenda` was added by mistake during the merge and also `leave`. Fixed now, thanks.

martin-grofcik:
@filiphr Is `leave` on the correct place now?

martin-grofcik:
@filiphr any update?

arthware:
As I also worked on BPMNErros for #3452 I looked through the changes and think they are fine as they are now. Would be good to have those in for 6.8.0 to round up the whole BPMNError story.

PR #3115, opened 2021-11-23
fix: DMM->DMN

oszhugc:
#### Check List:
* Unit tests:  NO 
* Documentation: YES

PR #3118, opened 2021-11-23
Fix DMM -> DMN

dbmalkovsky:


dbmalkovsky:
Maybe the `doc\userguide\src` directory should be deleted to avoid confusion as was the case in[ PR 3115](https://github.com/flowable/flowable-engine/pull/3115)

PR #3127, opened 2021-11-29
change DMNDI ids to static IDs instead of UUIDs

vzickner:
As of now IDs of DRD DI in decision services are changing for every export. This isn’t handy in case you would like to commit a DRD model on a VCS since you have unnecessary changes. This PR keeps existing IDs in case they are already available and re-uses them. Exporting a DRD twice will result in exactly the same DRD as before.

#### Check List:
* Unit tests: YES
* Documentation: NA


martin-grofcik:
+1

PR #3138, opened 2021-12-06
Added support for querying Tasks by case instance variables

WelschChristopher:
#### Check List:
* Unit tests: YES 
* Documentation: NA

### DIFF
--- a/modules/flowable-variable-service/src/main/java/org/flowable/variable/service/impl/AbstractVariableQueryImpl.java
+++ b/modules/flowable-variable-service/src/main/java/org/flowable/variable/service/impl/AbstractVariableQueryImpl.java
@@ -81,7 +81,7 @@ public abstract class AbstractVariableQueryImpl<T extends Query<?, ?>, U> extend
     }
 
     protected T variableValueEquals(Object value, String scopeType) {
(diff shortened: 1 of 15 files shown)


modules/flowable-variable-service/src/main/java/org/flowable/variable/service/impl/AbstractVariableQueryImpl.java
```diff
@@ -225,7 +304,11 @@ protected void addVariable(String name, Object value, QueryOperator operator, bo
                 throw new FlowableIllegalArgumentException("Only string values can be used with 'like' condition");
             }
         }
-        queryVariableValues.add(new QueryVariableValue(name, value, operator, localScope));
+        if (scopeType != null)
+            queryVariableValues.add(new QueryVariableValue(name, value, operator, scopeType));
+        else
+            queryVariableValues.add(new QueryVariableValue(name, value, operator, localScope));
+
```
dbmalkovsky, line 311: Minor nitpick:  the project uses braces on `if/then/else` even when the body is a single statement.

WelschChristopher: Thanks for your feedback. I added the brackets.

PR #3145, opened 2021-12-16
Allow null on dmn expressions

danielmkraus:
#### Check List:
* Unit tests: YES 
* Documentation: NA


filiphr:
Can you please explain the business use case of your change? This is a backwards incompatible change and in addition to that, it could be considered as an invalid change as well. If we decide to allow `null` to be a valid value then one could argue that `allOf(collection, null)` should return `true` if all elements of the `collection` are `null`. Similar could be set for the other functions

filiphr:
Closing this since no explanation was provided and it is a backwards incompatible change.

PR #3151, opened 2021-12-20
Fix and unit tests for adding identityLink and complete Task in same transaction

BertMatthys:
#### Check List:
* Unit tests: YES
* Documentation: not needed, bugfix


modules/flowable-engine/src/test/java/org/flowable/engine/test/api/task/TaskIdentityLinksTest.java
```diff
@@ -32,6 +35,8 @@
 
 import com.fasterxml.jackson.databind.node.ArrayNode;
 
+import static org.assertj.core.api.Assertions.*;
```
dbmalkovsky, line 38: Thanks for the PR.  Project formatting standards are not to use package imports but rather enumerate them all.  In addition static imports are at the top of the list of import statements.

PR #3168, opened 2021-12-31
Enable template name override

tstephen:
# Goal
Install flowable chart twice in the same cluster (for example as `dev` and `staging`)

## Steps to follow
- Already deployed flowable chart
- Override chart value `nameOverride: staging` and any other needed overrides like database etc.
- Deploy chart

## Expected result
- Two working flowable deployments, services, ingresses

## Actual result
- Deployment fails because the required `configmap` is already used and there is no way to override its name (as far as I can see).
```
Error: INSTALLATION FAILED: rendered manifests contain a resource that already exists. Unable to continue with install: ConfigMap "rest-configmap" in namespace "default" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "flowable-staging": current value is "flowable"
```

## Solution

- Expected `configmap` is named by `template.name`; permit this to be overridden as chart name is.

#### Check List:
* Unit tests: NO
* Documentation: NA

### DIFF
--- a/k8s/flowable/templates/_helpers.tpl
+++ b/k8s/flowable/templates/_helpers.tpl
@@ -35,7 +35,7 @@ Create chart name and version as used by the chart label.
 Template name.
 */}}
 {{- define "flowable.template" -}}
-{{- .Template.Name | replace "flowable/templates/" "" | trimSuffix ".yaml" | trunc 63 -}}
+{{- default .Template.Name .Values.nameOverride | replace "flowable/templates/" "" | trimSuffix ".yaml" | trunc 63 -}}
 {{- end -}}
 
 {{/*


yvoswillens:
Hi @tstephen 

In most cases I would advise to use different namespaces (or clusters) when there is the need to deploy the same release multiple times (for example with different configuration).
But; I can see that this could be handy in some cases.
The current changes in this PR are not sufficient I think. This change would cause some resources to have the same name. Which will cause issues.
I took a slightly different approach (prefixing the names with the release name). This enables deploying the same chart multiple times (with different release names).
Please have a look here; https://github.com/flowable/flowable-engine/commit/ff1dd833759368177b23c5533543bb3b0b6f451b

I also published a pre release Helm chart. You can test it like this;

```
helm repo add flowable-oss https://flowable.github.io/helm/
helm install flowable-staging flowable-oss/flowable --version 6.7.3-snapshot.2 --devel
```

I hope this is sufficient. I'm going to close this PR.

PR #3171, opened 2022-01-05
Extract computation of failedJobRetryTimeCycleValue in JobRetryCmd to make it overridable

vincz7777:
At some point in our development, we needed to make the computation of the failedJobRetryTimeCycle to be computed dynamically during the execution of the serviceTask (depending for instance on some input parameters provided by the user).

We propose to extract the retrieval of the failedJobRetryTimeCycleValue in a dedicated protected method, to allow to override it for such customization.

Thank you in advance,

#### Check List:
* Unit tests: NA
* Documentation: NA

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/cmd/JobRetryCmd.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/cmd/JobRetryCmd.java
@@ -71,10 +71,7 @@ public class JobRetryCmd implements Command<Object> {
         ExecutionEntity executionEntity = fetchExecutionEntity(commandContext, job.getExecutionId());
         FlowElement currentFlowElement = executionEntity != null ? executionEntity.getCurrentFlowElement() : null;
 
-        String failedJobRetryTimeCycleValue = null;
-        if (currentFlowElement instanceof ServiceTask) {
-            failedJobRetryTimeCycleValue = ((ServiceTask) currentFlowElement).getFailedJobRetryTimeCycleValue();
-        }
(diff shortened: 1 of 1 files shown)


modules/flowable-engine/src/main/java/org/flowable/engine/impl/cmd/JobRetryCmd.java
```diff
@@ -160,6 +157,13 @@ public Object execute(CommandContext commandContext) {
         return null;
     }
 
+	protected static String getFailedJobRetryTimeCycleValue(FlowElement currentFlowElement) {
+        if (currentFlowElement instanceof ServiceTask) {
+            return ((ServiceTask) currentFlowElement).getFailedJobRetryTimeCycleValue();
+        }
+		return null;
+	}
```
dbmalkovsky, line 165: Please format this code block so it aligns with surrounding code and matches the project's formatting.  Thanks.

vincz7777: Done, thanks :)

filiphr:
I am not sure that I understand the purpose of this change.

The `failedJobRetryTimeCycleValue` can be an expression, which means that you can write whatever you need to do there. If you want to change the default (when it is not modeled) you can do it in numerous different ways:

* Set the `asyncFailedJobWaitTime` or `defaultFailedJobWaitTime`
* Use a parse handler and set a `failedJobRetryTimeCycleValue` on the service tasks that do not have it

The `JobRetryCmd` is not meant to be extended.

filiphr:
Closing this PR since the change does not achieve what is explained in the description.

1. The `JobRetryCmd` is not meant to be extended
2. Even if it was, the new method is `static`, which means it cannot be overridden

PR #3180, opened 2022-01-17
Isarray null enhancement

qiudaoke:
#### Check List:
* Unit tests: YES / NO / NA


filiphr:
Thanks for the PR @qiudaoke. Have you seen NPE in the places where you did the changes? Not every place can have a null value arriving at it, in some places it is completely OK to not do a null check

qiudaoke:
> Thanks for the PR @qiudaoke. Have you seen NPE in the places where you did the changes? Not every place can have a null value arriving at it, in some places it is completely OK to not do a null check

When the customer uses it, it is scanned by the code scanning tool

filiphr:
Closing this since none of the places can lead to an NPE.

Some places use `JsonNode#path` which never returns `null` and some other places use the iteration of `ObjectNode#fields` and those cannot have a `null` value. If it is `null` then it would be `NullNode`.

Which code scanning tool is reporting this?

qiudaoke:
We can judge that these codes will not cause NPE, but Alibaba java coding guidelines tool will think that it will cause NPE
@filiphr

filiphr:
@qiudaoke if the tool is saying that there can be an NPE there, then I would consider this as a bug in the tool. We use `JsonNode#path` on purpose in order to avoid null checks

PR #3187, opened 2022-01-24
UI Modeler export app crashes with duplicate entries

mondhs:
Create BPMN and CMMN models. a CMMN model is refering to some BPMN models. Add all models to application. Click Export Application in UI moder. Export crashes as it was added same BPMN entry multiple times.

#### Check List:
* Unit tests: NA (Minor UI behavior change)
* Documentation: NA (Minor UI behavior change)


tijsrademakers:

PR #3192, opened 2022-01-27
UI Modeler cannot import ProcessTask process reference

aplatakis:
Issue:
After cmmn diagram import ProcessTask processRefExpression value is dropped
If processRef is used, then UI shows that value is not selected

Cause:
During import processRefExpression value is ignored

Details:
Process reference can be passed by processRef or processRefExpression. Modeler UI for cmmn uses processRefExpression and do not support processRef 

Solution:
* During import added condition to take processRefExpression value in case if processRef is empty. As far json converters do not know nothing about process definitions and cannot resolve process names during import(or maybe they are not imported yet), reference value is set for user to understand that value is imported. In case if user will open process definitions pop-up no definitions will be selected but still export will be able to export proper value. 
* Export updated to not check that process id was selected using process definitions pop-up. Process id is UI element and do not participate in logic execution, so it should not corrupt export of imported value even if it was not selected by user using pop-up

Unit tests: NA
Documentation: NA


tijsrademakers:

PR #3193, opened 2022-01-27
UI Modeler cannot import startEventCorrelationConfiguration value

aplatakis:
Issue:
* After cmmn diagram import case startEventCorrelationConfiguration value is dropped
* Modeler UI do not support startEventCorrelationConfiguration field for this value configuration
* On import case eventType is mapped to wrong UI element is not visible on UI

Cause:
* Wrong mapping for eventType
* Missing implementation for startEventCorrelationConfiguration support

Solution:

* Fixed eventType mapping
* Implemented startEventCorrelationConfiguration import/export. As far startEventCorrelationConfiguration supports only "storeAsUniqueReferenceId"(for other passed values do nothing), in UI added checkbox to select this option and based on that startEventCorrelationConfiguration value is imported and exported

Unit tests: NA
Documentation: NA


tijsrademakers:

PR #3198, opened 2022-01-31
Add missing caseDefinitionId for case migration

serlabel:
I tried to apply a case migration to all instances based on a specific case definition. Found out it wasn't working if I use the approach with just the case definition id – it's checking if the case definition is null, but then it's just setting it to null.

#### Check List:
* Unit tests: NO
* Documentation: NO


filiphr:
Thanks for the PR @serlabel. Can you please add some unit tests about this as well?

tijsrademakers:
Thanks, unit tests can de done in a separate PR.

PR #3201, opened 2022-02-02
Added function for send service task to be rendered on flowable UI ad…

aplatakis:
Add js function to fix issue that send event service task box was not rendered on flowable ui in process engine, when for process instance diagram was rendered

#### Check List:
* Unit tests: NA
* Documentation: NA


tijsrademakers:
Thanks!

PR #3217, opened 2022-02-17
sync Chinese document number with the English document

solojin:


tijsrademakers:
Closing PR because Chinese userguide is not supported in Flowable 7 at the moment.

PR #3219, opened 2022-02-17
BpmnXMLConverter add xsd Constructor

qiudaoke:
#### Check List:
* Unit tests: YES 
* Documentation: YES


filiphr (changes requested):
Thanks for the PR @qiudaoke.

We can't merge the PR with this change. The PR is changing a static variable from its constructor. This is not thread safe. 

Before we go ahead and discuss a potential alternative solution to this. Can you please tell us what kind of a business use case you have to change the BPMN Specification XSD that is used to perform the validation? Doing this might potentially lead to using an invalid and / or incorrect schema validation.

tijsrademakers:
Closing with feedback provided by Filip

PR #3225, opened 2022-02-23
PropertyPlaceholderConfigurer  Deprecated. as of 5.2

qiudaoke:
…tySourcesPlaceholderConfigurer instead which is more flexible through taking advantage of the Environment and PropertySource mechanisms.

#### Check List:
* Unit tests:  NO 
* Documentation: YES 
Deprecated. as of 5.2; use org.springframework.context.support.PropertySourcesPlaceholderConfigurer instead which is more flexible through taking advantage of the Environment and PropertySource mechanisms.

### DIFF
--- a/modules/flowable-engine-common/pom.xml
+++ b/modules/flowable-engine-common/pom.xml
@@ -144,6 +144,10 @@
             <artifactId>spring-beans</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-context</artifactId>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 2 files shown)


filiphr:
We do not want to have a dependency on `spring-context` on such a low level the engine. The `spring-context` is a big having a lot of configuration options. On top of that the javadoc of the `PropertyPlaceholderConfigurer` says:

>`PropertyPlaceholderConfigurer` is still appropriate for use when:
the `spring-context` module is not available (i.e., one is using Spring's `BeanFactory` API as opposed to `ApplicationContext`).

This is exactly the situation that we have. Therefore, we are going to reject this PR

PR #3250, opened 2022-03-31
fix bug to end a case task plan item

mkiener:
Fixing a bug where the case task plan item was not ended and stayed in active state if its referenced child case was manually terminated.

#### Check List:
* Unit tests: YES
* Documentation: NA

### DIFF
--- a/modules/flowable-cmmn-engine/src/test/java/org/flowable/cmmn/test/runtime/CompleteCaseTaskByManualTerminationTest.java
+++ b/modules/flowable-cmmn-engine/src/test/java/org/flowable/cmmn/test/runtime/CompleteCaseTaskByManualTerminationTest.java
@@ -1,0 +1,125 @@
+/* Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


modules/flowable-cmmn-engine/src/test/java/org/flowable/cmmn/test/runtime/CompleteCaseTaskByManualTerminationTest.java
```diff
@@ -0,0 +1,125 @@
...
+        assertThat(planItemInstances).extracting(PlanItemInstance::getName).containsExactly("Case Task");
+        assertThat(planItemInstances).extracting(PlanItemInstance::getState).containsExactly(ACTIVE);
+        assertThat(planItemInstances).extracting(PlanItemInstance::getReferenceType).containsExactly(ReferenceTypes.PLAN_ITEM_CHILD_CASE);
+        String childCaseId = planItemInstances.get(0).getReferenceId();
+
+        // manually terminate the case through the API which must also complete the case task and hence complete the root case and end it
+        cmmnRuntimeService.terminateCaseInstance(childCaseId);
+
+        HistoricCaseInstance historicCaseInstance = cmmnHistoryService.createHistoricCaseInstanceQuery().caseInstanceId(childCaseId).singleResult();
+        assertEquals(historicCaseInstance.getState(), CaseInstanceState.TERMINATED);
```
dbmalkovsky, line 60: Why isn't this using `assertThat()`?

PR #3264, opened 2022-04-05
Fix: Handling error situation in multi-instance call activities when in async completion job

BorisKlinkerSAP:
Note: We have a first proposal which is not yet covered with new tests and perhaps existing tests are currently broken. We first want to get your feedback on our suggestion (to continue this idea in the proposed way) before completing the PR with tests.

The flowable-engine dispatches a JOB_EXECUTION_FAILURE event when an exception is thrown during the EndExecutionOperation of a process instance. The failing job is converted to a DeadLetterJob.
 
We use a CallActivity to trigger (one or multiple) SubProcess instance(s). We have an event listener which handles failing subprocess instances.
 
We discovered that the context of the JOB_EXECUTION_FAILURE event is set differently depending on the asyncComplete flag of the CallActivity:
- In case asyncComplete=false the JOB_EXECUTION_FAILURE event is triggered in the context of the subprocess execution, the DeadLetterJob belongs to the subprocess execution.
- When asyncComplete=true the JOB_EXECUTION_FAILURE event is triggered in the context of the super process execution, the DeadLetterJob belongs to the super process execution.
 
From our perspective the JOB_EXECUTION_FAILURE event should always be dispatched in the context of the execution which caused the exception.
Currently this is the case for CallActivity triggered with asyncComplete=false, but not for asyncComplete=true.
Following PR corrects this behaviour.

Please let us know your opinion/feedback. Based on this we will complete this PR.

### DIFF
--- a/modules/flowable-job-service/src/main/java/org/flowable/job/service/event/impl/FlowableJobEventBuilder.java
+++ b/modules/flowable-job-service/src/main/java/org/flowable/job/service/event/impl/FlowableJobEventBuilder.java
@@ -67,8 +67,13 @@ public class FlowableJobEventBuilder {
             if (persistedObject instanceof Job) {
                 Job jobObject = (Job) persistedObject;
                 if (jobObject.getScopeType() == null) {
-                    event.setExecutionId(jobObject.getExecutionId());
-                    event.setProcessInstanceId(jobObject.getProcessInstanceId());
+                    if ("async-complete-call-actiivty".equals(jobObject.getJobHandlerType())) {
+                        event.setExecutionId(jobObject.getJobHandlerConfiguration());
+                        event.setProcessInstanceId(jobObject.getJobHandlerConfiguration());
+                    } else {
+                        event.setExecutionId(jobObject.getExecutionId());
+                        event.setProcessInstanceId(jobObject.getProcessInstanceId());
+                    }
                     event.setProcessDefinitionId(jobObject.getProcessDefinitionId());
                 } else {
                     event.setScopeType(jobObject.getScopeType());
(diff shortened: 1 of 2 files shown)


modules/flowable-job-service/src/main/java/org/flowable/job/service/event/impl/FlowableJobEventBuilder.java
```diff
@@ -67,8 +67,13 @@ protected static void populateEventWithCurrentContext(FlowableEngineEventImpl ev
             if (persistedObject instanceof Job) {
                 Job jobObject = (Job) persistedObject;
                 if (jobObject.getScopeType() == null) {
-                    event.setExecutionId(jobObject.getExecutionId());
-                    event.setProcessInstanceId(jobObject.getProcessInstanceId());
+                    if ("async-complete-call-actiivty".equals(jobObject.getJobHandlerType())) {
```
dbmalkovsky, line 70: Wouldn't it be better to use the public static variable `AsyncCompleteCallActivityJobHandler.TYPE` instead of a literal string?      At least it would hide the incorrect spelling of `activity` :smile:

BorisKlinkerSAP: > Wouldn't it be better to use the public static variable `AsyncCompleteCallActivityJobHandler.TYPE` instead of a literal string? At least it would hide the incorrect spelling of `activity` 😄

Hi @dbmalkovsky, thanks for your feedback. The typo in existing flowable code is quite unfortunate... excellent intention to cover it...
However, the change here is in FlowableJobEventBuilder which is located in maven module flowable-job-service, from here the AsyncCompleteCallActivityJobHandler class (and its constant) is not reachable as it is located in flowable-engine without dependency in that direction.

PR #3283, opened 2022-04-22
add unit tests around user event listener repetitions - 6.6

tiffmaelite:
More unit testing around what user event listener repetitions, to prove it works on 6.6 but not on 6.7. And also because more tests can not hurt.

#### Check List:
* Unit tests: YES 
* Documentation: NA


modules/flowable-cmmn-engine/src/test/resources/org/flowable/cmmn/test/runtime/UserEventListenerTest.testUserEventRepetitionWithRepeatingEntryCriteriaAndMaxInstanceCountOne.cmmn
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dbmalkovsky, line 1: Shouldn't these definitions be in the `src/test/resources/org/flowable/cmmn/test/eventlistener` package?

tijsrademakers:
Thanks for providing the tests, this has been handled in other commits that also fixed the issue.

PR #3295, opened 2022-05-03
Added bulk delete/terminate for case and process instances

WelschChristopher:
#### Check List:
* Unit tests: YES
* Documentation:  NA


modules/flowable-engine/src/test/java/org/flowable/engine/test/api/runtime/RuntimeServiceTest.java
```diff
@@ -17,6 +17,7 @@
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.assertj.core.api.Assertions.fail;
 
+import java.util.ArrayList;
```
dbmalkovsky, line 20: Looks like this import isn't used.

PR #3297, opened 2022-05-04
Improve user task name migration and allow migration of user task descriptions

Stoeoeoe:
When migrating a process instance, User Tasks were renamed without evaluating the name through the ExpressionManager. Descriptions were not migrated so far. This PR addresses these points.


modules/flowable-engine/src/main/java/org/flowable/engine/impl/dynamic/AbstractDynamicStateManager.java
```diff
@@ -892,7 +893,27 @@ protected ExecutionEntity migrateExecutionEntity(ExecutionEntity parentExecution
                     .executionId(childExecution.getId()).singleResult();
             task.setProcessDefinitionId(childExecution.getProcessDefinitionId());
             task.setTaskDefinitionKey(newFlowElement.getId());
-            task.setName(newFlowElement.getName());
+
+            // Set name
+            String name = null;
+            if(newFlowElement.getName() != null) {
```
dbmalkovsky, line 899: small nitpick: there should be a space in `if(` like other if statements.   See line 909 also.

filiphr:
Thanks for the fix @Stoeoeoe

PR #3304, opened 2022-05-10
Add debug logging before sending emails

tiffmaelite:
I recently had frequent issues around the resolution of email recipients (for example due to misconfiguration of the tenant-specific force-to parameter or because the bean called in the cc expression returned an incorrect value)
Due to the fact that many email parameters are calculated on-the-fly, errors with the mail activity are quite hard to analyze in non-local environments, and more debug logging would be very helpful.

#### Check List:
* Unit tests: NA
* Documentation: NA

### DIFF
--- a/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/behavior/impl/MailActivityBehavior.java
+++ b/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/behavior/impl/MailActivityBehavior.java
@@ -102,7 +102,7 @@ public class MailActivityBehavior extends CoreCmmnActivityBehavior {
             getFilesFromFields(attachments, planItemInstanceEntity, files, dataSources);
 
             if (StringUtils.isAllEmpty(toStr, ccStr, bccStr)) {
-                throw new FlowableException("No recipient could be found for sending email");
+                throw new FlowableException("No recipient could be found for sending email: fields to ("+ getEmailExprText(to)+"), cc ("+ getEmailExprText(cc)+") and bcc ("+ getEmailExprText(bcc)+") are all resolved to an empty value");
             }
 
             email = createEmail(textStr, htmlStr, attachmentsExist(files, dataSources));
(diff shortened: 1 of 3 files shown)


modules/flowable5-engine/src/main/java/org/activiti/engine/impl/bpmn/behavior/MailActivityBehavior.java
```diff
@@ -99,6 +103,24 @@ public void execute(DelegateExecution execution) {
             setCharset(email, charSetStr);
             attach(email, files, dataSources);
 
+            if(LOGGER.isDebugEnabled()) {
```
dbmalkovsky, line 106: Suggesting a nitpicking formatting change to match the rest of the code base: there should be a space in `if(`; it occurs in each of the files.  Thanks.

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/behavior/impl/MailActivityBehavior.java
```diff
@@ -98,7 +102,7 @@ public void execute(CommandContext commandContext, PlanItemInstanceEntity planIt
             getFilesFromFields(attachments, planItemInstanceEntity, files, dataSources);
 
             if (StringUtils.isAllEmpty(toStr, ccStr, bccStr)) {
-                throw new FlowableException("No recipient could be found for sending email");
+                throw new FlowableException("No recipient could be found for sending email: fields to ("+ getEmailExprText(to)+"), cc ("+ getEmailExprText(cc)+") and bcc ("+ getEmailExprText(bcc)+") are all resolved to an empty value");
```
dbmalkovsky, line 105: It appears this will throw a NPE if for example the `to` address is missing.  The `expression` is tested for null in `getEmailExprText` but not the value it contains.

filiphr:
Thanks a lot for this PR @tiffmaelite. Unfortunately, we will have to reject the PR in this state. The reason is that logging information like this might be in violation of GDRP, because emails or their contents might contain private user information.

We do understand the problem that you have with misconfiguration of the user task. What we would suggest is for you to open an issue in which we can discuss potential solutions for this. One solution would be to change the current mail activity behaviours and make them similar to the one that we have about HTTP. This would then also allow exposing some interceptors. People can then do whatever they need to do in these interceptors.

Thanks for your understanding.

PR #3307, opened 2022-05-13
Updated "List of process definitions" docs

verheyenkoen:
The docs currently state that the latest parameter can only be used in combination with the "key(Like)" parameter but I've found it also works with the "resourceName(Like)" parameter.

#### Check List:
* Unit tests: NA
* Documentation: YES


tijsrademakers:
Thanks

PR #3322, opened 2022-05-27
modularize channel processor

tiffmaelite:
In our project, we have to customize the registering of channels, where a custom container factory would be used for some pre-defined inbound channels only. In order to facilitate customization and extension of the existing channel processors, here is a proposal for a small modularization change.
(I also extracted an `unregisterEndpoint` method, to help keeping future change localized if at some point the opinions expressed in https://github.com/spring-projects/spring-framework/issues/24228 changed, or to allow easier customized endpoint unregistration.)

#### Check List:
* Unit tests: YES
* Documentation: NA

### DIFF
--- a/modules/flowable-event-registry-integration-test/src/test/java/org/flowable/eventregistry/integrationtest/ProcessWithDynamicChannelProcessorTestConfiguration.java
+++ b/modules/flowable-event-registry-integration-test/src/test/java/org/flowable/eventregistry/integrationtest/ProcessWithDynamicChannelProcessorTestConfiguration.java
@@ -1,0 +1,193 @@
+package org.flowable.eventregistry.integrationtest;
+
+import java.time.Duration;
+import java.util.Optional;
+
+import javax.jms.ConnectionFactory;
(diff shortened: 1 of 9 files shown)


modules/flowable-event-registry-integration-test/src/test/java/org/flowable/eventregistry/integrationtest/ProcessWithDynamicChannelProcessorTestConfiguration.java
```diff
@@ -0,0 +1,193 @@
+package org.flowable.eventregistry.integrationtest;
```
dbmalkovsky, line 1: Please include the license header; thanks.

tiffmaelite: fixed

tiffmaelite:
use case will be solved with another implementation: default channel processors are not meant to be extended

PR #3326, opened 2022-05-31
Chore: Add missing license header block

dbmalkovsky:
Code was just pushed today without the license header


filiphr:
Good catch @dbmalkovsky. Thanks

Issue #3372, opened 2022-07-07
Make exception_msg_ more than 255 characters in Job interface


loic-seguin:
Hi,

The column exception_msg_ in table act_ru_deadletter_job gets truncated at 255 characters but the max length of this column is 4000.
And it is apparently confirmed by this code in the interface JobInfo I suppose : 

```
public interface Job extends JobInfo {
    int MAX_EXCEPTION_MESSAGE_LENGTH = 255;
```

Is it possible to have this configurable in ProcessEngineConfigurationImpl for example or somewhere else ?

This feature is related to this post I have done on the [forum](https://forum.flowable.org/t/column-exception-msg-in-table-act-ru-deadletter-job-gets-truncated-to-255/9511)

Regards All,

And thanks for your feedback

PR #3382, opened 2022-07-08
Increase exception message length from 255 characters to 2000

loic-seguin:


loic-seguin:
This PR is linked to this issue https://github.com/flowable/flowable-engine/issues/3372 and this post on flowable forum https://forum.flowable.org/t/column-exception-msg-in-table-act-ru-deadletter-job-gets-truncated-to-255/9511

tijsrademakers:
Did you test this against Oracle? Because it doesn't have 4000 characters there: https://github.com/flowable/flowable-engine/blob/c755db8381106a27d3ccd11f1458a3e486bd389f/modules/flowable-job-service/src/main/resources/org/flowable/job/service/db/create/flowable.oracle.create.job.sql#L21

tijsrademakers:
@loic-seguin , it's not possible to change the column size in the database script for Oracle, because some versions of Oracle have a limitation to 2000 characters for text columns. I meant it as a reference that the value of the job info error message length can't be changed to 4000.

loic-seguin:
ha ok I didn't know that...So 2000 characters will be fine then for me...I repush

loic-seguin:
False positive on jdk 1.8 no ?

filiphr:
Yes the JDK 8 failure is a false negative.

Just to make sure that this works on different DBs can you please add some tests as well? Have a look at the `ManagementService#getXXXExceptionStacktrace()` and where it is used for ideas where you can add the tests

filiphr:
Btw, you can use `RandomStringUtils` from commons-lang3 in our tests to generate a large text message.

loic-seguin:
I have written a test. Not sure how I can import org.apache.commons.lang.RandomStringUtils in the groovy script in the test though.

filiphr:
@loic-seguin I think that easiest would be to pass the generated message as a variable in the process and use it in the script as a variable.

That way we won't need to have long lines as well :)

loic-seguin:
Done

PR #3385, opened 2022-07-11
Revert "Increase exception message length from 255 characters to 2000…

EdouardGermain:
/!\ breaking change

There is many other ways to resolve this issue but this one shouldn't be allowed. You can't dynamically change the the size of the VARCHAR in cockroachDN without throwing many exceptions.

… (#3382)"

This reverts commit fd137b20f3c0bb7029c22e3239b06a3b718cf0ad.

#### Check List:
* Unit tests: YES / NO / NA


EdouardGermain:
> Hi @EdouardGermain , can you give a bit more background into the exceptions that you shared, most specifically the Postgres one. The change just makes use of the column size that was already there. Does your postgres setup not have a 4000 VARCHAR column for the exception message?

Sorry for my english, its quite hard to explain in this langage. Do we have to drop the schema each time we upgrade flowable ? because this change doesnt allow me to run my application..

EdouardGermain:
i can work with my fork, its ok

NicolasTBD:
Same issue there, I will use your fork until they fix it, thanks.

tijsrademakers:
@EdouardGermain, @NicolasTBD , can you please share what the issue is that you run into. It's not clear what the problem is with increasing the exception message size.

filiphr:
To add to what @tijsrademakers said. Before we merged the original PR we had a look at the current definition of the `EXCEPTION_MSG_` column in the different job tables. Both Postgres and CockroachDB have 4000 for this column.

If you would like to see that then have a look at:
* https://github.com/flowable/flowable-engine/blob/main/modules/flowable-job-service/src/main/resources/org/flowable/job/service/db/create/flowable.cockroachdb.create.job.sql
* https://github.com/flowable/flowable-engine/blob/main/modules/flowable-job-service/src/main/resources/org/flowable/job/service/db/create/flowable.postgres.create.job.sql

In addition to that, our Postgres tests on our CI are green.

PR #3391, opened 2022-07-15
Make use of DynamicUserTaskCallback within InjectUserTaskInProcessIns…

domi-87:
* Unit tests: NA
* Documentation: NA


arthware:
Thx for the PR! :) Would be nice to have this code change backed by a test case, though.
Maybe you can add a test case in the class referenced below:

https://github.com/flowable/flowable-engine/blob/982be7e47facf17bb779571ba663bdf8f86947a3/modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/dynamic/DynamicBpmnInjectionTest.java#L40-L66

tijsrademakers:
Thanks for the PR!

PR #3392, opened 2022-07-15
Support script type executionListener (BPMN)

arthware:
The executionListener extensions has been adapted to also support scripting natively.

#### Check List:
* Unit tests: YES
* Documentation: NO


modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/listener/ScriptExecutingListener.java
```diff
@@ -0,0 +1,86 @@
+package org.flowable.engine.impl.bpmn.listener;
```
dbmalkovsky, line 1: This file is missing the license header.

arthware: fixed

PR #3393, opened 2022-07-15
Support of isMarkerVisible at ExclusiveGateway according to BPMN 2.0

domi-87:
* Unit tests: NA
* Documentation: NA


filiphr (changes requested):
Thanks for this PR @domi-87.

Since Flowable is not using this, but even if we decide to use it the default value currently is `true`, i.e. the marker is shown. I would suggest that we also reflect this in the Java model and only export it to the XML when the value is `false`.

In addition to that, can you please add a test case for this?

domi-87:
Hi @filiphr 

Thanks for your feedback. Can you give me a hint on how/where to best locate the test? Or rather if I can use a test as a model....

filiphr:
Have a look at the [`ParallelGatewayConverterTest`](https://github.com/flowable/flowable-engine/blob/main/modules/flowable-bpmn-converter/src/test/java/org/flowable/editor/language/xml/ParallelGatewayConverterTest.java). Let us know if you need some more information for the test case

domi-87:
@filiphr What do you think about my last commit. Adjusted the default behavior accordingly and added a test.

PR #3399, opened 2022-07-21
added the variable instance scope for RestResponseFactory and CmmnRes…

WelschChristopher:
…tResponseFactory

#### Check List:
* Unit tests: YES
* Documentation: NA

### DIFF
--- a/modules/flowable-cmmn-rest/src/main/java/org/flowable/cmmn/rest/service/api/CmmnRestResponseFactory.java
+++ b/modules/flowable-cmmn-rest/src/main/java/org/flowable/cmmn/rest/service/api/CmmnRestResponseFactory.java
@@ -598,12 +598,11 @@ public class CmmnRestResponseFactory {
         }
         result.setTaskId(variableInstance.getTaskId());
 
-        RestVariableScope scope = RestVariableScope.GLOBAL;
-        boolean localCmmn = variableInstance.getSubScopeId() != null && !variableInstance.getSubScopeId().equals(variableInstance.getScopeId()); // cmmn
-        boolean taskLocal = variableInstance.getTaskId() != null;
-
-        if (localCmmn || taskLocal) {
+        RestVariableScope scope;
+        if (variableInstance.getSubScopeId() != null && !variableInstance.getSubScopeId().equals(variableInstance.getScopeId()) ||  variableInstance.getTaskId() != null) {
             scope = RestVariableScope.LOCAL;
+        } else {
+            scope = RestVariableScope.GLOBAL;
         }
 
         result.setVariable(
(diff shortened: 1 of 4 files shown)


modules/flowable-cmmn-rest/src/main/java/org/flowable/cmmn/rest/service/api/CmmnRestResponseFactory.java
```diff
@@ -597,7 +597,18 @@ public VariableInstanceResponse createVariableInstanceResponse(VariableInstance
...
         result.setTaskId(variableInstance.getTaskId());
-        result.setVariable(createRestVariable(variableInstance.getName(), variableInstance.getValue(), null, variableInstance.getId(), VARIABLE_VARINSTANCE, false, urlBuilder));
+
+        RestVariableScope scope = RestVariableScope.GLOBAL;
+        boolean localCmmn = variableInstance.getSubScopeId() != null && !variableInstance.getSubScopeId().equals(variableInstance.getScopeId()); // cmmn
+        boolean taskLocal = variableInstance.getTaskId() != null;
+
+        if (localCmmn || taskLocal) {
+            scope = RestVariableScope.LOCAL;
+        }
```
filiphr, line 607: Instead of doing lit like this we can implement this in the following way:

```suggestion
        RestVariableScope scope;
        if (variableInstance.getSubScopeId() != null && !variableInstance.getSubScopeId().equals(variableInstance.getScopeId()) ||  variableInstance.getTaskId() != null) {
            scope = RestVariableScope.LOCAL;
        } else {
            scope = RestVariableScope.GLOBAL;
        }
```

If it is easier for you you can also extract a method that will do this logic in the if statements

modules/flowable-cmmn-rest/src/test/java/org/flowable/cmmn/rest/service/api/runtime/VariableInstanceCollectionResourceTest.java
```diff
@@ -103,7 +104,32 @@ public void testQueryVariableExcludeLocalVariable() throws Exception {
...
+
+        runtimeService.setLocalVariable(planItemInstance.getId(), "myLocalVar", "test2");
+
+        String url = CmmnRestUrls.createRelativeResourceUrl(CmmnRestUrls.URL_VARIABLE_INSTANCES);
+
+        JsonNode node = assertResultsPresentInDataResponse(url + "?caseInstanceId=" + caseInstance.getId(), 3, "myLocalVar", "test2");
+        assertThat(node.path("scope").asText()).isEqualToIgnoringCase(RestVariable.RestVariableScope.LOCAL.name());
+        node = assertResultsPresentInDataResponse(url + "?caseInstanceId=" + caseInstance.getId() + "&excludeLocalVariables=true", 1, "myVar",
+                "test1");
+        assertThat(node.path("scope").asText()).isEqualToIgnoringCase(RestVariable.RestVariableScope.GLOBAL.name());
```
filiphr, line 128: Instead of doing the `assertResultsPresentInDataResponse` I think that the test would be clearer if we get the response and do `assertThatJson` on it instead.

filiphr (changes requested):
I've left some comments for the CMMN implementation. The same comments are valid for the BPMN one. 

I think that it would make the comment clearer if we change it slightly.

PR #3407, opened 2022-07-26
Support providing initial identity links when starting a process

epictecch:
#### Check List:
* Unit tests: YES
* Documentation: NA


filiphr:
Idea was initially discussed in https://forum.flowable.org/t/inherit-owner-from-parent-process/9536

PR #3408, opened 2022-07-26
#3406 fix DbSqlSession may override existing keys silently when inser…

### LINKED ISSUE #3406: DbSqlSession may override existing key when insert entity
**Describe the bug**
`insertedObjects.get(clazz).put(entity.getId(), entity)` in `DbSqlSession` may override existing key silently, which is really hard to trace and debug the bug.

**Expected behavior**
Should avoid override or throw an Exception for what is happening, but not override the key silently.

**Additional context**
Flowable version: 6.5.0 to 6.7.2

The scene is somehow probabilistic the inserted values (etc. variables/candidate users or others) disappeared after start the process, and the possibility of occuarance can be about 10%.
I substitute the ID generator with the famous SnowFlake generator for distributed usage, but flowable is using cached bulk insert after ID generation, which makes it really fast and easy to trigger the issue.

thkdog:
fix #3406
Check key existing before put entity, give it a second chance to generate key when duplicated, throw Exception if it fails both 2 opportunities.

#### Check List:
* Unit tests: NO
* Documentation: NO

### DIFF
--- a/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/db/DbSqlSession.java
+++ b/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/db/DbSqlSession.java
@@ -76,20 +76,25 @@ public class DbSqlSession implements Session {
     // insert ///////////////////////////////////////////////////////////////////
 
     public void insert(Entity entity, IdGenerator idGenerator) {
-        if (entity.getId() == null) {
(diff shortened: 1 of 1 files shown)


filiphr:
Closing the PR since the problem was in an incorrect implementation of the Flowable ID Generator. Have a look at https://github.com/flowable/flowable-engine/issues/3406#issuecomment-1196438131 for more information

PR #3411, opened 2022-07-28
Task/refactor to variable container

arthware:
A refactoring to change the ScriptingEngines to take a VariableContainer  instead of a VariableScope (much smaller, easier-to-handle and delegate-to interface).

This refactoring makes it much easier to add additional variables to the script execution context, because VariableContainer is basically a map. VariablScope on the other hand is pretty broad and very hard to implement or provide delegation implementations.

### DIFF
--- a/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/scripting/ScriptingEngines.java
+++ b/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/scripting/ScriptingEngines.java
@@ -25,7 +25,7 @@ import javax.script.ScriptException;
 
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.flowable.common.engine.api.FlowableException;
-import org.flowable.variable.api.delegate.VariableScope;
+import org.flowable.common.engine.api.variable.VariableContainer;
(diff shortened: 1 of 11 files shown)


modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/scripting/ScriptingEngines.java
```diff
@@ -70,25 +70,46 @@ public void setScriptEngineFactories(List<ScriptEngineFactory> scriptEngineFacto
...
+        Bindings bindings = createBindings(variableContainer, null, storeScriptVariables);
         Object result = evaluate(script, language, bindings);
 
         return new ScriptEvaluationImpl(bindings, result);
     }
 
-    public Object evaluate(String script, String language, VariableScope variableScope) {
-        return evaluateWithEvaluationResult(script, language, variableScope).getResult();
+    /**
+     * Evaluates the script using scripting enginesm, creating a
```
dbmalkovsky, line 88: `enginesm` ?

arthware: done, was an unwanted commit

PR #3426, opened 2022-08-11
Improve (error) tracing for script invocations #3425

### LINKED ISSUE #3425: Improve (error) tracing for script invocations
Flowable provides the ability to use scripts at different places in models (taskListener, executionListener, script tasks, httpHandlers, human tasks etc.)

* Scripts are therefore executed originating from different contexts using the `ScriptEngines` class. 
In case of script invocations resulting in errors, it is hard to trace down which script failed, because the exception does not provide enough context information, where the error originated. This is especially true for big modeling projects relying in heavy scripting.

* The ScriptEngines shall be extended with APIs, such that clients can tag script invocations with additional meta information. This allows to add enough information on the caller side, to be able to trace down the problem and locate the origin in case of errors.

* The tracing ability shall not be restricted to errors only, but shall also allow to extract meta information about script invocations, such that they could be piped into a metrics data provider for operational analytics, like micrometer.

* Tracing and monitoring errors is the main use case

arthware:
* The ScriptEngines has been extended with APIs, such that clients can now tag script requests with meta information. This allows to add enough information to be able to trace down the origin in case a script invocation fails.
* The APIs have been designed to be flexible enough to also potentially use them as a source for operational analytics metrics using a `ScriptTraceListener` on `ScriptingEngines`.

#### Check List:
* Unit tests: YES 
* Documentation: NO 
* Fixes #3425

### DIFF
--- a/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/scripting/ScriptTrace.java
+++ b/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/scripting/ScriptTrace.java
@@ -1,0 +1,39 @@
+/* Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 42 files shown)


arthware:
NOTE: This PR is still in draft state for discussion.

modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/scripting/ScriptTrace.java
```diff
@@ -0,0 +1,116 @@
...
+ * limitations under the License.
+ */
+package org.flowable.common.engine.impl.scripting;
+
+import java.time.Duration;
+import java.util.Objects;
+import java.util.Set;
+
+/**
+ * Captures meta information about a script invocatioon, like the start time,
```
arthware, line 20: typo

PR #3431, opened 2022-08-15
ReceiveTask  setValues() method type question

LoveMyOrange:
#### Check List:
* Unit tests: NA
* Documentation: NA

Hello ,Is the (setValues()) method type of the receiveTask wrong, The current input parameter is ManualTask.
ManualTask should not appear in this location, Perhaps this is the negligence of your developers

This will cause us to always call the method of the parent class when we want to call the setValues method


filiphr:
>Hello ,Is the (setValues()) method type of the receiveTask wrong, The current input parameter is ManualTask.
ManualTask should not appear in this location, Perhaps this is the negligence of your developers

This is a typo. However, it didn't have any significant impact as the method was only calling to the parent method.
In addition to that, the word "negligence" is a strong word. I would prefer if we do not use such words for such small things.

PR #3441, opened 2022-08-18
VariableCollectionResource and VariableInstanceResponse expose the su…

WelschChristopher:
…bScopeId/ExecutionId

#### Check List:
* Unit tests: YES / NO / NA


modules/flowable-cmmn-rest/src/main/java/org/flowable/cmmn/rest/service/api/runtime/variable/VariableInstanceResponse.java
```diff
@@ -26,6 +26,7 @@ public class VariableInstanceResponse {
     protected String caseInstanceId;
     protected String caseInstanceUrl;
     protected String taskId;
+    protected String subScopeId;
```
filiphr, line 29: Let's name this `planItemInstanceId`, similar to how we are not using `scopeId`, but `caseInstanceId`

filiphr (commented):
Looks good, only one small change to align with the way we are using the `scopeId` in the CMMN `VariableInstanceResponse`

PR #3449, opened 2022-08-26
support owner and assignee within process builder API

mkiener:
Adding support for initial owner and assignee within the process instance builder API to be in sync with the case instance builder API

#### Check List:
* Unit tests: YES
* Documentation: NA


filiphr:
Manually merged in https://github.com/flowable/flowable-engine/commit/1aba176766ac1fc458d64002aa64f6e00794b5d6

Issue #3452, opened 2022-08-29
Support throwing BPMN Errors from executionListeners and taskListener


arthware:
Description to be done:

https://forum.flowable.org/t/error-handling-in-flowable-bpmn-process/9583

PR #3453, opened 2022-08-29
#3452 Support throwing bpmn error executionListener

arthware:
#### Check List:
* Unit tests: YES


arthware:
Adapted implementation approach on different branch

Issue #3400, opened 2022-07-21
I have a bug with sqlserver


wangneal:
**Describe the bug**
The flowable I use is 6.6.0，and the sqlserver I use is 2016;I ran into this problem although it did not affect use
2022-07-21 22:38:13.076 [main] INFO  o.flowable.spring.SpringProcessEngineConfiguration:1035 - Executing configure() of class org.flowable.eventregistry.spring.configurator.SpringEventRegistryConfigurator (priority:100000)
2022-07-21 22:38:13.178 [main] ERROR druid.sql.Statement:149 - {conn-10005, stmt-20005} execute error. select version()
com.microsoft.sqlserver.jdbc.SQLServerException: 'version' 不是可以识别的 内置函数名称。
	at com.microsoft.sqlserver.jdbc.SQLServerException.makeFromDatabaseError(SQLServerException.java:216)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.getNextResult(SQLServerStatement.java:1515)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.doExecuteCursored(SQLServerStatement.java:1879)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.doExecuteStatement(SQLServerStatement.java:766)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement$StmtExecCmd.doExecute(SQLServerStatement.java:689)
	at com.microsoft.sqlserver.jdbc.TDSCommand.execute(IOBuffer.java:5696)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.executeCommand(SQLServerConnection.java:1715)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeCommand(SQLServerStatement.java:180)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeStatement(SQLServerStatement.java:155)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeQuery(SQLServerStatement.java:616)
	at com.alibaba.druid.filter.FilterChainImpl.statement_executeQuery(FilterChainImpl.java:2883)
	at com.alibaba.druid.filter.FilterAdapter.statement_executeQuery(FilterAdapter.java:2514)
	at com.alibaba.druid.filter.FilterEventAdapter.statement_executeQuery(FilterEventAdapter.java:302)
[...]

PR #3464, opened 2022-09-13
Fix issue 3400: the query sql of db's version is invalid to some databases.

pandaapo:
Fix issue #3400. The query sql `select version()` is valid in MySQL, PostgreSQL or CockroachDB, but it has syntax error in Oracle, H2, HSQL, Sql Server or DB2.

### DIFF
--- a/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/AbstractEngineConfiguration.java
+++ b/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/AbstractEngineConfiguration.java
@@ -502,7 +502,19 @@ public abstract class AbstractEngineConfiguration {
 
             // CRDB does not expose the version through the jdbc driver, so we need to fetch it through version().
             if (PRODUCT_NAME_POSTGRES.equalsIgnoreCase(databaseProductName)) {
-                try (PreparedStatement preparedStatement = connection.prepareStatement("select version() as version;");
+                String dbVersionQuery = "select version() as version;";
+                if ("oracle".equalsIgnoreCase(databaseType)) {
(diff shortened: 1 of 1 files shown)


modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/AbstractEngineConfiguration.java
```diff
@@ -502,7 +502,19 @@ public void initDatabaseType() {
 
             // CRDB does not expose the version through the jdbc driver, so we need to fetch it through version().
             if (PRODUCT_NAME_POSTGRES.equalsIgnoreCase(databaseProductName)) {
```
jbarrez, line 504: The logic is in this if - i.e. it will only be triggered when databaseProductName matches postgres. 

This value is determined above:
  DatabaseMetaData databaseMetaData = connection.getMetaData();
  String databaseProductName = databaseMetaData.getDatabaseProductName();

For oracle, db2, etc. the code in the if should thus never be reached. 
How did you reproduce bug on those databases?

jbarrez: I'm not sure I follow that reasoning, the code is already generic for each database:

databaseType = databaseTypeMappings.getProperty(databaseProductName);

which is based on the database metadata. Only for CRDB/Postgres an extra check is done, because CRDB returns the same value in its metadata as postgres.

filiphr: @pandaapo I am not sure which error you are trying to fix. The error in #3400 is within Liquibase not within Flowable

PR #3466, opened 2022-09-15
Use XMLStreamReader.getElementText() instead of getText() for elements without children

roumn:
For Woodstox (used in Jboss/Wildfly) this property needs to be set, otherwise long CDATA[] parts might be split in multiple parsing events

As discussed with Filip

#### Check List:
* Unit tests: NO
* Documentation: NO

### DIFF
--- a/modules/flowable-cmmn-converter/src/main/java/org/flowable/cmmn/converter/StandardEventXmlConverter.java
+++ b/modules/flowable-cmmn-converter/src/main/java/org/flowable/cmmn/converter/StandardEventXmlConverter.java
@@ -42,7 +42,7 @@ public class StandardEventXmlConverter extends CaseElementXmlConverter {
                 TimerEventListener timerEventListener = (TimerEventListener) conversionHelper.getCurrentCmmnElement();
                 timerEventListener.setTimerStartTriggerStandardEvent(event);
             } else {
-                conversionHelper.getCurrentSentryOnPart().setStandardEvent(xtr.getText());
+                conversionHelper.getCurrentSentryOnPart().setStandardEvent(event);
             }
         } catch (XMLStreamException e) {
             throw new FlowableException("Error converting standard event", e);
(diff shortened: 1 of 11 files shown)


roumn:
I will add a test

filiphr:
>I will add a test

Thanks. I've approved the actions to run in order to make sure that everything works as expected

modules/flowable-cmmn-converter/src/main/java/org/flowable/cmmn/converter/StandardEventXmlConverter.java
```diff
@@ -34,12 +36,16 @@ public boolean hasChildElements() {
...
-            timerEventListener.setTimerStartTriggerStandardEvent(event);
-        } else {
-            conversionHelper.getCurrentSentryOnPart().setStandardEvent(xtr.getText());
+        try {
+            String event = xtr.getElementText();
+            if (conversionHelper.getCurrentCmmnElement() instanceof TimerEventListener) {
+                TimerEventListener timerEventListener = (TimerEventListener) conversionHelper.getCurrentCmmnElement();
+                timerEventListener.setTimerStartTriggerStandardEvent(event);
+            } else {
+                conversionHelper.getCurrentSentryOnPart().setStandardEvent(xtr.getText());
```
filiphr, line 45: I think that here we need to use the `event` that we got from the `xtr.getElementText()` call on line 40

filiphr (commented):
Thanks for working on an alternative solution @roumn.

I've left one comment that I think we should do. In addition to that I think that we also need to change the way we handle the condition in the `ExtensionElementsXMLConverter`

More precisely:

https://github.com/flowable/flowable-engine/blob/1740df5147c78a14246fe1ce4bc8414e98004ad0/modules/flowable-cmmn-converter/src/main/java/org/flowable/cmmn/converter/ExtensionElementsXMLConverter.java#L171-L175

I believe that it is safe to do:

```java
completionNeutralRule.setCondition(xtr.getElementText())
```

roumn:
Thank you for the feedback @filiphr, I incorporated your change suggestions.

PR #3507, opened 2022-10-26
Initialize tenant ID when evaluating adapter delegate expressions

al-matic:
of adapter channel models (Event Registry). The tenant ID should be set even though there are no variables in the VariableContainer.

#### Check List:
* Unit tests: YES
* Documentation: NA

### DIFF
--- a/modules/flowable-event-registry/src/main/java/org/flowable/eventregistry/impl/pipeline/DelegateExpressionInboundChannelModelProcessor.java
+++ b/modules/flowable-event-registry/src/main/java/org/flowable/eventregistry/impl/pipeline/DelegateExpressionInboundChannelModelProcessor.java
@@ -56,16 +56,19 @@ public class DelegateExpressionInboundChannelModelProcessor implements ChannelMo
             boolean fallbackToDefaultTenant) {
         
         if (channelModel instanceof DelegateExpressionInboundChannelModel) {
-            registerChannelModel((DelegateExpressionInboundChannelModel) channelModel, eventRegistry);
+            registerChannelModel((DelegateExpressionInboundChannelModel) channelModel, tenantId, eventRegistry);
         }
     }
 
-    protected void registerChannelModel(DelegateExpressionInboundChannelModel channelModel, EventRegistry eventRegistry) {
+    protected void registerChannelModel(DelegateExpressionInboundChannelModel channelModel, String tenantId, EventRegistry eventRegistry) {
         String delegateExpression = channelModel.getAdapterDelegateExpression();
         if (StringUtils.isNotEmpty(delegateExpression)) {
+            VariableContainerWrapper variableContainer = new VariableContainerWrapper(Collections.emptyMap());
+            variableContainer.setVariable("tenantId", tenantId);
+            variableContainer.setTenantId(tenantId);
             Object channelAdapter = engineConfiguration.getExpressionManager()
                 .createExpression(delegateExpression)
-                .getValue(new VariableContainerWrapper(Collections.emptyMap()));
+                .getValue(variableContainer);
             if (!(channelAdapter instanceof InboundEventChannelAdapter)) {
                 throw new FlowableException(
                     "DelegateExpression inbound channel model with key " + channelModel.getKey() + " resolved channel adapter delegate expression to "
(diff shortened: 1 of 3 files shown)


filiphr:
Thanks @al-matic. The changes look OK to me. However, there are no tests. Can you please add some tests and then we can merge this.

Thanks

al-matic:
@filiphr I amended the existing test, I hope it is now acceptable.

modules/flowable-event-registry/src/test/java/org/flowable/eventregistry/test/CustomEventProcessingPipelineTest.java
```diff
@@ -14,12 +14,13 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Map;
+import java.beans.FeatureDescriptor;
+import java.util.*;
```
filiphr, line 18: The project doesn't use wildcard imports. Please use the IDE settings provided in https://github.com/flowable/flowable-engine/tree/main/ide-settings

filiphr (changes requested):
I've left one comment regarding the imports.

Apart from that, I would suggest not doing so many changes in the current test. I would instead suggest adding a new test case. The changes in the `@BeforeEach`  are not needed and also the use of a custom expression manager is not needed.

What you could do is the following:

```
TestOutboundChannelAdapter testOutboundChannelAdapterAcme = new TestOutboundChannelAdapter();
TestOutboundChannelAdapter testOutboundChannelAdapterMegacorp = new TestOutboundChannelAdapter();
Map<String, TestOutboundChannelAdapter> adapters = new HashMap<>();
adapters.put("acme", testOutboundChannelAdapterAcme);
adapters.put("megacorp", testOutboundChannelAdapterMegacorp);
 Map<Object, Object> beans = eventEngineConfiguration.getExpressionManager().getBeans();
        beans.put("adapters", adapters);
        beans.put("testOutboundEventProcessingPipeline", testOutboundEventProcessingPipeline);
```


then in the expression for the adapter you can do `${adapters.get(tenantId)}`, publish an event to different tenants and verify that the right one got the right data.

You can do something similar for the inbound.

al-matic:
> then in the expression for the adapter you can do `${adapters.get(tenantId)}`

I do not think this would test my changes, because this worked before. And this type of expression is exactly what I wanted to avoid, that is being forced to specify ("hard-code") the tenant ID in the *.channel definition file. For example:

```
{
  "key": "LoanAccount",
  "name": "Loan Account Channel",
  "channelType": "inbound",
  "type": "expression",
  "adapterDelegateExpression": ${adapters.get("company_X")}
```

Instead I wanted to be able to have the exact same channel definiton for all tenants, for example:

```
{
  "key": "LoanAccount",
  "name": "Loan Account Channel",
  "channelType": "inbound",
  "type": "expression",
  "adapterDelegateExpression": ${myInboundAdapter}
```

where `${myInboundAdapter}` would somehow create a new tenant-specific adapter depending on the tenant ID of the channel definiton deployment. For this I needed to add the custom expression resolver in my project, but also for the test.

Maybe I could have achieved my goal somehow differently, not sure (I am still learning).

filiphr:
You are right @al-matic. I was thinking about something else. With your changes the expression that would work would be:

```
"adapterDelegateExpression": "${adapters.get(variableContainer.tenantId)}"
```

However thinking a bit more about it I don't like the fact that you need to use that. What I would then suggest is the following.

You change the implementation where you set the tenant id and you instead pass a map that will contain the tenantId. Doing this means that my proposed example from before would work without a problem since the default EL Resolver will kick in and provide the value (if that doesn't work we should fix it). 

Something like:

```
VariableContainerWrapper variableContainer = new VariableContainerWrapper(Collections.emptyMap());
variableContainer.setVariable("tenantId", tenantId);
variableContainer.setTenantId(tenantId);
```

The reason why I am asking this is because writing a custom `ELResolver` is a bit more complex and ideally you shouldn't need one to achieve what you are looking for. So for you you can more easily write an adapter factory that can create an adapter based on a tenant id.

e.g.

```
  "adapterDelegateExpression": "${myInboundAdapterFactory.createAdapter(tenantId)}"
```

You can of course still use a custom `ELResolver`. However, I do believe that doing an adapter factory is slightly easier to achieve this.

al-matic:
OK, if you wish I will change it, but I don't want to use this in our system:
```
  "adapterDelegateExpression": "${myInboundAdapterFactory.createAdapter(tenantId)}"
```
because I am wary that one tenant might get access to another tenant's channel adapter (if somebody deploys a *.channel file again with a "hard-coded" tenant ID for whatever reason).

al-matic:
@filiphr I impemented all of the changes that you requested.

filiphr:
Thanks @al-matic. I've manually merged this in 999071fe3a28138bb981d7adb8eaf800db5e41c2

PR #3510, opened 2022-10-29
Acquire only one exclusive external worker job per process instance

### LINKED ISSUE #3509: Job-service: Parallel exclusive external worker tasks with the same topic block acquiring external worker jobs
**Describe the bug**
When multiple exclusive external worker jobs with the same topic belonging to the same process instance are available for acquiring, no external worker jobs will be acquired.
This is caused by the AcquireExternalWorkerJobsCmd which will try to lock the same process instance for each external worker job.

Locking the process instance for the first external worker job will succeed, but the second one will cause a FlowableOptimisticLockingException.
The FlowableOptimisticLockingException will rollback the acquire process and after a couple of retries no external worker jobs will be acquired.

**Expected behavior**
When multiple exclusive external worker jobs with the same topic belonging to the same process instance are available for acquiring, non exclusive external worker jobs and exclusive worker jobs (one per process instance) will be acquired.

**Code**
ParallelExternalWorker

**Additional context**
Flowable 6.7.3-SNAPSHOT

basclaessen:
Fixes #3509

#### Check List:
* Unit tests: YES
* Documentation: NO

### DIFF
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/externalworker/ExternalWorkerServiceTaskTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/externalworker/ExternalWorkerServiceTaskTest.java
@@ -20,6 +20,7 @@ import static org.assertj.core.api.Assertions.tuple;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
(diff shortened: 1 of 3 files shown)


basclaessen:
This solution has a side effect. When one exclusive external worker job (a) is acquired and an other exclusive external worker job (b) belonging to the same process/case instance with the same topic is ready for acquiring a FlowableOptimisticLockingException is thrown and the acquire transaction is rolled back.
This will prevent other external external worker jobs which are ready for acquiring with the same topic (exclusive or not) belonging to other process/case instances to be acquired until (a) is completed.

I'll work on another solution proposal for #3509 which does not have the side effect mentioned above.

filiphr:
Thanks for working on this and good point @basclaessen. We are going to review this and try to come up with the right solution. Ideally we should try to avoid the side effect in this solution.

PR #3525, opened 2022-11-15
fix: do not attempt to delete entity if not found

ikaakkola:
When deleting an entity with ID, it is entirely possible that the entity does not exist. In such cases, do not attempt to delete it as that will cause a NPE in DbSqlSession.

### DIFF
--- a/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/db/AbstractDataManager.java
+++ b/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/db/AbstractDataManager.java
@@ -90,7 +90,9 @@ public abstract class AbstractDataManager<EntityImpl extends Entity> implements 
     @Override
     public void delete(String id) {
         EntityImpl entity = findById(id);
-        delete(entity);
+        if (entity != null) {
+            delete(entity);
+        }
     }
 
     @Override
--- a/ide-settings/eclipse/Preferences-Java-CodeStyle-OrganizeImports.importorder
+++ b/ide-settings/eclipse/Preferences-Java-CodeStyle-OrganizeImports.importorder
@@ -1,0 +1,9 @@
+#Organize Import Order
+#Tue Nov 15 14:50:24 EET 2022
+0=\#
+1=java
+2=javax
+3=org
+4=com
+5=liquibase
+6=
--- a/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/persistence/entity/AbstractEntityManager.java
+++ b/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/persistence/entity/AbstractEntityManager.java
@@ -92,7 +92,9 @@ public abstract class AbstractEntityManager<EntityImpl extends Entity, DM extend
     @Override
     public void delete(String id) {
         EntityImpl entity = findById(id);
-        delete(entity);
+        if (entity != null) {
+            delete(entity);
+        }
     }
 
     @Override


modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/db/AbstractDataManager.java
```diff
@@ -19,7 +19,6 @@
 import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
-
```
dbmalkovsky, line 22: This is a very minor issue that has no effect on the execution of the code but as per the other PR where I pointed you to the IDE settings there is a blank line between groups of imports that you are removing here.

ikaakkola: Indeed, trying to manually keep these correct is difficult.

As there was no .importorder configuration for Eclipse available, I created one that matches the IntelliJ style and appears to provide the correct results.

filiphr:
@ikaakkola this change is a really low level change that we are not willing to accept. If you have some specific use case in mind that is failing we can look into that specifically.

e.g. Deleting a user and / or a group is null safe and it is done specifically only for those two entities in their respective entity managers.

ikaakkola:
> @ikaakkola this change is a really low level change that we are not willing to accept. If you have some specific use case in mind that is failing we can look into that specifically.

I don't really see how a 'low level change' would necessarily be a bad thing and as such disagree with closing this.

Any of the callers to delete(String id) can potentially fail if the entity they are deleting has already been deleted by anything (internal, eg. another thread - should not happen, but I do not see any immediate guarantees on this, or external, eg. someone doing stupid things in the database). 

In either case - a NullPointerException is not the correct result - either ignore the failure (as this PR does, which to me seems like the correct approach) or throw some suitable exception instead (callers of the delete(String id) function seem to prefer FlowableException()).

The specific use-case where this **will happen** , is the "In memory" data-managers (https://github.com/flowable/flowable-engine/pull/3531) as they do not provide transactional guarantees on the same level as the JDBC based world does, but this only hi-lights the problem - there is nothing in the implementation guaranteeing that an entity with id _x_ is available at a later time when the actual delete is to happen.

> 
> e.g. Deleting a user and / or a group is null safe and it is done specifically only for those two entities in their respective entity managers.

deleting a user and / or group are not the only callers of the abstract EntityManager delete(String id) method, there are various other callers to the method.

jbarrez:
@ikaakkola : this is one of the oldest classes in the engine and the null check was not added for a reason: as you know, we're focussed on squeezing out as much performance as possible. While it is a micro-optimization (which in comparison with one network hop is extremely tiny), doing an if check for each entity would result in millions of null-checks. 
We did the benchmark many years ago (can't find it anymore) and back then it made a smallish improvement under extreme high load. Maybe things have changed now, no idea. But just wanted to give you some insight on why we did this, as it was a conscious decision.

If it's important for correct functioning of https://github.com/flowable/flowable-engine/pull/3531, we can still add a specific parent class for the in-memory use case.

ikaakkola:
@jbarrez it is certainly possible that there would be a _minuscule_ difference in performance, but that difference will be so small it might as well not exist.

I understand that touching the 'base' classes of the Engine is not something that should be done lightly, and I am fine with adding a specific parent class for the in-memory use case (within that PR). 

I did read more closely as to the implementations EntityManager and there are multiple implementations that do call `delete(String id)` directly , without performing any sanity checks on if the entity actually exists or not - including the one I think I encountered these NPEs in the first place:

https://github.com/flowable/flowable-engine/blob/main/modules/flowable-entitylink-service/src/main/java/org/flowable/entitylink/service/impl/persistence/entity/HistoricEntityLinkEntityManagerImpl.java does not override `delete(String id)`

 HistoricEntityLinkServiceImpl calls it here: https://github.com/flowable/flowable-engine/blob/main/modules/flowable-entitylink-service/src/main/java/org/flowable/entitylink/service/impl/HistoricEntityLinkServiceImpl.java#L76
 
 which in turn is called from here: https://github.com/flowable/flowable-engine/blob/main/modules/flowable-engine/src/main/java/org/flowable/engine/impl/history/DefaultHistoryManager.java#L441

I don't understand what guarantees that the entity is not deleted by some other thread/process/external operation between `recordEntityLinkDeleted(EntityLinkEntity entityLink)`and the actual `delete(String id)`implementation?

Having a process fail with a NullPointerException in such a case just does not seem like the correct approach at all.

PR #3527, opened 2022-11-15
fix: always get lock owner and expiration time

ikaakkola:
A jobInfoEntity might not contain owner and/or locktime, in such cases get them directly from process engine configuration. If lockOwner or time is null here, the job will effectively be unlocked due to null values in the database.

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/cfg/DefaultInternalJobManager.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/cfg/DefaultInternalJobManager.java
@@ -84,14 +84,11 @@ public class DefaultInternalJobManager extends ScopeAwareInternalJobManager {
                 CountingExecutionEntity countingExecutionEntity = (CountingExecutionEntity) execution;
                 
                 if (job instanceof TimerJobEntity) {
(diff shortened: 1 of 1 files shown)


modules/flowable-engine/src/main/java/org/flowable/engine/impl/cfg/DefaultInternalJobManager.java
```diff
@@ -18,7 +18,6 @@
 import java.util.Date;
 import java.util.GregorianCalendar;
 import java.util.List;
-
```
dbmalkovsky, line 21: This is a very minor issue that has no effect on the execution of the code but as per the other PR where I pointed you to the IDE settings there is a blank line between groups of imports that you are removing here.

ikaakkola:
Some tests appear to be a little flaky and sometimes fail. I don't have the required permissions to re-run the test jobs.

filiphr:
@ikaakkola before considering to merge this, it would be could to provide a test case that show cases the problem. In order for the scope (execution) to be locked then the job has to be locked itself. Therefore, reaching a scenario when trying to lock the scope and the `lockOwner` and `lockExpirationTime` are `null` is something that should not happen.

PR #3528, opened 2022-11-15
fix: order timer jobs by due date (oldest first)

ikaakkola:
timer jobs should be processed in due date order, and the oldest job should be handled first.


tijsrademakers:
Thanks for the PR, but we have to reject it. This will make the timer job query slower and older timer jobs have no priority over newer timer jobs that are due.

ikaakkola:
Sounds reasonable. I will leave my local override of this in place then, thank you.

PR #3531, opened 2022-11-23
Feature: purely in-memory (non-database) DataManager implementations for various entity types

### LINKED ISSUE #3532: Feature: In-Memory (non-database, non-sql) DataManagers
**Is your feature request related to a problem? Please describe.**

Running hundreds of BPMN processes (sort of as 'lambda functions') per second against in-memory databases (H2 or HSQLDB) run into performance and locking problems.

**Describe the solution you'd like**

A faster, less database heavy option to execute BPMN processes, where the entities are not stored to SQL at all, but instead are kept in-memory in a (Concurrent) Map.

**Additional context**

The implementation does not need to provide full transaction support, as the intention is to run (mostly synchronous, though asynchronous must also work) short lived BPMN processes on it. If they fail in any non-expected way, the implementation is free to simply discard data (though  creating a DeadLetter job would be good) while making sure the entities created within the failed process are also removed upon a critical failure.

ikaakkola:
(Fixes #3532)

This PR adds a new module ('flowable-inmemory-data') that provides in-memory, non-database DataManager implementations for activities, executions, variables, jobs, event subscriptions and identity links.

The in-memory DataManagers are based on concurrent maps and are _fast_ compared to their Mybatis counterparts (even for simple processes  4-5x faster than H2 or HSQLDB in-memory database). They do not suffer from optimistic locking failures but provide no true rollbacks (see Limitations below)

The intended use of the high performance In-memory DataManagers is "lambda" style execution of BPMN processes where the individual processes are very short lived, usually synchronous and with little to no external integrations, but the amount of processes executed is high (we are running this implementation with 100+ process instances/second). 

## Limitations
The DataManager implementations are for entities of active processes only - there is no history implementations of in-memory DataManagers and as such, any BPMN engine running these datamanagers should always set historyLevel to NONE.

The DataManagers do not support Native SQL queries at all. Any Native SQL query against any of the implementations will simply throw an exception.

The implementation has partial support for transactions, where inserts and deletes will be rolled back, but updates are left as-is. 

## Implementation details
The implementation is done as a separate module because some of the current modules "cheat" and access data from other modules directly on the database level. This means that for example the Timer JobDataManager cannot be implemented as part of flowable-job-service , as it needs access to process definition classes and repository service, which the 'flowable-job-service' module does not have.  Also having this as a separate module makes it clear that it is (somewhat) a custom feature that standard BPMN execution environments are probably not expected to ever use.

The entity types that this implementation supports are the ones that our testing has shown to get the most database operations (ByteArrayDataManager would be the one that is still left out of this feature for now) and as such provide the most benefit for performance. There is no _technical_ reason not to support all the DataManagers, but implementing them is a rather slow process and things like deployments and definitions do not see that much database operations.

#### Check List:
* Unit tests: YES
* Documentation: NO


ikaakkola:
*this is still missing pom.xml changes for the parent project*

modules/flowable-inmemory-data/src/main/java/org/flowable/engine/data/inmemory/AbstractEntityComparator.java
```diff
@@ -0,0 +1,122 @@
+/* 
+ * Copyright 2022 Qvantel Finland Oy
+ *
```
dbmalkovsky, line 3: Please see https://github.com/flowable/flowable-engine/blob/main/ide-settings/idea/fileTemplates/includes/Copyright.java for the project's standard copyright block; thanks

ikaakkola: The intention with Apache 2.0 license is to include the copyright owner(s) (See https://www.apache.org/licenses/LICENSE-2.0#apply) ; Copyright does not mean the licensing terms would be different - an indeed Flowable seems to contain license headers with Copyright information in place (eg. 

The copyright was not added to any _modified_ source, only to the new contributions within this PR.

Is there a legal reason to not include the Copyright part of the Apache License? Are there alternative suggestions on how to handle Copyright information? Does the Flowable Project provide some further documentation on this matter?

ikaakkola: Our company policy, like many others, is to include copyright information in Opensource Contributions. It does not necessarily need to be in the _copyright headers_, but that is the logical place.

tijsrademakers: Hi @ikaakkola , thanks for the PR and the additional information about the reason of the copyright statement. For the Flowable project we only allow the standard Apache header. Instead you can mention the author above the class construct to show that you or your company has added the source code. Would that also work for you?

tijsrademakers: Hi @ikaakkola , thanks for the follow-up feedback. What would be matching with our project standards is to use @author someone (company name) or @author company name. We don't use the word copyright in the Flowable source code and would like to keep following that pattern.

ikaakkola: I believe that will be fine as well. I will update the PR accordingly.

modules/flowable-inmemory-data/src/test/java/org/flowable/engine/data/inmemory/execution/MemoryExecutionDataManagerTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+            Date before3 = Date.from(Instant.now());
+            Thread.sleep(10);
+            processEngine.getRuntimeService().startProcessInstanceByKey("triggerableExecution", map("test-variable-1", 3));
+            ProcessInstanceQuery query = processEngine.getRuntimeService().createProcessInstanceQuery();
+
+            // instance2 should be absent from result as it falls between the
+            // dates of the or() query.
+            List<ProcessInstance> r = query.variableExists("test-variable-1").or().startedAfter(before3).startedBefore(before2).list();
+            assertThat(r.size()).isEqualTo(2);
```
dbmalkovsky, line 172:
```suggestion
            assertThat(r).hasSize(2);
```

modules/flowable-inmemory-data/src/test/java/org/flowable/engine/data/inmemory/activity/MemoryActivityDataManagerTest.java
```diff
@@ -0,0 +1,118 @@
...
+        try {
+            ProcessInstance instance = processEngine.getRuntimeService().startProcessInstanceByKey("triggerableAsyncExecution");
+            ActivityInstanceEntity activity = waitForActivity(instance.getProcessInstanceId(), "service1");
+            assertThat(activity).isNotNull();
+
+            assertThat(actManager.findById(activity.getId())).isNotNull();
+            assertThat(actManager.findActivityInstancesByExecutionIdAndActivityId(activity.getExecutionId(), "service1").size()).isEqualTo(1);
+            assertThat(actManager.findActivityInstancesByExecutionIdAndActivityId(activity.getExecutionId(), "foo").size()).isEqualTo(0);
+            assertThat(actManager.findUnfinishedActivityInstancesByExecutionAndActivityId(activity.getExecutionId(), "service1").size()).isEqualTo(1);
+            assertThat(actManager.findUnfinishedActivityInstancesByExecutionAndActivityId(activity.getExecutionId(), "foo").size()).isEqualTo(0);
```
dbmalkovsky, line 72: lines 68-72 could be replaced with:
<pre>
            assertThat(actManager.findActivityInstancesByExecutionIdAndActivityId(activity.getExecutionId(), "service1")).hasSize(1);
            assertThat(actManager.findActivityInstancesByExecutionIdAndActivityId(activity.getExecutionId(), "foo").isEmpty());
            assertThat(actManager.findUnfinishedActivityInstancesByExecutionAndActivityId(activity.getExecutionId(), "service1")).hasSize(1);
            assertThat(actManager.findUnfinishedActivityInstancesByExecutionAndActivityId(activity.getExecutionId(), "foo").isEmpty());
</pre>

dbmalkovsky: There are other occurrences of this same change in other tests that I will not enumerate.

modules/flowable-inmemory-data/src/main/java/org/flowable/engine/data/inmemory/util/QueryUtil.java
```diff
@@ -0,0 +1,389 @@
...
+                return false;
+            }
+            return requested.doubleValue() == actual.doubleValue();
+        case EXISTS:
+            return actual != null;
+        case GREATER_THAN:
+            if (actual == null) {
+                return false;
+            }
+            return actual.doubleValue() > requested.doubleValue();
```
dbmalkovsky, line 358: why is boxing of the value being done?  the values can be compared directly especially starting with JDK 5.

modules/flowable-inmemory-data/src/test/java/org/flowable/engine/data/inmemory/execution/MemoryExecutionDataManagerTest.java
ikaakkola: updated this and all similar occurrences to use more appropriate assertion methods.

modules/flowable-inmemory-data/src/main/java/org/flowable/engine/data/inmemory/util/QueryUtil.java
ikaakkola: removed unnecessary unboxing

modules/flowable-inmemory-data/src/test/java/org/flowable/engine/data/inmemory/activity/MemoryActivityDataManagerTest.java
ikaakkola: updated this and all similar occurrences to use more appropriate assertion methods.

ikaakkola:
I will rebase this to flowable-6.8.0 , marking as draft for now

ikaakkola:
@tijsrademakers @dbmalkovsky any comments on if you are interested in this feature?

Would it make more sense for you, if only the changes towards other parts of Flowable (required for this feature to function) would be included in a PR and I would maintain this module independently ?

tijsrademakers:
Hi @ikaakkola, thanks for your question. We will have a discussion about it in the coming days, and then update this PR with feedback on the feature.

zhenchuan9r:
> Hi @ikaakkola, thanks for your question. We will have a discussion about it in the coming days, and then update this PR with feedback on the feature.

Please let me know your opinion on this PR.
We had similar optimization targets but did it in a different way.

ikaakkola:
I will opt to release and maintain this feature as a separate module. To help achieve that, I will create a few PRs that make it easier to override the datamanagers without resorting to Reflection.

Issue #3467, opened 2022-09-18
TaskService.getIdentityLinksForTask(taskId)


maratkalibek:
**Describe the bug**

TaskService.getIdentityLInksForTask(taskId) is duplicating IdentityLink in results. Looks like same IdentityLink added to list twice.

**Expected behavior**

TaskService.getIdentityLInksForTask(taskId) should not duplicate IdentityLink in results.

**Code**

![image](https://user-images.githubusercontent.com/4571314/190918885-3d8a813f-158c-42f9-ac9e-4719126257ce.png)

**Additional context**

https://github.com/flowable/flowable-engine/blob/main/modules/flowable-task-service/src/main/java/org/flowable/task/service/impl/persistence/entity/TaskEntityImpl.java#L247-L270

Here, getCandidates method returns set, but getIdentityLinks returns map. So duplicates are removed for getCandidates method.

kendoziyu:
I've had a similar issue too. This problem is easy to reproduce.

**Flowable Engine Version**
As of the latest 6.7.2-release, this issue has not been resolved
```
<dependency>
    <groupId>org.flowable</groupId>
    <artifactId>flowable-engine</artifactId>
    <version>6.7.2</version>
</dependency>
```

**How to reproduce**
Just follow the [Getting Started](https://www.flowable.com/open-source/docs/bpmn/ch02-GettingStarted).

Here is my sample project zip package：
[flowable-example.zip](https://github.com/flowable/flowable-engine/files/10111322/flowable-example.zip)


Set the breakpoint to `org.flowable.engine.impl.util.IdentityLinkUtil` as shown
<img width="839" alt="taskEntity getIdentityLinks" src="https://user-images.githubusercontent.com/18433396/204482467-14c1743f-f33b-4b83-a76f-85afdc871c0d.png">

Then you see the problem
<img width="855" alt="issue" src="https://user-images.githubusercontent.com/18433396/204483741-48077c43-43da-4e00-8b95-0a6f5ffb3003.png">

arthware:
Quickly checked this: This is a runtime behavior inconsistency on the TaskEntity, which seems to happen only during process execution. Problem is gone, when the task is loaded again, or the taskService.getIdentityLinksForTask method is used. 

Root cause of the problem is, that taskEntityImpl#getIdentityLinks is not a simple getter like expected, but it initializes identity links too. 

[...]

PR #3537, opened 2022-11-29
#3467 Fix duplicate addition of same identity link in IdentityLinkUtil

arthware:
Not sure if my fix is the proper way to fix this issue. For now, it just makes the testcase pass.

* Unit tests: YES


arthware:
@filiphr It is just a quick-win fix. Not really sure, if that is ok. Fact is, that the code never shields of adding the same identity link twice. The cleaner way would probably be to add a "addIdentityLink" method on TaskEntity and handle duplicates there.

PR #3540, opened 2022-12-01
Make sure that Parallel multi instance local variables are properly transformed

### LINKED ISSUE #3536: Historic variables API crashes after terminating process instance during multi-instance sub process or call activity
**Describe the bug**
Consulting the `GET /history/historic-variable-instances` API crashes for variables "nrOfActiveInstances" and "nrOfCompletedInstances" variables when you (prematurely) stop a process instance (with the `DELETE runtime/process-instances/{processInstanceId}` API) during a multi-instance execution (eg. using the `subProcess` or `callActivity` constructs).

**Expected behavior**
Loading the historic variables should never crash obviously.

**Code**
Deploy this model ([ModelWithSubProcess.bpmn20.xml.zip](https://github.com/flowable/flowable-engine/files/10114350/ModelWithSubProcess.bpmn20.xml.zip)) to a Flowable engine, start a new process instance using the `POST runtime/process-instances` API, retrieve the process instance id from the response body and stop it again using the `DELETE runtime/process-instances/{processInstanceId}`

Then go and retrieve the historic variables using the `GET history/historic-variable-instances?processInstanceId={processInstanceId}&size=500` API.

This will result in an HTTP 500 server error with content: 
image

... and log the following exception: 
[flowable-exception.txt](https://github.com/flowable/flowable-engine/files/10114434/flowable-exception.txt)

Notice that this does work if you query for a specific variableName as long as it is not "nrOfActiveInstances" or "nrOfCompletedInstances".

If you look at the database, the most noticeable thing is that the `var_type_` in `act_hi_varinst` for both variables is "bpmnParallelMultiInstanceCompleted". This is also the case when the process is still running, but at that time the API does not "crash". If you complete the process instance in a normal manner (by submitting the 3 tasks until an end event is reached), the `var_type_` for both variables is changed to "integer" and the API also continues to work as expected.

**Additional context**
Tested in Flowable 6.7.2
Able to reproduce in our own production and UAT environments as well as in an empty spawned [docker flowable-ui container from docker hub](https://hub.docker.com/r/flowable/flowable-ui).

The same bug occurred when we use a multi-instance `callActivity` (as opposed to a mutli-instance `subProcess` in the example model above).

filiphr:
When a parallel multi instance execution is interrupted or deleted then we need to transform the temporary runtime variables into permanent historic variables

fixes #3536

#### Check List:
* Unit tests: YES
* Documentation: NA

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/MultiInstanceActivityBehavior.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/MultiInstanceActivityBehavior.java
@@ -337,6 +337,13 @@ public abstract class MultiInstanceActivityBehavior extends FlowNodeActivityBeha
     
     @Override
     public void interrupted(DelegateExecution execution) {
+        if (execution.isMultiInstanceRoot()) {
+            // We are only performing the interrupt logic for multi instance root executions
+            internalInterrupted(execution);
+        }
+    }
+
+    protected void internalInterrupted(DelegateExecution execution) {
         if (hasVariableAggregationDefinitions(execution)) {
             Map<String, VariableAggregationDefinition> aggregationsByTarget = groupAggregationsByTarget(execution, aggregations.getOverviewAggregations(),
                     CommandContextUtil.getProcessEngineConfiguration());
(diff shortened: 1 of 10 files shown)


verheyenkoen:
@filiphr Are the failing tests somehow related to this change or just flaky behavior?

filiphr:
That's only a flaky behaviour @verheyenkoen

verheyenkoen:
Any ETA on next release? Last release is almost one year old?

jbarrez:
See https://forum.flowable.org/t/new-release-any-time-soon/9636/6

PR #3559, opened 2022-12-19
feature: request body for HTTP DELETE

ikaakkola:
Added support for optional request body to the DELETE operation.

Added tests for http DELETE requests with and without body

#### Check List:
* Unit tests: YES
* Documentation: NA

### DIFF
--- a/modules/flowable-http-common/src/main/java/org/flowable/http/common/impl/apache/client5/ApacheHttpComponents5FlowableHttpClient.java
+++ b/modules/flowable-http-common/src/main/java/org/flowable/http/common/impl/apache/client5/ApacheHttpComponents5FlowableHttpClient.java
@@ -170,6 +170,7 @@ public class ApacheHttpComponents5FlowableHttpClient implements FlowableAsyncHtt
                 }
                 case "DELETE": {
                     request = AsyncRequestBuilder.delete(uri);
+                    setRequestEntity(requestInfo,request);
                     break;
                 }
                 default: {
--- a/modules/flowable-http-common/src/main/java/org/flowable/http/common/impl/apache/client5/ApacheHttpComponents5FlowableHttpClient.java
+++ b/modules/flowable-http-common/src/main/java/org/flowable/http/common/impl/apache/client5/ApacheHttpComponents5FlowableHttpClient.java
@@ -170,7 +170,7 @@ public class ApacheHttpComponents5FlowableHttpClient implements FlowableAsyncHtt
                 }
                 case "DELETE": {
                     request = AsyncRequestBuilder.delete(uri);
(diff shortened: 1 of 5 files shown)


modules/flowable-http-common/src/main/java/org/flowable/http/common/impl/apache/client5/ApacheHttpComponents5FlowableHttpClient.java
```diff
@@ -170,6 +170,7 @@ public AsyncExecutableHttpRequest prepareRequest(HttpRequest requestInfo) {
                 }
                 case "DELETE": {
                     request = AsyncRequestBuilder.delete(uri);
+                    setRequestEntity(requestInfo,request);
```
dbmalkovsky, line 173: Nitpick to match project formatting style; add a space
```suggestion
                    setRequestEntity(requestInfo, request);
```

ikaakkola: formatted sources properly.

as a side note, the Eclipse formatter settings from https://github.com/flowable/flowable-engine/tree/main/ide-settings/eclipse often change existing source files when applying formatter (hence trying to manually match formatting rules for now) - the formatter rules do not define a maximum line width at all, and some indentations for broken lines differ from what (many?) of the source files actually seem to be using..

ikaakkola:
Some tests appear to be a little flaky and sometimes fail. I don't have the required permissions to re-run the test jobs.

filiphr:
@ikaakkola don't worry about the flaky tests. If only one JDK fails then it is completely fine. We won't take those into consideration.

modules/flowable-http/src/test/java/org/flowable/http/FlowableHttpClientArgumentProvider.java
```diff
@@ -32,7 +33,8 @@ public Stream<? extends Arguments> provideArguments(ExtensionContext context) th
         HttpClientConfig config = createClientConfig();
         return Stream.of(
                 Arguments.of(new SpringWebClientFlowableHttpClient(config)),
-                Arguments.of(new ApacheHttpComponentsFlowableHttpClient(config))
+                Arguments.of(new ApacheHttpComponentsFlowableHttpClient(config)),
+                Arguments.of(new ApacheHttpComponents5FlowableHttpClient(config))
```
filiphr, line 37: Good catch with this one @ikaakkola

filiphr (approved):
Thanks for your PR @ikaakkola. It is a good improvement and great for spotting the missing http client in the `FlowableHttpClientArgumentsProvider`

PR #3595, opened 2023-01-27
Add ability to store completer of HumanTasks and UserTasks in a variable

Stoeoeoe:
When modeling processes, it is often useful to know the user who completed a task. For instance when modeling reviews, we often want to reassign a task back to the original completer of a task.
Potentially, we could also add an additional flag which allows us to store this variable locally (i.e. as a task variable) which may useful in some cases.

### DIFF
--- a/modules/flowable-bpmn-converter/src/main/resources/org/flowable/impl/bpmn/parser/flowable-bpmn-extensions.xsd
+++ b/modules/flowable-bpmn-converter/src/main/resources/org/flowable/impl/bpmn/parser/flowable-bpmn-extensions.xsd
@@ -927,7 +927,7 @@
   <attribute name="taskCompleterVariableName" type="string" default="">
     <annotation>
       <documentation>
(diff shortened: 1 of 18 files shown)


modules/flowable-bpmn-converter/src/main/resources/org/flowable/impl/bpmn/parser/flowable-bpmn-extensions.xsd
```diff
@@ -923,4 +923,12 @@
       </documentation>
     </annotation>
   </attribute>
+
+  <attribute name="taskCompleterVariableName" type="string" default="">
+    <annotation>
+      <documentation>
+        Allows to specify a variable name on a user task where the completer's user ID will be stored.
```
dbmalkovsky, line 930: Slightly better:
```suggestion
        Optional variable name on a user task where the completer's user ID will be stored.
```

PR #3644, opened 2023-04-29
The 'archetypes' pom should be included together with its child - 'flowable-archetype-unittest'; One of identical files is not to be used.

ghRocs:
#### Check List:
* Unit tests:  NA
* Documentation:  NA
* Maven Install: YES


tijsrademakers:
Thanks for the PR, but we only support PRs on the main branch, and from there it will be moved to other branches if needed. Could you do a new PR on the main branch?

PR #3645, opened 2023-05-02
Add VariableInstanceEnhancer value hook interface

arthware:

PR #3647, opened 2023-05-09
#3645 Improvements to create / update variable instance APIs to creat…

arthware:
…e / update value.

* Decouple variable creation from value set
* Allow to pass value, when inserting / updating variable
* Move tenantId to last parameter to be consistent

#### Check List:
* Unit tests: YES / NO / NA


modules/flowable-engine/src/main/java/org/flowable/engine/impl/persistence/entity/ExecutionEntityImpl.java
```diff
@@ -845,7 +845,7 @@ protected VariableInstanceEntity createVariableInstance(String variableName, Obj
         variableListenerSession.addVariableData(variableInstance.getName(), VariableListenerSessionData.VARIABLE_CREATE, 
                 variableInstance.getProcessInstanceId(), ScopeTypes.BPMN, variableInstance.getProcessDefinitionId());
         
-        Clock clock = CommandContextUtil.getProcessEngineConfiguration().getClock();
+            Clock clock = CommandContextUtil.getProcessEngineConfiguration().getClock();
```
dbmalkovsky, line 848: Nit pick: I don't think this indentation is needed and needs to be removed.

arthware: Fixed.

PR #3672, opened 2023-06-29
Fix issue with not processing correct list of fail codes with commas and spaces

shami13:
#### Check List:
* Unit tests: YES
* Documentation: NO


modules/flowable-http-common/src/test/java/org/flowable/http/common/impl/ExpressionUtilsTest.java
```diff
@@ -0,0 +1,15 @@
+package org.flowable.http.common.impl;
```
dbmalkovsky, line 1: Please include the standard license header block at the beginning of the file.

shami13: Done

modules/flowable-http-common/src/test/java/org/flowable/http/common/impl/ExpressionUtilsTest.java
```diff
@@ -0,0 +1,15 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.*;
```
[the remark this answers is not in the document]
shami13: Done

filiphr:
Thanks for the PR @shami13. Can you please remove `ExpressionUtilsTest` and add a test which is using a process for this. The reason for that is due to the fact that this should work for models, independent of how the implementation looks like. If we refactor something and stop using the `ExpressionUtils` things should keep working.

You can add a test case in `HttpServiceTaskTest`.

shami13:
> Thanks for the PR @shami13. Can you please remove `ExpressionUtilsTest` and add a test which is using a process for this. The reason for that is due to the fact that this should work for models, independent of how the implementation looks like. If we refactor something and stop using the `ExpressionUtils` things should keep working.
> 
> You can add a test case in `HttpServiceTaskTest`.

Hi @filiphr 
Thnx for the comment. I want to keep ExpssesionUtilsTest, because it's the correct Unit test checking for small changes.

And I totally agree with you about some kind of integration test in the scope of HttpServiceTaskTest. I've verified it and have found that this functionality already use in the scope of the test, and doesn't work correctly, so I fixed this bug and this functionality is already covered in HttpServiceTaskTest

shami13:
Hi @filiphr @dbmalkovsky 

Could you finalise review?

filiphr:
>Thnx for the comment. I want to keep ExpssesionUtilsTest, because it's the correct Unit test checking for small changes.

We won't keep the `ExpressionUtilsTest`. It is not the right test, in our project we don't test such small changes like that. For us the important thing is the fact that if you model fail status codes with an empty string that it gets trimmed, the fact that it uses that util is an implementation detail, if we refactor that we shouldn't need to change the tests.

I see that you modified an existing test case, so we'll need some more time to finalise the review, since we need to understand why the change in that test case is needed.

shami13:
Hi @filiphr 
I've removed this test.

Issue was that value to failStatusCode and handleStatusCode "4XX, 5XX" parsed like 2 values "4XX" and " 5XX" and check like https://github.com/flowable/flowable-engine/blob/d27f3f5feeccb340635c949d5ec1b170fb58d529/modules/flowable-http-common/src/main/java/org/flowable/http/common/impl/BaseHttpActivityDelegate.java#L182 never will work for all values except first. And in test that updated you have logic that not worked and looks like not tested correctly

filiphr:
>never will work for all values except first. And in test that updated you have logic that not worked and looks like not tested correctly

That's what I feared indeed. We'll need to adapt the test to test the intended behaviour, similar to how it is for the 4xx test

shami13:
@filiphr 
I'm not sure that get correctly point what test do you mean.

Probably I could make test like this https://github.com/flowable/flowable-engine/blob/d27f3f5feeccb340635c949d5ec1b170fb58d529/modules/flowable-http/src/test/java/org/flowable/http/bpmn/HttpServiceTaskTest.java#L522

Where checking that all error codes from fall or handle processed correctly. Is that you expected?

shami13:
@filiphr 
So, I don't need to make any changes now, await your more detailed review. Is it correct?

filiphr:
Yes that is correct @shami13. Nothing more that you need to do for now

shami13:
Hi @filiphr 
Do you have any updates ?

shami13:
@filiphr @dbmalkovsky Could we merge it?

filiphr:
@shami13 I've manually integrated this in 71cb904361f2c3e7afc52f8634721364380ac664. Sorry that it took some time, I wanted to understand why the test needed fixing. I adapted the test a bit by changing the model in order for the assertions to stay the same

PR #3707, opened 2023-08-16
Chore: add missing @Test annotation

dbmalkovsky:
Old style JUnit test method in JUnit 4 class; add missing annoation


filiphr:
Good catch @dbmalkovsky, thanks

PR #3712, opened 2023-08-18
Optimized boolean transformations

im-pratham:
#### Check List:
* Unit tests:  NA
* Documentation: NA

Optimised calls to boolean parameter checking with using primitives and not doing unnecessary boxing.


filiphr:
Thanks @im-pratham

PR #3727, opened 2023-09-01
Allow LocalDate and Instants for user and human task due dates

Stoeoeoe:
With these changes, users can now set dynamic LocalDate and Instant due dates in UserTasks and HumanTasks.

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/UserTaskActivityBehavior.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/UserTaskActivityBehavior.java
@@ -1,9 +1,9 @@
 /* Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/UserTaskActivityBehavior.java
```diff
@@ -251,10 +254,16 @@ protected void handleDueDate(CreateUserTaskBeforeContext beforeContext, Expressi
...
-                    throw new FlowableIllegalArgumentException("Due date expression does not resolve to a Date or Date string: " + activeTaskDueDate);
+                } else if (dueDate instanceof Instant) {
+                    task.setDueDate(Date.from((Instant) dueDate));
+                } else if (dueDate instanceof LocalDate) {
+                    Date localDueDate = Date.from(((LocalDate) dueDate).atStartOfDay()
+                            .atZone(ZoneId.systemDefault())
+                            .toInstant());
+                    task.setDueDate(localDueDate);
                 }
+            } else {
```
amporsim, line 265: This `else` part must be moved in the inner if-statement

Stoeoeoe: Thanks, should have caught that. I have added an additional test to test for invalid dates as well.

jbarrez:
Many thanks. Merged manually https://github.com/flowable/flowable-engine/commit/70faf36b652bfd30f30fb99c6dbf4ed89d2bb637 and https://github.com/flowable/flowable-engine/commit/e636aa3c423d59030caedd3759ecaabbd0c7fa32

PR #3731, opened 2023-09-12
Migration: Allow to configure a pre/post expression which is executed…

amporsim:
For process migration we allow the users to define expressions which are triggered before/after the migration, this possibilty is missing currently for case migration. There is already an option to register a custom CaseInstanceMigrationCallback, but this callback service would be executed for every case migration. Often the migration looks different from case to case.

### DIFF
--- a/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/migration/CaseInstanceMigrationManagerImpl.java
+++ b/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/migration/CaseInstanceMigrationManagerImpl.java
@@ -256,6 +256,10 @@ public class CaseInstanceMigrationManagerImpl extends AbstractCmmnDynamicStateMa
 
     protected void doMigrateCaseInstance(CaseInstanceEntity caseInstance, CaseDefinition caseDefinitionToMigrateTo, CaseInstanceMigrationDocument document, CommandContext commandContext) {
         LOGGER.debug("Start migration of case instance with Id:'{}' to case definition identified by {}", caseInstance.getId(), printCaseDefinitionIdentifierMessage(document));
+        if (document.getPreUpgradeExpression() != null && !document.getPreUpgradeExpression().isEmpty()) {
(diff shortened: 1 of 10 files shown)


modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/migration/CaseInstanceMigrationManagerImpl.java
```diff
@@ -256,6 +256,10 @@ public void migrateHistoricCaseInstancesOfCaseDefinition(String caseDefinitionId
 
     protected void doMigrateCaseInstance(CaseInstanceEntity caseInstance, CaseDefinition caseDefinitionToMigrateTo, CaseInstanceMigrationDocument document, CommandContext commandContext) {
         LOGGER.debug("Start migration of case instance with Id:'{}' to case definition identified by {}", caseInstance.getId(), printCaseDefinitionIdentifierMessage(document));
+        if(document.getPreUpgradeExpression() != null && !document.getPreUpgradeExpression().isEmpty()){
```
dbmalkovsky, line 259:
```suggestion
        if (document.getPreUpgradeExpression() != null && !document.getPreUpgradeExpression().isEmpty()) {
```
Typically the project uses spaces here

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/migration/CaseInstanceMigrationManagerImpl.java
```diff
@@ -293,6 +297,10 @@ protected void doMigrateCaseInstance(CaseInstanceEntity caseInstance, CaseDefini
                 caseInstanceMigrationCallback.caseInstanceMigrated(caseInstance, caseDefinitionToMigrateTo, document);
             }
         }
+
+        if(document.getPostUpgradeExpression() != null && !document.getPostUpgradeExpression().isEmpty()){
```
dbmalkovsky, line 301:
```suggestion
        if (document.getPostUpgradeExpression() != null && !document.getPostUpgradeExpression().isEmpty()) {
```

PR #3734, opened 2023-09-14
no longer set category job for every element on xml if category was set in config enableCategory

nhatanh2996:
…et on config

#### Check List:
* Unit tests: YES / NO / NA

### DIFF
--- a/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/util/JobUtil.java
+++ b/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/util/JobUtil.java
@@ -26,6 +26,7 @@ import org.flowable.common.engine.api.scope.ScopeTypes;
 import org.flowable.common.engine.api.variable.VariableContainer;
 import org.flowable.job.service.JobService;
 import org.flowable.job.service.impl.persistence.entity.JobEntity;
+import org.springframework.util.CollectionUtils;
 
 /**
  * @author Filip Hrisafov
@@ -62,18 +63,22 @@ public class JobUtil {
             job.setElementName(((CaseElement) baseElement).getName());
         }
 
-        List<ExtensionElement> jobCategoryElements = baseElement.getExtensionElements().get("jobCategory");
-        if (jobCategoryElements != null && jobCategoryElements.size() > 0) {
-            ExtensionElement jobCategoryElement = jobCategoryElements.get(0);
-            if (StringUtils.isNotEmpty(jobCategoryElement.getElementText())) {
-                Expression categoryExpression = cmmnEngineConfiguration.getExpressionManager().createExpression(jobCategoryElement.getElementText());
-                Object categoryValue = categoryExpression.getValue(variableContainer);
-                if (categoryValue != null) {
-                    job.setCategory(categoryValue.toString());
+        if(CollectionUtils.isEmpty(cmmnEngineConfiguration.getEnabledJobCategories())){
+            List<ExtensionElement> jobCategoryElements = baseElement.getExtensionElements().get("jobCategory");
+            if (jobCategoryElements != null && jobCategoryElements.size() > 0) {
+                ExtensionElement jobCategoryElement = jobCategoryElements.get(0);
+                if (StringUtils.isNotEmpty(jobCategoryElement.getElementText())) {
+                    Expression categoryExpression = cmmnEngineConfiguration.getExpressionManager().createExpression(jobCategoryElement.getElementText());
+                    Object categoryValue = categoryExpression.getValue(variableContainer);
+                    if (categoryValue != null) {
+                        job.setCategory(categoryValue.toString());
+                    }
                 }
             }
+        }else{
+            String category = cmmnEngineConfiguration.getEnabledJobCategories().get(0);
+            job.setCategory(category);
         }
-
 
         job.setTenantId(variableContainer.getTenantId());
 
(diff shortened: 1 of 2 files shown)


modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/util/JobUtil.java
```diff
@@ -62,19 +63,23 @@ protected static JobEntity createJob(VariableContainer variableContainer, BaseEl
...
 
-        List<ExtensionElement> jobCategoryElements = baseElement.getExtensionElements().get("jobCategory");
-        if (jobCategoryElements != null && jobCategoryElements.size() > 0) {
-            ExtensionElement jobCategoryElement = jobCategoryElements.get(0);
-            if (StringUtils.isNotEmpty(jobCategoryElement.getElementText())) {
-                Expression categoryExpression = cmmnEngineConfiguration.getExpressionManager().createExpression(jobCategoryElement.getElementText());
-                Object categoryValue = categoryExpression.getValue(variableContainer);
-                if (categoryValue != null) {
-                    job.setCategory(categoryValue.toString());
+        if(CollectionUtils.isEmpty(cmmnEngineConfiguration.getEnabledJobCategories())){
```
dbmalkovsky, line 66:
```suggestion
        if (CollectionUtils.isEmpty(cmmnEngineConfiguration.getEnabledJobCategories())) {
```
The project standards are to use spaces

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/util/JobUtil.java
```diff
@@ -62,19 +63,23 @@ protected static JobEntity createJob(VariableContainer variableContainer, BaseEl
...
+                ExtensionElement jobCategoryElement = jobCategoryElements.get(0);
+                if (StringUtils.isNotEmpty(jobCategoryElement.getElementText())) {
+                    Expression categoryExpression = cmmnEngineConfiguration.getExpressionManager().createExpression(jobCategoryElement.getElementText());
+                    Object categoryValue = categoryExpression.getValue(variableContainer);
+                    if (categoryValue != null) {
+                        job.setCategory(categoryValue.toString());
+                    }
                 }
             }
+        }else{
```
dbmalkovsky, line 78:
```suggestion
        } else {
```

modules/flowable-engine/src/main/java/org/flowable/engine/impl/util/JobUtil.java
```diff
@@ -45,18 +46,25 @@ public static JobEntity createJob(ExecutionEntity execution, BaseElement baseEle
...
-        List<ExtensionElement> jobCategoryElements = baseElement.getExtensionElements().get("jobCategory");
-        if (jobCategoryElements != null && jobCategoryElements.size() > 0) {
-            ExtensionElement jobCategoryElement = jobCategoryElements.get(0);
-            if (StringUtils.isNotEmpty(jobCategoryElement.getElementText())) {
-                Expression categoryExpression = processEngineConfiguration.getExpressionManager().createExpression(jobCategoryElement.getElementText());
-                Object categoryValue = categoryExpression.getValue(execution);
-                if (categoryValue != null) {
-                    job.setCategory(categoryValue.toString());
+
+        if(CollectionUtils.isEmpty(processEngineConfiguration.getEnabledJobCategories())){
```
dbmalkovsky, line 50:
```suggestion
        if (CollectionUtils.isEmpty(processEngineConfiguration.getEnabledJobCategories())) {
```

modules/flowable-engine/src/main/java/org/flowable/engine/impl/util/JobUtil.java
```diff
@@ -45,18 +46,25 @@ public static JobEntity createJob(ExecutionEntity execution, BaseElement baseEle
...
+                ExtensionElement jobCategoryElement = jobCategoryElements.get(0);
+                if (StringUtils.isNotEmpty(jobCategoryElement.getElementText())) {
+                    Expression categoryExpression = processEngineConfiguration.getExpressionManager().createExpression(jobCategoryElement.getElementText());
+                    Object categoryValue = categoryExpression.getValue(execution);
+                    if (categoryValue != null) {
+                        job.setCategory(categoryValue.toString());
+                    }
                 }
             }
+        }else{
```
dbmalkovsky, line 62:
```suggestion
        } else {
```

nhatanh2996: ok I forgot format, hihi

nhatanh2996: @dbmalkovsky done, pls re-check

tijsrademakers:
Hi, thanks for the PR. If we want to have a default job category then it should be a new configuration property in the process engine and cmmn engine configuration. We should not just pick the first one in the list of enabled job categories. Is that something you would like to change in the PR?

nhatanh2996:
> Hi, thanks for the PR. If we want to have a default job category then it should be a new configuration property in the process engine and cmmn engine configuration. We should not just pick the first one in the list of enabled job categories. Is that something you would like to change in the PR?

ya maybe we need to new config for default jobCategory. wait my update

nhatanh2996:
@tijsrademakers re-check pls

tijsrademakers:
The idea of the default job category looks good, but would be good to check if the default job category is in the enabled job categories already before adding it. And when checking the BPMN/CMMN extension element, this should always happen also when the default job category is set. Because it should always be possible to overwrite the default job category with the BPMN/CMMN extension element.

Issue #3659, opened 2023-06-01
createHistoricProcessInstanceQuery().list() returns DefinitionCategory.


duan847:
```
historyService.createHistoricProcessInstanceQuery().processDefinitionCategory("test").list();
```
This interface can use `processDefinitionCategory()` as the query condition, but the return value does not include `DefinitionCategory`.

filiphr:
This has been done as part of PR #3743

PR #3743, opened 2023-09-15
#3659: Return definition category

amporsim:
#### Check List:
* Unit tests: YES


amporsim:
Sorry I have overlooked the existing PR https://github.com/flowable/flowable-engine/pull/3660
This PR does also cover the runtime process instances.

modules/flowable-engine/src/main/java/org/flowable/engine/impl/persistence/entity/ExecutionEntityImpl.java
```diff
@@ -168,14 +168,19 @@ public class ExecutionEntityImpl extends AbstractBpmnEngineVariableScopeEntity i
      */
     protected Integer processDefinitionVersion;
 
+    /**
+     * persisted reference to the process definition version.
```
dbmalkovsky, line 172:
```suggestion
     * Persisted reference to the process definition version.
```
I know this is insignificant but to be consistent with other Javadoc (see the next couple of variables)

PR #3753, opened 2023-09-26
Handle Error in the execution of JavaDelegate

loic-seguin:
Hello everyone,

This pull request is related to this ticket in this forum https://forum.flowable.org/t/oom-error-and-incoherent-datas-in-db/10869.

We are using flowable 6.8.0 and we had an OutOfMemory error in production which leaded to 387 activity instances with 'endTime' set to null and with no related job in database. 

The process instances that were linked to these incoherent activity instances were stucked.

We had to re-create a job manually in database to make the process instances go forward. 

In this PR, we reproduced the case and caught a Throwable to handle this case.

It is quite urgent for us :)

It would be good to create a minor 6.8.1 as well! 

For now we are far from upgrading to 7.0.0.

Thank you very much for your concern.

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/ServiceTaskJavaDelegateActivityBehavior.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/ServiceTaskJavaDelegateActivityBehavior.java
@@ -114,6 +114,13 @@ public class ServiceTaskJavaDelegateActivityBehavior extends TaskActivityBehavio
                 }
                 
                 throw e;
+            } catch (Throwable t) {
+                if (processEngineConfiguration.isLoggingSessionEnabled()) {
+                    BpmnLoggingSessionUtil.addErrorLoggingData(LoggingSessionConstants.TYPE_SERVICE_TASK_EXCEPTION,
+                            "Service task with java class " + javaDelegate.getClass().getName() + " threw throwable " + t.getMessage(), t, execution);
+                }
+
+                throw new RuntimeException(t);
             }
             
         } else {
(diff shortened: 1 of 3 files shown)


tijsrademakers:
Hi,

Thanks for the PR, but we should step back a bit first. When an out of memory error occurs, the current transaction to the database will not be committed and therefore the database state should be back to what it was before the out of memory occurred. When there is an async service task this could mean that no job is present anymore, because it failed 3 times already and the job moved to a deadletter job. Did you check the deadletter job table as well?

loic-seguin:
Unfortunately we did look in deadletterjob, in act_ru_job and in act_ru_timer_job, we have nothing.

My test is reproducing the problem normally.
If you remove the catch(Throwable t) in the code of this PR and rerun the test, you will never have a deadletterjob and if you put a breakpoint inside the call() method and wait a bit, you look for the activities instances and you will end up with an activity with no endTime and no job enywhere.

Are you sure the transaction is configured to be rollback as well if we get a Throwable ?

loic-seguin:
image

loic-seguin:
I am not able to access the database h2 in memory (I have nothing in the tables) so I created a postgre container in local and I launched the tests on it (we are also using a postgres in production). 
I confirm that the transaction is committed with no deadletterjob, no act_ru_job, no act_ru_timer_job.

tijsrademakers:
I just checked your unit test, this is not an async service task, but a regular sync service task. So what happens if the service task throws an exception is that the whole start of the process instance transaction is rolled back and therefore the database doesn't contain any information about the process instance. You would have to add an async flag on the service task to make this work.

loic-seguin:
I pushed again with async true on the service task and I remove the timer from xml so there is only the start event, the service task and the end event.
I am still reproducing the problem.

loic-seguin:
I have checked that the processInstance is not null in my junit to make sure that the process instance is well started.
Do you have any insight about this problem ?

loic-seguin:
@tijsrademakers Do you have an update concerning this issue ? I have updated what you said. I still reproduce the problem with my unit test.

tijsrademakers:
Hi @loic-seguin ,
Yes we will we do some final checking today, it looks like we will catch Throwable instead of Exception in the CommandContextInterceptor to solve the issue. Will update this PR when we have finalised the testing.
Thanks!

tijsrademakers:
Following up on the message earlier today. This has now been fixed with this commit [dafa95139e609de32e8f57181b58dd4c63119937](https://github.com/flowable/flowable-engine/commit/dafa95139e609de32e8f57181b58dd4c63119937).
Will close this PR now and merge the commit also to the 6.x branches.

Issue #3754, opened 2023-09-27
Problem while selecting the correct error event subprocess


marcberger:
**Describe the bug**
I have a BPMN with a subprocess containing another subprocess.
In both subprocesses, i can throw BPMN Errors.
I added, in each subprocess, an error event subprocesses to handle the BPMNError occuring inside.

When the engine search for an appropriate error subprocess to handle a BPMNError, the selected one is not always the same.

I put a full explanation of the problem on the flowable forum : 
https://forum.flowable.org/t/bpmnerror-error-handling-subprocess-possible-bug-at-least-strange-behaviour/10873

**Expected behavior**
If i have a BPMN containing subprocess1 containing subprocess2 i expect that : 

-  A BPMNError happening in a task of subprocess1 will be handled by the error event subprocess of subprocess1.
-  A BPMNError happening in a task of subprocess2 will be handled by the error event subprocess of subprocess2.

**Code**
[extract of the file where the error happens: ](https://github.com/flowable/flowable-engine/blob/6c5b934c0b0fa8b33a3612a11f503c06b81a369a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/helper/ErrorPropagation.java#L155C1-L175C18)

**Additional context**
flowable 7.0.0

PR #3756, opened 2023-09-27
EventSubProcess for error treatment should not be searched in children

marcberger:
#3754 : this patch will prevent searching within children subprocesses when searching for the correct event processing subprocesses


modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/helper/ErrorPropagation.java
```diff
@@ -160,7 +160,7 @@ protected static void executeCatch(Map<String, List<Event>> eventMap, DelegateEx
                             String refActivityId = refId.substring(0, refId.indexOf('#'));
                             String refProcessDefinitionId = refId.substring(refId.indexOf('#') + 1);
                             if (parentExecution.getProcessDefinitionId().equals(refProcessDefinitionId) &&
-                                            currentContainer.getFlowElement(refActivityId) != null) {
+                                            currentContainer.getFlowElement(refActivityId,false) != null) {
```
dbmalkovsky, line 163:
```suggestion
                                            currentContainer.getFlowElement(refActivityId, false) != null) {
```
Project standards is to include a space between parameters

marcberger:
Thanks for thé review.
It is m'y first contribution hère

PR #3757, opened 2023-09-29
The JaxB binding file of flowable-cxf must be updated to use Jakarta EE 9/10

cdeneux:


filiphr:
Thanks for the PR @cdeneux. Do you by any chance have an idea why the tests were passing prior to your change? I

cdeneux:
Hi @filiphr

The file 'flowable-binding.xjc' is used by flowable CXF to define the XML/Java mapping of datetime types.

I had a compilation error building my project in which I generate source code from XSD/WSDL files using the binding file 'flowable-binding.xjc' extracted from flowable-cxf. The source code generation is done using the Maven plugin 'org.patrodyne.jvnet:hisrc-higherjaxb-maven-plugin' that generate files with explicit imports to Jakarta packages if the JaxB version is defined to '3.0', and explicit imports to Javax packages if JaxB version is set to '2.0'
Previously, with Flowable 6.x, the source code generation was done with the Maven plugin 'org.jvnet.jaxb2.maven2:maven-jaxb2-plugin' that generate files with explicit imports to Javax packages.

I think that unit tests of flowable-cxf succeeded because flowable-cxf does not contain JaxB source code generation. We could add source code generation to highlight this package name changes through a compilation error ?

filiphr:
Thanks for the explanation @cdeneux, I understand now. Since there are different plugins, I don't think that we need to add something for this. Your changes are enough.

Just merged this

PR #3787, opened 2023-11-28
Support manual event subscription based process event-registry start

mkiener:
Add support for dynamic, manually created event subscriptions with any combination of correlation parameter values for a event-registry based process start.

#### Check List:
* Unit tests: YES
* Documentation: NO


filiphr (commented):
Thanks @mkiener, as we've discussed here are the comments that we should look into

PR #3789, opened 2023-11-29
Chore: expressions with type arguments can be replaced a with <> (JDK 9+)

dbmalkovsky:
Ran `mvn clean test` successfully.

### DIFF
--- a/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/test/impl/CmmnTestRunner.java
+++ b/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/test/impl/CmmnTestRunner.java
@@ -207,16 +207,16 @@ public class CmmnTestRunner extends BlockJUnit4ClassRunner {
             cmmnEngineConfiguration,
             TABLENAMES_EXCLUDED_FROM_DB_CLEAN_CHECK,
             true,
-                new Command<>() {
+            new Command<>() {
 
-                    @Override
-                    public Void execute(CommandContext commandContext) {
-                        SchemaManager schemaManager = CommandContextUtil.getCmmnEngineConfiguration(commandContext).getSchemaManager();
-                        schemaManager.schemaDrop();
-                        schemaManager.schemaCreate();
-                        return null;
-                    }
+                @Override
+                public Void execute(CommandContext commandContext) {
+                    SchemaManager schemaManager = CommandContextUtil.getCmmnEngineConfiguration(commandContext).getSchemaManager();
+                    schemaManager.schemaDrop();
+                    schemaManager.schemaCreate();
+                    return null;
                 }
+            }
 
         );
     }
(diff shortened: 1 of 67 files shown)


modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/test/impl/CmmnTestRunner.java
```diff
@@ -207,16 +207,16 @@ protected void assertDatabaseEmpty(FrameworkMethod method) {
...
             true,
-            new Command<Void>() {
-
-                @Override
-                public Void execute(CommandContext commandContext) {
-                    SchemaManager schemaManager = CommandContextUtil.getCmmnEngineConfiguration(commandContext).getSchemaManager();
-                    schemaManager.schemaDrop();
-                    schemaManager.schemaCreate();
-                    return null;
+                new Command<>() {
```
filiphr, line 210: This looks like the formatting was changed as well. Is that mandatory?

modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/persistence/deploy/DefaultDeploymentCache.java
```diff
@@ -41,21 +41,23 @@ public DefaultDeploymentCache() {
...
-                boolean removeEldest = size() > limit;
-                if (removeEldest && LOGGER.isTraceEnabled()) {
-                    LOGGER.trace("Cache limit is reached, {} will be evicted", eldest.getKey());
-                }
-                return removeEldest;
-            }
-
-        });
+        this.cache = Collections.synchronizedMap(
+                new LinkedHashMap<>(limit + 1, 0.75f, true) { // +1 is needed, because the entry is inserted first, before it is removed
```
filiphr, line 45: Do we need to change the formatting here? Can we perhaps keep only the change only to the `<>`?

filiphr (commented):
Thanks @dbmalkovsky. Most of it looks good to me. There are some changes in formatting in few places and I've left a comment for those. Can you please have a look?

btw, this is not a JDK 7+ feature. This is the enhanced diamond operator for JDK  9+, where you do not need to define the parameter in abstract inner classes.

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/test/impl/CmmnTestRunner.java
dbmalkovsky: reverted to original foirmatting

modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/persistence/deploy/DefaultDeploymentCache.java
dbmalkovsky: reverted to original formatting

filiphr:
Thanks @dbmalkovsky

PR #3815, opened 2024-01-07
Initial Spring Boot AOT Native support

jbarrez:
This PR adds initial support for Flowable + Spring Boot using the Spring Boot AOT Native image engine. 
Using the aot.properties approach, the new processors will only kick in when building native images (they won't interfere with 'regular' projects). 

Current scope of classes/resources that are added to the native image build:

- Flowable classes
- Mybatis classes that rely on reflection
- Mybatis mapper XML files
- BPMN/CMMN/DMN resources

Original code by @joshlong and further continued in a pair programming hacking session.

### DIFF
--- a/modules/flowable-spring-boot/flowable-spring-boot-starters/flowable-spring-boot-autoconfigure/src/main/java/org/flowable/spring/aot/FlowableBeanFactoryInitializationAotProcessor.java
+++ b/modules/flowable-spring-boot/flowable-spring-boot-starters/flowable-spring-boot-autoconfigure/src/main/java/org/flowable/spring/aot/FlowableBeanFactoryInitializationAotProcessor.java
@@ -1,0 +1,250 @@
+/* Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


modules/flowable-spring-boot/flowable-spring-boot-starters/flowable-spring-boot-autoconfigure/src/main/java/org/flowable/spring/aot/FlowableBeanFactoryInitializationAotProcessor.java
```diff
@@ -0,0 +1,304 @@
...
+
+/**
+ * @author Josh Long
+ * @author Joram Barrez
+ */
+class FlowableBeanFactoryInitializationAotProcessor implements BeanFactoryInitializationAotProcessor {
+
+    private final PathMatchingResourcePatternResolver resolver = new PathMatchingResourcePatternResolver();
+
+    private final Logger log = LoggerFactory.getLogger(getClass());
```
dbmalkovsky, line 66: Nit pick comment; ignoring it will not hurt my feelings.

There are 31 occurrences of loggers that use `getClass()`  and are not `static`.

There are 415 that use the form: `public static final Logger LOGGER = LoggerFactory.getLogger(xxxx.class);`, where the variable is uppercase and `static` and explicitly mentions the class.   I would assume consistency would mean that new code should use this later format.

filiphr: I actually prefer the use of a final field for this. It allows to inherit the class and when logging to see the actual instance of the class that is doing the logging. I actually think that we should update our `public static final Logger` to `protected final Logger`

PR #3824, opened 2024-01-22
fix: add getHandlerTypes() for the different Job Queries

ikaakkola:
getHandlerTypes should be implemented by all the different job queries, not just the default JobQueryImpl

#### Check List:
* Unit tests: NA
* Documentation: NA


ikaakkola:
This is required to implement feature complete custom Job DataManagers , the MyBatis based DataManager implementations use "handlerTypes" directly, eg.

https://github.com/flowable/flowable-engine/blob/main/modules/flowable-job-service/src/main/resources/org/flowable/job/service/db/mapping/entity/DeadLetterJob.xml#L266-L271

But as there is no getter, a custom DataManager the available options of "DeadletterJobQueryImpl" would need to use reflection to access 'handlerTypes' field.

filiphr:
Makes sense, thanks @ikaakkola

PR #3825, opened 2024-01-22
feat: Reconfigurable DataManagers for some Entity types

ikaakkola:
Allow (re)configuration of the DataManagers of various EntityManager implementations, at runtime, via MutableEntityManager interface.

This change allows the implementation of custom DataManagers without having to reinitialize all EntittManagers or use reflection to set the new DataManager into each EntityManager instance. This matches the expectation that after calling one of the setDataManager methods, the given DataManager is really used, also by the existing entity manager.

#### Check List:
* Unit tests: NA
* Documentation: NA

### DIFF
--- a/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/persistence/entity/AbstractEntityManager.java
+++ b/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/persistence/entity/AbstractEntityManager.java
@@ -23,7 +23,7 @@ import org.flowable.common.engine.impl.persistence.entity.data.DataManager;
  * @author Filip Hrisafov
  */
 public abstract class AbstractEntityManager<EntityImpl extends Entity, DM extends DataManager<EntityImpl>>
-    implements EntityManager<EntityImpl> {
+    implements MutableEntityManager<EntityImpl, DM> {
 
     protected DM dataManager;
     protected String engineType;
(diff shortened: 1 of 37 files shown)


filiphr:
@ikaakkola why is it not possible to just set the data manager on the appropriate engine configuration?

e.g. when you want to configure a custom data manager you'll just set it on the `ProcessEngineConfigurationImpl` in `EngineConfigurator#beforeInit`

ikaakkola:
> @ikaakkola why is it not possible to just set the data manager on the appropriate engine configuration?
> 
> e.g. when you want to configure a custom data manager you'll just set it on the `ProcessEngineConfigurationImpl` in `EngineConfigurator#beforeInit`

Thank you. I was so stuck on not being able to override datamanagers in the entitymanagers I never realised I could do it at a separate time (though this does require some changes in my custom DataManager implementations as they have some inter-datamanager relations to accomplish what is normally done at the database level).

filiphr:
>(though this does require some changes in my custom DataManager implementations as they have some inter-datamanager relations to accomplish what is normally done at the database level).

Not sure how you are doing this in your module. However, you could get the other data managers through the engine configuration. We are doing that in some of our entity managers, e.g. in https://github.com/flowable/flowable-engine/blob/93e003fb990da9342474e6aa1ecac8a2d35475db/modules/flowable-engine/src/main/java/org/flowable/engine/impl/persistence/entity/ExecutionEntityManagerImpl.java#L1132

PR #3826, opened 2024-01-23
feat: ServiceConfigurator support for various service configurations

ikaakkola:
Add support for ServiceConfigurator to Variable, Task, IdentityLink, EntityLink, EventSubscription and Batch ServiceConfiguration to allow customisation at init time.

The implementation follows the logic of JobServiceConfiguration that already has support for ServiceConfigurator.

#### Check List:
* Unit tests: YES
* Documentation: NA

### DIFF
--- a/modules/flowable-batch-service/src/main/java/org/flowable/batch/service/BatchServiceConfiguration.java
+++ b/modules/flowable-batch-service/src/main/java/org/flowable/batch/service/BatchServiceConfiguration.java
@@ -29,7 +29,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 /**
  * @author Tijs Rademakers
  */
-public class BatchServiceConfiguration extends AbstractServiceConfiguration {
+public class BatchServiceConfiguration extends AbstractServiceConfiguration<BatchServiceConfiguration> {
 
     // SERVICES
     // /////////////////////////////////////////////////////////////////
@@ -52,12 +52,21 @@ public class BatchServiceConfiguration extends AbstractServiceConfiguration {
         super(engineName);
     }
 
+    @Override
+    protected BatchServiceConfiguration getService() {
+        return this;
+    }
+
     // init
     // /////////////////////////////////////////////////////////////////////
 
     public void init() {
+        configuratorsBeforeInit();
+        
         initDataManagers();
         initEntityManagers();
+
+        configuratorsAfterInit();
     }
 
     // Data managers
@@ -96,7 +105,7 @@ public class BatchServiceConfiguration extends AbstractServiceConfiguration {
         this.batchService = batchService;
         return this;
     }
-    
+
     public BatchDataManager getBatchDataManager() {
         return batchDataManager;
     }
--- a/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/AbstractServiceConfiguration.java
+++ b/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/AbstractServiceConfiguration.java
@@ -12,8 +12,11 @@
  */
 package org.flowable.common.engine.impl;
 
+import java.util.Collection;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeSet;
 
 import org.flowable.common.engine.api.delegate.event.FlowableEventDispatcher;
 import org.flowable.common.engine.api.delegate.event.FlowableEventListener;
@@ -29,31 +32,36 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 /**
  * @author Tijs Rademakers
  */
-public abstract class AbstractServiceConfiguration {
-    
+public abstract class AbstractServiceConfiguration<S> {
+
     protected final Logger logger = LoggerFactory.getLogger(AbstractServiceConfiguration.class);
-    
+
     /** The tenant id indicating 'no tenant' */
     public static final String NO_TENANT_ID = "";
 
     protected String engineName;
+
+    protected Collection<ServiceConfigurator<S>> configurators;
+
     protected boolean enableEventDispatcher = true;
     protected FlowableEventDispatcher eventDispatcher;
     protected List<FlowableEventListener> eventListeners;
     protected Map<String, List<FlowableEventListener>> typedEventListeners;
     protected List<EventDispatchAction> additionalEventDispatchActions;
-    
+
     protected HistoryLevel historyLevel;
-    
+
     protected ObjectMapper objectMapper;
 
     protected Clock clock;
     protected IdGenerator idGenerator;
-    
+
     public AbstractServiceConfiguration(String engineName) {
         this.engineName = engineName;
     }
-    
+
+    protected abstract S getService();
+
     public boolean isHistoryLevelAtLeast(HistoryLevel level) {
         if (logger.isDebugEnabled()) {
             logger.debug("Current history level: {}, level required: {}", historyLevel, level);
@@ -76,7 +84,53 @@ public abstract class AbstractServiceConfiguration {
     public void setEngineName(String engineName) {
         this.engineName = engineName;
     }
-    
+
+    public Collection<ServiceConfigurator<S>> getConfigurators() {
+        return configurators;
+    }
+
+    public void setConfigurators(Collection<ServiceConfigurator<S>> configurators) {
+        initConfigurators();
+        this.configurators.clear();
+        if (configurators != null) {
+            this.configurators.addAll(configurators);
+        }
+    }
+
+    public AbstractServiceConfiguration<S> addConfigurator(ServiceConfigurator<S> configurator) {
+        initConfigurators();
+        this.configurators.add(configurator);
+        return this;
+    }
+
+    protected void initConfigurators() {
+        if (this.configurators == null) {
+            this.configurators = new TreeSet<>(Comparator.comparingInt(ServiceConfigurator::getPriority));
+        }
+    }
+
+    protected void configuratorsBeforeInit() {
+        if (this.configurators == null || this.configurators.isEmpty()) {
+            return;
+        }
+        final S service = getService();
+        this.configurators.stream().forEach(c -> {
+            logger.info("Executing beforeInit() of {} (priority: {})", c.getClass(), c.getPriority());
+            c.beforeInit(service);
+        });
+    }
+
+    protected void configuratorsAfterInit() {
+        if (this.configurators == null || this.configurators.isEmpty()) {
+            return;
+        }
+        final S service = getService();
+        this.configurators.stream().forEach(c -> {
+            logger.info("Executing afterInit() of {} (priority: {})", c.getClass(), c.getPriority());
+            c.afterInit(service);
+        });
+    }
+
     public boolean isEventDispatcherEnabled() {
         return getEventDispatcher() != null && getEventDispatcher().isEnabled();
     }
@@ -85,7 +139,7 @@ public abstract class AbstractServiceConfiguration {
         return enableEventDispatcher;
     }
 
-    public AbstractServiceConfiguration setEnableEventDispatcher(boolean enableEventDispatcher) {
+    public AbstractServiceConfiguration<S> setEnableEventDispatcher(boolean enableEventDispatcher) {
         this.enableEventDispatcher = enableEventDispatcher;
         return this;
     }
@@ -94,7 +148,7 @@ public abstract class AbstractServiceConfiguration {
         return eventDispatcher;
     }
 
-    public AbstractServiceConfiguration setEventDispatcher(FlowableEventDispatcher eventDispatcher) {
+    public AbstractServiceConfiguration<S> setEventDispatcher(FlowableEventDispatcher eventDispatcher) {
         this.eventDispatcher = eventDispatcher;
         return this;
     }
@@ -103,7 +157,7 @@ public abstract class AbstractServiceConfiguration {
         return eventListeners;
     }
 
-    public AbstractServiceConfiguration setEventListeners(List<FlowableEventListener> eventListeners) {
+    public AbstractServiceConfiguration<S> setEventListeners(List<FlowableEventListener> eventListeners) {
         this.eventListeners = eventListeners;
         return this;
     }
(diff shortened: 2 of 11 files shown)


ikaakkola:
To add more detail to why this is required, taking "VariableServiceConfiguration" as an example, the exact same applies to all the ServiceConfigurations improved with this PR.

VariableServiceConfiguration is initialized here https://github.com/flowable/flowable-engine/blob/main/modules/flowable-engine/src/main/java/org/flowable/engine/impl/cfg/ProcessEngineConfigurationImpl.java#L981 which ends up calling https://github.com/flowable/flowable-engine/blob/main/modules/flowable-variable-service/src/main/java/org/flowable/variable/service/VariableServiceConfiguration.java#L78-L82 - after "configuratorsBeforeInit()" , so if one tries to access VariableServiceConfiguration in beforeInit to set the DataManager, it will fail as it has not yet been instantiated.

The init() method of VariableServiceConfiguration initialises DataManagers and EntityManagers without any way of setting a custom DataManager before the EntityManager is also initialized. 
Thus, to have a custom DataManager one also needs to setVariableInstanceEntityManager(manager) with a new entity manager instance , because dataManager within the default VariableInstanceEntityManagerImpl implementation is immutable (my previous approach at this made dataManager mutable, but that was not a clean approach).  

So to be able to initialize the standard VariableServiceConfiguration with the standard VariableInstanceEntityManagerImpl but with a custom DataManager, one needs a way to set DataManager inside VariableServiceConfiguration before the entity manager is initialized - that is accomplished with ServiceConfigurator , exactly as JobServiceConfiguration already does it.

I believe this is the cleanest and most maintainable approach to the problem especially because it simply mimics what has already been implemented for JobServiceConfiguration.

filiphr:
Thanks @ikaakkola. This one makes a lot of sense. It also looks good for merging. I am only contemplating if we can move something to the `AbstractServiceConfiguration` and avoid the duplication we have for

```
 List<ServiceConfigurator<BatchServiceConfiguration>> configurators;
        if (this.configurators != null) {
            configurators = new ArrayList<>(this.configurators);
            configurators.sort(Comparator.comparingInt(ServiceConfigurator::getPriority));
        } else {
            configurators = Collections.emptyList();
        }

        for (ServiceConfigurator<BatchServiceConfiguration> configurator : configurators) {
            configurator.beforeInit(this);
        }
```

Will talk with the team and merge this soon

ikaakkola:
I agree the configurator logic is very repetitive, it could probably be generalized in the abstract class.

filiphr:
@ikaakkola I talked with the team and we think that it would be better if we can move this repetitive bits in the `AbstractEngineConfigurator`. I made a small spike in 9cbde32fe389f7abd8ebc40a30af1375dbae3a31 to see how it can look like. Feel free to use it as an inspiration.

ikaakkola:
@filiphr added generic ServiceConfigurator support for AbstractServiceConfiguration and implemented it in the concrete classes. 

I also considered moving the before and after init calls to a AbstractServiceConfiguration (so one would not forget to call the configuratorsBefore/AfterInit), something like:

```
    protected void init(Consumer<S> serviceInit) {
        configuratorsBeforeInit();
        serviceInit.accept(getService());
        configuratorsAfterInit();
    }
```

and in concrete ServiceConfiguration class call like:

```
    public void init() {
        init( s -> {
            s.initDataManagers();
            s.initEntityManagers();
        });
    }
```

but this doesn't seem to be a pattern used in the code base, so opted not to introduce it for now.

filiphr:
>I also considered moving the before and after init calls to a AbstractServiceConfiguration (so one would not forget to call the configuratorsBefore/AfterInit), something like:

That looks interesting. However, we are indeed not using something like that in the codebase. Perhaps at some point down the line. I'll go through your changes now

modules/flowable-batch-service/src/main/java/org/flowable/batch/service/BatchServiceConfiguration.java
```diff
@@ -23,18 +25,20 @@
...
  * @author Tijs Rademakers
  */
-public class BatchServiceConfiguration extends AbstractServiceConfiguration {
+public class BatchServiceConfiguration extends AbstractServiceConfiguration<BatchServiceConfiguration> {
 
     // SERVICES
     // /////////////////////////////////////////////////////////////////
 
     protected BatchService batchService = new BatchServiceImpl(this);
+    protected Collection<ServiceConfigurator<BatchServiceConfiguration>> configurators;
```
filiphr, line 41: Isn't this obsolete?

modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/AbstractServiceConfiguration.java
```diff
@@ -29,31 +33,36 @@
...
+
     protected final Logger logger = LoggerFactory.getLogger(AbstractServiceConfiguration.class);
-    
+
     /** The tenant id indicating 'no tenant' */
     public static final String NO_TENANT_ID = "";
 
     protected String engineName;
+
+    protected Set<ServiceConfigurator<S>> configurators;
```
filiphr, line 45: Why not `Collection`?

modules/flowable-engine-common/src/test/java/org/flowable/common/engine/impl/ServiceConfiguratorTest.java
```diff
@@ -0,0 +1,140 @@
+package org.flowable.common.engine.impl;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
```
filiphr, line 4: We try to use AssertJ for our assertions. Can we perhaps adapt this?

modules/flowable-entitylink-service/src/main/java/org/flowable/entitylink/service/EntityLinkServiceConfiguration.java
```diff
@@ -32,40 +35,50 @@
...
  */
-public class EntityLinkServiceConfiguration extends AbstractServiceConfiguration {
+public class EntityLinkServiceConfiguration extends AbstractServiceConfiguration<EntityLinkServiceConfiguration> {
 
     // SERVICES
     // /////////////////////////////////////////////////////////////////
 
     protected EntityLinkService entityLinkService = new EntityLinkServiceImpl(this);
     protected HistoricEntityLinkService historicEntityLinkService = new HistoricEntityLinkServiceImpl(this);
+    protected Collection<ServiceConfigurator<EntityLinkServiceConfiguration>> configurators;
```
filiphr, line 45: Isn't this obsolete?

modules/flowable-eventsubscription-service/src/main/java/org/flowable/eventsubscription/service/EventSubscriptionServiceConfiguration.java
```diff
@@ -27,21 +29,22 @@
...
  */
-public class EventSubscriptionServiceConfiguration extends AbstractServiceConfiguration {
+public class EventSubscriptionServiceConfiguration extends AbstractServiceConfiguration<EventSubscriptionServiceConfiguration> {
 
     // SERVICES
     // /////////////////////////////////////////////////////////////////
 
     protected EventSubscriptionService eventSubscriptionService = new EventSubscriptionServiceImpl(this);
-   
+    protected Collection<ServiceConfigurator<EventSubscriptionServiceConfiguration>> configurators;
```
filiphr, line 38: Isn't this obsolete?

modules/flowable-identitylink-service/src/main/java/org/flowable/identitylink/service/IdentityLinkServiceConfiguration.java
```diff
@@ -30,13 +33,14 @@
...
  */
-public class IdentityLinkServiceConfiguration extends AbstractServiceConfiguration {
+public class IdentityLinkServiceConfiguration extends AbstractServiceConfiguration<IdentityLinkServiceConfiguration> {
 
     // SERVICES
     // /////////////////////////////////////////////////////////////////
 
     protected IdentityLinkService identityLinkService = new IdentityLinkServiceImpl(this);
     protected HistoricIdentityLinkService historicIdentityLinkService = new HistoricIdentityLinkServiceImpl(this);
+    protected Collection<ServiceConfigurator<IdentityLinkServiceConfiguration>> configurators;
```
filiphr, line 43: Isn't this obsolete?

modules/flowable-job-service/src/main/java/org/flowable/job/service/JobServiceConfiguration.java
filiphr, line 1: I think that we should remove the `configurators` field from the `JobServiceConfiguration`

modules/flowable-task-service/src/main/java/org/flowable/task/service/TaskServiceConfiguration.java
```diff
@@ -46,8 +48,9 @@ public class TaskServiceConfiguration extends AbstractServiceConfiguration {
 
     protected TaskService taskService = new TaskServiceImpl(this);
     protected HistoricTaskService historicTaskService = new HistoricTaskServiceImpl(this);
-    
+
     protected IdmIdentityService idmIdentityService;
+    protected Collection<ServiceConfigurator<TaskServiceConfiguration>> configurators;
```
filiphr, line 53: Isn't this obsolete?

modules/flowable-variable-service/src/main/java/org/flowable/variable/service/VariableServiceConfiguration.java
```diff
@@ -42,14 +45,15 @@ public class VariableServiceConfiguration extends AbstractServiceConfiguration {
 
     protected VariableService variableService = new VariableServiceImpl(this);
     protected HistoricVariableService historicVariableService = new HistoricVariableServiceImpl(this);
+    protected Collection<ServiceConfigurator<VariableServiceConfiguration>> configurators;
```
filiphr, line 48: Isn't this obsolete?

modules/flowable-variable-service/src/main/java/org/flowable/variable/service/VariableServiceConfiguration.java
```diff
@@ -141,15 +154,15 @@ public VariableServiceConfiguration setVariableInstanceDataManager(VariableInsta
...
         return historicVariableInstanceDataManager;
     }
 
     public VariableServiceConfiguration setHistoricVariableInstanceDataManager(HistoricVariableInstanceDataManager historicVariableInstanceDataManager) {
-        this.historicVariableInstanceDataManager = historicVariableInstanceDataManager;
-        return this;
-    }
+      this.historicVariableInstanceDataManager = historicVariableInstanceDataManager;
+      return this;
+  }
```
filiphr, line 165: The formatting on this looks a bit off

modules/flowable-variable-service/src/main/java/org/flowable/variable/service/VariableServiceConfiguration.java
```diff
@@ -159,25 +172,25 @@ public VariableServiceConfiguration setVariableInstanceEntityManager(VariableIns
...
 
     public VariableServiceConfiguration setHistoricVariableInstanceEntityManager(HistoricVariableInstanceEntityManager historicVariableInstanceEntityManager) {
-        this.historicVariableInstanceEntityManager = historicVariableInstanceEntityManager;
-        return this;
-    }
-    
+      this.historicVariableInstanceEntityManager = historicVariableInstanceEntityManager;
+      return this;
+  }
+
```
filiphr, line 184: The formatting on this looks a bit off

filiphr (changes requested):
Thanks @ikaakkola. Looks good. I've left some comments, some small cleanups before we merge this. 

Let me know if you don't agree with something.

modules/flowable-engine-common/src/test/java/org/flowable/common/engine/impl/ServiceConfiguratorTest.java
ikaakkola: Certainly.

modules/flowable-variable-service/src/main/java/org/flowable/variable/service/VariableServiceConfiguration.java
ikaakkola: Wonder what happened here, Eclipse formatter is usually spot on. Fixed.

modules/flowable-job-service/src/main/java/org/flowable/job/service/JobServiceConfiguration.java
```diff
@@ -524,9 +505,9 @@ public List<AsyncRunnableExecutionExceptionHandler> getAsyncRunnableExecutionExc
     }
 
     public JobServiceConfiguration setAsyncRunnableExecutionExceptionHandlers(List<AsyncRunnableExecutionExceptionHandler> asyncRunnableExecutionExceptionHandlers) {
-        this.asyncRunnableExecutionExceptionHandlers = asyncRunnableExecutionExceptionHandlers;
-        return this;
-    }
+      this.asyncRunnableExecutionExceptionHandlers = asyncRunnableExecutionExceptionHandlers;
+      return this;
+  }
```
filiphr, line 510: I think I missed this formatting change earlier. 

```suggestion
        this.asyncRunnableExecutionExceptionHandlers = asyncRunnableExecutionExceptionHandlers;
        return this;
    }
```

filiphr (changes requested):
@ikaakkola I've missed one place for the formatting and the `JobServiceConfiguration` still has the `configurators` field, I think we should remove that as well.

modules/flowable-job-service/src/main/java/org/flowable/job/service/JobServiceConfiguration.java
ikaakkola: these are really weird, I wonder what my Eclipse is doing, sorry for the hassle..

modules/flowable-job-service/src/main/java/org/flowable/job/service/JobServiceConfiguration.java
ikaakkola: removed.

modules/flowable-job-service/src/main/java/org/flowable/job/service/JobServiceConfiguration.java
filiphr: all good. We have some code style in https://github.com/flowable/flowable-engine/tree/main/ide-settings. Separate for IntelliJ and Eclipse. However, I think that the one for Eclipse might not be up to date 😄

filiphr (approved):
Thanks for all the changes and contribution @ikaakkola. Will merge this once the tests are green

PR #3832, opened 2024-01-31
Update spring

sflatley1:
#### Check List:
* Unit tests: YES
* Documentation:  NA

Updated:
spring boot -> 3.1.6, 
spring framework  -> 6.0.14
spring framework version -> spring security -> 6.1.5

These updates address CVE-2023-34053 and CVE-2023-34042.

Failing unit tests had me: 

- Update the bmpn20 file's delay as the associated timeout test was not timing out
- Address exceptions JPATestRestCase had getting a message from a test database.  The Message interface is defined in the test source, the database configuration is in the test source and the unit tests tested that a message was retrieved.  Why? I removed all that and have the base class returning a message.
- Update a test in AllEnginesAutoConfigurationTest - I removed the order constraint when checking if objects exists in a list.

All unit tests pass.


filiphr:
Thanks for the PR @sflatley1.

How does this PR resolve those 2 CVEs?

* [CVE-2023-34053](https://spring.io/security/cve-2023-34042) has been fixed in Spring Security 6.1.4+ and we use 6.1.5 
* [CVE-2023-34042](https://spring.io/blog/2023/11/27/cve-2023-34053-cve-2023-34055-spring-framework-and-spring-boot) which says it has been fixed in Spring Framework 6.0.14 and Boot 2.7.18, and we use Framework 6.0.14 and Boot 3.1

What is the need to add `spring-context` to some modules?

>Address exceptions JPATestRestCase had getting a message from a test database. The Message interface is defined in the test source, the database configuration is in the test source and the unit tests tested that a message was retrieved. Why? I removed all that and have the base class returning a message.

The test is meant to test JPA. So what you now did has nothing to do with JPA anymore.

> Update a test in AllEnginesAutoConfigurationTest - I removed the order constraint when checking if objects exists in a list.

The order there is actually important. What wasn't working there?

sflatley1:
Thanks so much for the quick notification.

 

How does this PR resolve those 2 CVEs?

I’m looking into this.  Our build server’s dependency tracker is showing this: 



The test is meant to test JPA. So what you now did has nothing to do with JPA anymore.

                This can be put back in.  I don’t usually write unit tests for code that is not in the src tree.

 

Update a test in AllEnginesAutoConfigurationTest - I removed the order constraint when checking if objects exists in a list

                It failed because the order was not correct. 

 

Sean

filiphr:
@sflatley1 I've worked on upgrading to Spring Boot 3.2 and have prepared PR #3846 for this. There it is explained why we needed the `spring-context` and the change for the order in the `AllEnginesAutoConfigurationTest` (I was initially wrong about it mattering). 

In any case I am closing now this in favor of PR #3846

Issue #3835, opened 2024-02-02
Fluent assertions


grofcik:
**Is your feature request related to a problem? Please describe.**

Assertions are not readable. Instead of:
```
@Test 
 void followHappyPath(RuntimeService runtimeService, TaskService taskService, HistoryService historyService) { 
     ProcessInstance helloWorldProcess = runtimeService.createProcessInstanceBuilder().processDefinitionKey("P001-helloWorld").start(); 
  
     assertThat(helloWorldProcess).isNotNull(); 
     List<String> path = new ArrayList<>(List.of("theStart", "theStart-theSayHelloUserTask", "theSayHelloUserTask")); 
     assertThat(runtimeService.createActivityInstanceQuery() 
             .processInstanceId(helloWorldProcess.getId()).orderByActivityInstanceStartTime().asc().list()) 
             .as("The hello world process has to go directly through theStart -> theSayHelloUserTask") 
             .extracting(ActivityInstance::getActivityId) 
             .containsExactlyInAnyOrderElementsOf(path); 
     assertThat(runtimeService.createVariableInstanceQuery().processInstanceId(helloWorldProcess.getId()).list()) 
             .extracting("name", "value") 
             .containsExactly(Tuple.tuple("initiator", null)); 
  
     Task userTask = taskService.createTaskQuery().processInstanceId(helloWorldProcess.getId()).singleResult(); 
  
     assertThat(userTask).extracting(Task::getName).isEqualTo("Say hello world"); 
     assertThat(userTask).extracting(Task::getTaskDefinitionKey).isEqualTo("theSayHelloUserTask"); 
  
     taskService.complete(userTask.getId(), Map.of("greeting", "Hello World!")); 
  
     assertThat(runtimeService.createActivityInstanceQuery().processInstanceId(helloWorldProcess.getId()).singleResult()) 
             .as("The hello world process is finished, no instance is present in the runtime.") 
             .isNull(); 
     HistoricProcessInstance historicProcessInstance = historyService.createHistoricProcessInstanceQuery().processInstanceId(helloWorldProcess.getId()).singleResult(); 
     assertThat(historicProcessInstance) 
             .as("The hello world process must be present in the history") 
             .isNotNull(); 
     assertThat(historicProcessInstance.getEndTime()) 
             .as("The hello world process must be finished") 
             .isNotNull(); 
     assertThat(historyService.createHistoricVariableInstanceQuery().processInstanceId(helloWorldProcess.getId()).list()) 
[...]

PR #3837, opened 2024-02-02
#3835 fluent assertions

martin-grofcik:
#### Check List:
* Unit tests: YES 
* Documentation: NA yet


modules/flowable-assertions/flowable-process-assertions/src/main/java/org/flowable/assertions/process/FlowableProcessAssertions.java
```diff
@@ -0,0 +1,19 @@
+package org.flowable.assertions.process;
```
dbmalkovsky, line 1: Missing copyright header.

modules/flowable-assertions/flowable-process-assertions/src/main/java/org/flowable/assertions/process/HistoricProcessInstanceAssert.java
```diff
@@ -0,0 +1,150 @@
...
+import org.assertj.core.api.ListAssert;
+import org.flowable.engine.ProcessEngine;
+import org.flowable.engine.history.HistoricActivityInstance;
+import org.flowable.engine.history.HistoricProcessInstance;
+import org.flowable.identitylink.api.IdentityLink;
+import org.flowable.identitylink.api.history.HistoricIdentityLink;
+import org.flowable.task.api.history.HistoricTaskInstance;
+import org.flowable.variable.api.history.HistoricVariableInstance;
+
+import static org.flowable.assertions.process.FlowableProcessAssertions.assertThat;
```
dbmalkovsky, line 14: static imports by project standards go first in the import order.   Also missing license header here and other classes.

modules/flowable-assertions/flowable-process-assertions/src/main/java/org/flowable/assertions/process/HistoricProcessInstanceAssert.java
```diff
@@ -0,0 +1,150 @@
...
+import static org.flowable.assertions.process.Utils.*;
```
dbmalkovsky, line 15: Project standards do not allow for package level imports; they should be enumerated individually.

modules/flowable-assertions/flowable-process-assertions/src/main/java/org/flowable/assertions/process/HistoricProcessInstanceAssert.java
```diff
@@ -0,0 +1,150 @@
...
+    /**
+     * Assert <b>historic</b> process instance exists in the history and is finished.
+     *
+     * @return Historic process instance assertion.
+     */
+    public HistoricProcessInstanceAssert isFinished() {
+        processExistsInHistory();
+
+        if (processServicesProvider.getHistoryService().createHistoricProcessInstanceQuery().finished().processInstanceId(actual.getId()).count() != 1) {
+            failWithMessage(getProcessDescription(actual)+" to be finished, but is running in history.");
```
dbmalkovsky, line 55:
```suggestion
            failWithMessage(getProcessDescription(actual) + " to be finished, but is running in history.");
```

modules/flowable-assertions/flowable-process-assertions/src/main/java/org/flowable/assertions/process/HistoricProcessInstanceAssert.java
```diff
@@ -0,0 +1,150 @@
...
+     *
+     * @param variableName variable to check.
+     * @return Historic process instance assertion
+     */
+
+    public HistoricProcessInstanceAssert hasVariable(String variableName) {
+        processExistsInHistory();
+
+        if (processServicesProvider.getHistoryService().createHistoricProcessInstanceQuery().processInstanceId(actual.getId()).variableExists(variableName).count() != 1) {
+            failWithMessage(getProcessDescription(actual)+" has variable <%s> but variable does not exist in history.", variableName);
```
dbmalkovsky, line 78:
```suggestion
            failWithMessage(getProcessDescription(actual) + " has variable <%s> but variable does not exist in history.", variableName);
```

There are other examples of the same spacing issue.

modules/flowable-assertions/flowable-process-assertions/src/main/java/org/flowable/assertions/process/ProcessInstanceAssert.java
```diff
@@ -0,0 +1,182 @@
...
+
+        if (processServicesProvider.getRuntimeService().createProcessInstanceQuery().processInstanceId(actual.getId()).variableExists(variableName).count() != 0) {
+            failWithMessage(getProcessDescription(actual)+" does not have variable <%s> but variable exists.", variableName);
+        }
+
+        return this;
+    }
+
+    /**
+     * Assert that process instance does on exist in <b>runtime</b>.
```
dbmalkovsky, line 94:
```suggestion
     * Assert that process instance does not exist in <b>runtime</b>.
```

modules/flowable-assertions/flowable-process-assertions/src/main/java/org/flowable/assertions/process/HistoricProcessInstanceAssert.java
martin-grofcik: resolved

modules/flowable-assertions/flowable-process-assertions/src/main/java/org/flowable/assertions/process/HistoricProcessInstanceAssert.java
martin-grofcik: resolved

filiphr (commented):
This is an interesting idea @martin-grofcik. It allows for some good testing of processes / cases. I have some questions / ideas.

* In my opinion I think that a single `flowable-assertj` module where we would add optional support for different modules is more correct. Users don't need to care which module for asserting they want to bring. They can just add `flowable-assertj` and use whatever they want from there. We would need to have different entry points, but that is fine.
* I see that in the asserts we are doing DB calls all the time. What is the purpose of that? e.g. for `hasVariableWithValue` we first do `hasVariable` which will do a DB call and then we do another DB call. I think that it would be better if we treat the instance as is and don't do additional DB calls if it isn't needed.

martin-grofcik:
Hi @filiphr.

re: single `flowable-assertj` I will change the structure.

re: DB calls.
What I would like to achieve is:

https://github.com/flowable/flowable-engine/blob/ee24b3c87b3432d86672fd8d6c8e9e239d0e7e27/modules/flowable-assertions/flowable-process-assertions/src/test/java/org/flowable/assertions/process/LongRunningProcessInstanceAssertTest.java#L35-L55

One `asserThatProcessInstance` evolves in the time, so there is no need to create a new instance. If we "cache" the instance status, it is not possible to achieve it.

filiphr:
@martin-grofcik my question is why would you like to cache the instance? Why not do something like

```java
 asserThatProcessInstance(processInstance.getId()).isRunning() 
         .activities().extracting(ActivityInstance::getActivityId).containsExactlyInAnyOrder( 
             "theStart", "theStart-theTask", "theTask1" 
         ); 
  
 taskService.complete(geTaskIdForProcessInstance(twoTasksProcess.getId(), taskService)); 
  
asserThatProcessInstance(processInstance.getId()).isRunning().activities().extracting(ActivityInstance::getActivityId).containsExactlyInAnyOrder( 
         "theStart", "theStart-theTask", "theTask1", "theTask1-theTask2", "theTask2" 
 ); 

 taskService.complete(geTaskIdForProcessInstance(twoTasksProcess.getId(), taskService)); 
  
asserThatProcessInstance(processInstance.getId()).doesNotExist().inHistory() 
         .activities().extracting(HistoricActivityInstance::getActivityId).containsExactlyInAnyOrder( 
             "theStart", "theStart-theTask", "theTask1", "theTask1-theTask2", "theTask2", "theTask-theEnd", 
                 "theEnd" 
         ); 
```

Perhaps we can even have a `ProcessAssert` that you can instantiate and then use it via `processAssert.assertThatXXX`

martin-grofcik:
```java
assertThatProcessInstance(processInstance.getId()).isRunning()
``` 
is too verbose and non fluent. I would prefer 
```java
assertThat(processInstance).isRunning()...
```
 from the code reading perspective. `assertThat(processInstance)` still allows us to keep query result in the instance.

I do not know whether following adds the value:

```java
   var assertThatOneTaskProcess = assertThat(processInstance).isRunning();

  taskService.complete(oneTaskProcessTask.getId());

  assertThatOneTaskProcess.doesNotExist();
```
My initial impl was storing process instance state in the assertion instance. So I do not mind.

**Conclusion:** What do you prefer?

martin-grofcik:
Any update?

Issue #3841, opened 2024-02-07
Missing Groovy engine in OSGi environment


karelmaxa:
**Describe the bug**

Using Groovy in OSGi environment throws `FlowableException` (_Can't find scripting engine for 'groovy'_). For example, the exception will be thrown when you use the following execution listener:

```
<flowable:executionListener event="start" class="org.flowable.engine.impl.bpmn.listener.ScriptExecutionListener">
    <flowable:field name="script">
        <flowable:string>execution.setVariable('foo', 'bar')</flowable:string>
    </flowable:field>
    <flowable:field name="language">
        <flowable:string>groovy></flowable:string>
    </flowable:field>
</flowable:executionListener>
```

**Expected behavior**

Groovy language can be used in OSGi environment.

**Code**

The `Extender` component tries to read the factory class from the first line of the corresponding Groovy configuration file ([groovy-jsr223/javax.script.ScriptEngineFactory](https://github.com/apache/groovy/blob/359ee64eda6bdafde7fa5c53ccab38f5e89df493/subprojects/groovy-jsr223/src/main/resources/META-INF/services/javax.script.ScriptEngineFactory)), but it is the license header.

https://github.com/flowable/flowable-engine/blob/356c0084c2c24e52f18bcac88e9e828f4b69ccae/modules/flowable-osgi/src/main/java/org/flowable/osgi/Extender.java#L371

**Additional context**

I'm using the following components:

- flowable-engine-6.7.2
- flowable-osgi-6.7.2
- groovy-all-2.4.21

I will try to submit a pull request with the fix.

PR #3843, opened 2024-02-07
Fix parsing of service provider file (#3841).

karelmaxa:
This PR fixes wrong parsing of the service provider configuration file in the OSGi environment.

The specification of the configuration file is the following (see `ServiceLoader` JavaDoc):

_The provider-configuration file must be encoded in UTF-8. Space and tab characters surrounding each service provider's name, as well as blank lines, are ignored. The comment character is '#' (U+0023 NUMBER SIGN); on each line all characters following the first comment character are ignored_

#### Check List:
* Unit tests: YES
* Documentation: NO


modules/flowable-osgi/src/test/java/org/flowable/osgi/BundleScriptEngineResolverTest.java
```diff
@@ -0,0 +1,55 @@
+package org.flowable.osgi;
```
dbmalkovsky, line 1: Please include the standard license header file; thank you.

karelmaxa: Sure, the license header has been added.

filiphr (changes requested):
Thanks @karelmaxa. Everything looks great, only one small change

modules/flowable-osgi/src/main/java/org/flowable/osgi/Extender.java
```diff
@@ -367,9 +369,13 @@ public void unregister() {
         @Override
         public ScriptEngine resolveScriptEngine(String name) {
             try {
-                BufferedReader in = new BufferedReader(new InputStreamReader(configFile.openStream()));
-                String className = in.readLine();
-                in.close();
+                String className;
+                try (var input = configFile.openStream()) {
```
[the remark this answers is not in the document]
karelmaxa: Sure, changed.

pavelhoral:
A bit off topic (explanation for the crashing tests):

The build crashes because of [`FlowableMyBatisRuntimeHints`](https://github.com/flowable/flowable-engine/blob/f0864bced3b058ca9120aebe87764cd9510d685e/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/aot/FlowableMyBatisRuntimeHints.java) uses non-exported shaded (internal) Javassist class from mybatis dependency. Then the class [`FlowableAppRuntimeHints`](https://github.com/flowable/flowable-engine/blob/f0864bced3b058ca9120aebe87764cd9510d685e/modules/flowable-app-engine/src/main/java/org/flowable/app/engine/impl/aot/FlowableAppRuntimeHints.java) uses imports from spring-core dependency, which is [not defined as OSGi at all](https://www.infoq.com/news/2012/10/spring-osgi-gradle/). And there might be others (I did not dig deeper). I think that the current state of main branch is kind of totally broken from the OSGi point of view.

So this seems like it needs to be addressed at some architectural design level whether OSGi is a concern or not. I am just a "passer by", so I just wanted to share this knowledge without knowing what to do with this information.

filiphr:
Thanks for the explanation @pavelhoral. Those hints were moved just today. We'll look into fixing those OSGi tests. Seems like you know your way around OSGi, do you perhaps know if there is a way to exclude some classes? e.g. those runtime hints are not needed in an OSGi world.

pavelhoral:
I am just an OSGi user... but I can try to look into it later. Normally you [would define the packages](https://github.com/flowable/flowable-engine/blob/f0864bced3b058ca9120aebe87764cd9510d685e/modules/flowable-engine-common/pom.xml#L21) that should not be exported with an exclamation mark in front of it. But there is a conflict [with the configuration in the root POM file](https://github.com/flowable/flowable-engine/blob/f0864bced3b058ca9120aebe87764cd9510d685e/pom.xml#L76) that practically marks all module packages for export (so it can not be overridden).

karelmaxa:
I have rebased the branch with the master, so the OSGi test fix should now be included.

filiphr:
Thanks @karelmaxa. The rebase wasn't entirely needed due to the way GitHub does PR checks. The tests are run by applying your changes on top of the target branch (`main`) and then running the tests.

PR #3846, opened 2024-02-09
Spring Boot 3.2 and other dependencies upgrade

filiphr:
The following dependencies have been upgraded:

* Spring Boot 3.2.2
* Spring Framework 6.1.3
* Spring Security 6.2.1
* Spring AMQP 3.1.1
* Spring Kafka 3.1.1
* Spring LDAP 3.2.1
* Reactor Netty 1.1.15
* SLF4J 2.0.11
* Groovy 4.0.17
* JUnit Jupiter 5.10.1
* Mockito 5.7.0
* Byte Buddy 1.14.11
* Testcontainer 1.19.3
* Artemis 2.31.2
* Http Client5 5.2.3
* Hibernate Core 6.4.1.Final
* Spring Data JPA 3.2.2

The assertion in `AllEnginesAutoConfigurationTest` has been changed to use `containsExactlyInAnyOrder` since the order there isn't important, and it seems like the beans are created in a different order now.
The `AppEngineConfiguration` is going to anyways order all the available contributors before applying them. The configurators are only the ones that have been registered during bean registration.

Our Spring Kafka integration has been adapted to use the new `SameIntervalTopicReuseStrategy` instead of the deprecated (in 3.0) and removed (in 3.1) spring-kafka `FixedDelayStrategy`. It is the same thing, just named differently for clarification reasons by the Spring Kafka team.

We have also added `spring-context` as a test dependency in `flowable-http` , because the `SpringWebClientFlowableHttpClient` is using `ReactorClientHttpConnector` from `spring-web`, which since Spring Framework 6.1 is implementing `SmartLifecycle` and thus needs spring-context

### DIFF
--- a/distro/src/notice.txt
+++ b/distro/src/notice.txt
@@ -97,32 +97,32 @@ org.apache.httpcomponents       httpclient                  4.5.13          Apac
 org.apache.httpcomponents       httpcore                    4.4.15          Apache License, Version 2.0
 org.apache.httpcomponents       httpmime                    4.5.13          Apache License, Version 2.0
 org.apache.geronimo.bundles     json                        20090211_1      The Apache Software License, Version 2.0
-org.apache.groovy               groovy                      4.0.15          The Apache Software License, Version 2.0
-org.apache.groovy               groovy-jsr223               4.0.15          The Apache Software License, Version 2.0
+org.apache.groovy               groovy                      4.0.17          The Apache Software License, Version 2.0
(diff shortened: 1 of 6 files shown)


dbmalkovsky (approved):
Very nice work  @filiphr; LGTM and thank you.

filiphr:
Thanks @dbmalkovsky

sflatley1:
Thanks!

flavienlaurent:
Hello, any idea when it will be released? Thanks

PR #3852, opened 2024-02-27
Fix order by clause missing "order by"

el-hackerino:
The generated ORDER BY clause is incorrect, leading to syntax errors in the resulting query. The ELSE branch in lines 64-55 does it correctly.
I'm not set up to add/run tests for this, but I think this change should be looked at ASAP since the bug breaks a lot of queries.

#### Check List:
* Unit tests: NO
* Documentation: NO


el-hackerino:
Oops, wrong base branch..

PR #3853, opened 2024-02-27
Fix order by clause missing "order by"

el-hackerino:
The generated ORDER BY clause is incorrect, leading to syntax errors in the resulting query. The ELSE branch in lines 64-55 did it correctly (but I refactored the code to combine the code).

I'm not set up to add/run tests for this, but I think this change should be looked at ASAP since the bug breaks a lot of queries.

#### Check List:
* Unit tests: NO
* Documentation: NO


filiphr:
@el-hackerino can you please provide an example query where this is causing issues? This code has been like this since 2017 and we have not received any issues about it.

Perhaps you are using it incorrectly. How are you passing the `orderBy` parameter?

filiphr:
@el-hackerino can you please share the java code you are using for this?

filiphr:
@el-hackerino I've added a test case for this and I'll merge it once the tests pass.

PR #3861, opened 2024-03-07
Fix possible null pathInfo

marcberger:
Hello,

This patch is an attemps to fix error like that : 

java.lang.NullPointerException: Cannot invoke "String.replace(java.lang.CharSequence, java.lang.CharSequence)" because "pathInfo" is null
org.flowable.rest.service.api.repository.DeploymentResourceResource.getDeploymentResource(DeploymentResourceResource.java:87) 

caused by pathInfo beeing null.


tijsrademakers:
Hi, can you please add a failing unit test showing when an issue can happen? And also make sure that typos (ressourceURI) are fixed.

PR #3864, opened 2024-03-20
fix: Solution to the problem of parallel gateway aggregation blocking

wukaifang1024:
Use moveExecutionsToSingleActivityId to jump to the branch of the parallel gateway, execute the approval task, and when the process flows to the convergence of the parallel gateway, it is blocked and does not continue to flow to the next node.


tijsrademakers:
Hi, we have no idea yet what the background is of this PR. We at least require a unit test showing the issue without the code change. But also the code change doesn't look ok at this moment with doing an activity instance query with just the process instance id. So I'm closing this PR for now, please reopen / recreate it with unit tests and more explanation.

PR #3889, opened 2024-05-14
#3884 Expression language support for conditional sequence flow and conditional events

### LINKED ISSUE #3884: Conditions should support more than JUEL expressions
**Is your feature request related to a problem? Please describe.**

Conditions (and other places with `FormalExpressions`) currenly ignores the `language` attribute and always assumes that the content is JUEL. Some conditions are hard or impossible to write in JUEL (i.e. "is the country of one of the addresses in the US?"), so it would be great to be able to use i.e. Groovy instead.

**Describe the solution you'd like**

`SequenceFlow`s (and `BoundaryEvents` etc.) should parse the `language` attribute of the `bpmn:condition`. `ExpressionManager.createExpression()` should use the language to load the `ScriptEngine` using `javax.script.ScriptEngineManager`, probably only accepting languages from a configurable whitelist.

It looks like a fairly straightforward task, and I will be happy to supply a PR, if I can get an indication that it is likely to be merged.

rasmusfaber:
This fixes #3884 by adding support for other expression languages as specified by the `language` tag in the `FormalExpression`, i.e.

```
   <sequenceFlow id="flow">
      <conditionExpression xsi:type="tFormalExpression" language="groovy">
          customer.addresses.any { it.country == 'US' }
      </conditionExpression>
    </sequenceFlow>
```

#### Check List:
* Unit tests: YES
* Documentation: NA (I couldn't find anything at least)


modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/el/ScriptEngineExpression.java
```diff
@@ -0,0 +1,85 @@
...
+	public Object getValue(VariableContainer variableContainer) {
+		ScriptEngineRequest.Builder builder = ScriptEngineRequest.builder()
+				.script(expressionText)
+				.language(expressionLanguage)
+				.variableContainer(variableContainer);
+		return getScriptingEngines().evaluate(builder.build()).getResult();
+	}
+
+	private ScriptingEngines getScriptingEngines() {
+		CommandContext commandContext = Context.getCommandContext();
```
rasmusfaber, line 59: Perhaps you prefer a different way to get the dependency? But this looks like the way it is done elsewhere, except they mostly use `CommandContextUtil` (which is not accesible from flowable-engine-common).

filiphr:
@rasmusfaber thanks for the PR can you looked into `org.flowable.engine.impl.Condition`, `org.flowable.engine.impl.scripting.ScriptCondition` and `org.flowable.engine.impl.el.UelExpressionCondition`? I think that what you are looking for fits more into using `ScriptCondition` instead of the `UelExpressionCondition`.

With that approach we won't need the `ScriptEngineExpression` and we won't need to get the process engine configuration in the way we are doing it now.

rasmusfaber:
> @rasmusfaber thanks for the PR can you looked into `org.flowable.engine.impl.Condition`, `org.flowable.engine.impl.scripting.ScriptCondition` and `org.flowable.engine.impl.el.UelExpressionCondition`?

@filiphr Oh, yes. That was much cleaner. Thank you. I have updated the PR.

filiphr:
Thanks @rasmusfaber. 

I see now that you did some changes in the `ScriptCondition`. I would instead prefer to have a new method in the `ConditionUtil` and create a `Condition` from there. If the language is `null` or empty then we create the `UelExpressionCondition` otherwise we create a `ScriptCondition`. The `ScriptCondition` should not behave like a `UelExpressionCondition` if there is no language

rasmusfaber:
> I would instead prefer to have a new method in the `ConditionUtil` and create a `Condition` from there.

Sure. How is this? I also renamed the `sequenceFlowId` parameter to `elementId`, since it could now also be the id of an event.

filiphr:
Looks good, thanks @rasmusfaber

PR #3915, opened 2024-06-25
Fix word spel error

youxiaxiaomage:
#### Check List:
* Unit tests: NO
* Documentation: NO
* Fix word spel error


filiphr:
Hey @youxiaxiaomage, thanks for the change. However, unfortunately we cannot accept this PR. The reason for this is the backwards compatibility. There can already be jobs for that particular job handler that will fail to execute if we change the typo in the type name. Therefore, I'm going to close this PR

PR #3958, opened 2024-09-10
fixed throwing an exception when process contains two boundary events…

ywling417:
… and the first is compensation

#### Check List:
* Unit tests: YES 
* Documentation: YES

### DIFF
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/event/compensate/CompensateEventTest.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/event/compensate/CompensateEventTest.java
@@ -13,12 +13,6 @@
 
 package org.flowable.engine.test.bpmn.event.compensate;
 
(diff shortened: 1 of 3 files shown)


modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/event/compensate/CompensateEventTest.java
```diff
@@ -13,12 +13,6 @@
 
 package org.flowable.engine.test.bpmn.event.compensate;
 
-import static org.assertj.core.api.Assertions.assertThat;
```
dbmalkovsky, line 16: Please follow the project import ordering; see https://github.com/flowable/flowable-engine/blob/main/ide-settings/eclipse/Preferences-Java-CodeStyle-OrganizeImports.importorder

ywling417: I have already made the necessary changes

PR #3971, opened 2024-10-02
exposed local plan item instance variables via query and rest api

WelschChristopher:
#### Check List:
* Unit tests: YES / NO / NA

### DIFF
--- a/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/history/HistoricPlanItemInstanceQueryImpl.java
+++ b/modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/history/HistoricPlanItemInstanceQueryImpl.java
@@ -520,7 +520,7 @@ public class HistoricPlanItemInstanceQueryImpl extends AbstractQuery<HistoricPla
     @Override
     public List<HistoricPlanItemInstance> executeList(CommandContext commandContext) {
         List<HistoricPlanItemInstance> historicPlanItems;
-        if(includeLocalVariables){
+        if (includeLocalVariables){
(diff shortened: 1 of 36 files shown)


modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/history/HistoricPlanItemInstanceQueryImpl.java
```diff
@@ -511,7 +519,12 @@ public long executeCount(CommandContext commandContext) {
 
     @Override
     public List<HistoricPlanItemInstance> executeList(CommandContext commandContext) {
-        List<HistoricPlanItemInstance> historicPlanItems = CommandContextUtil.getHistoricPlanItemInstanceEntityManager(commandContext).findByCriteria(this);
+        List<HistoricPlanItemInstance> historicPlanItems;
+        if(includeLocalVariables){
```
dbmalkovsky, line 523:
```suggestion
        if (includeLocalVariables){
```

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/history/HistoricPlanItemInstanceQueryImpl.java
```diff
@@ -511,7 +519,12 @@ public long executeCount(CommandContext commandContext) {
...
+            historicPlanItems = CommandContextUtil.getHistoricPlanItemInstanceEntityManager(commandContext).findWithVariablesByCriteria(this);
+        }else{
```
dbmalkovsky, line 525:
```suggestion
        } else {
```

PR #3973, opened 2024-10-04
fix a signal creation bug

robert-gurol-ebcont:
Reopening https://github.com/flowable/flowable-engine/pull/3920

@tijsrademakers I _**specifically**_ asked for pointers in that pull request about where to add a unit test for this - I don't see through the big overall structure of the repository. I want to add unit tests, but I need your help to tell me what I can build on.


tijsrademakers:
Hi, I think a good place to look for is here https://github.com/flowable/flowable-engine/tree/main/modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/event/signal. You can add it to the existing SignalEventTest unit test for example. Let me know if you need more info.

robert-gurol-ebcont:
Thank you! I'll have a look :heart: 

PS: The existing tests still run with my change applied :see_no_evil:

tijsrademakers:
That's good that this test still runs. But it would be great if there could be a test added that's failing without your fix if possible.

PR #4003, opened 2024-12-19
Chore: remove unused import

dbmalkovsky:
Not too bad; the only unused import in about the last 45 days


tijsrademakers:
It's good to see that we are improving there :-), thanks

PR #4010, opened 2025-01-09
fix XXE vulnerability in Flowable release 6.8.1

dengjfeng:
#### Check List:
* Unit tests: NA
* Documentation: NA


tijsrademakers:
Hi, as you can see here we already do this on the xml converter level. This is the link for BPMN, but we have similar code for CMMN:

https://github.com/flowable/flowable-engine/blob/706e10ffbe6ef6e79409ea003186d5806a50cbc4/modules/flowable-bpmn-converter/src/main/java/org/flowable/bpmn/converter/BpmnXMLConverter.java#L271

PR #4034, opened 2025-03-10
Fix potential NullPointerException in addAuthorizationsForNewProcessD…

yannaingtun:
**Description:**
This PR adds a null check for the process parameter in addAuthorizationsForNewProcessDefinition() to prevent potential NullPointerExceptions when this method is called with a null process.

**Issue:**
When the process parameter is null, calling process.getCandidateStarterUsers() or process.getCandidateStarterGroups() throws a NullPointerException, which can cause application crashes.

**Fix:**
Added a conditional check to verify the process parameter is not null before attempting to access its methods.

**References:**
Similar issue was fixed in: https://github.com/Activiti/Activiti/commit/f4ad99958ce6cc0672880732cc69d66fad862da6
This implementation applies the same fix pattern to prevent null dereference


filiphr:
@yannaingtun can you please provide a test case for this?

I do not see how this piece of code can throw a `NullPointerException`. Especially, because in the 2 methods before this gets called we get the `process` from the parsed deployment and invoke something on it without doing a null check.

PR #4038, opened 2025-03-14
fix: When a node in the main process moves to the node that calls the activity && processEngineConfiguration.isEnableEntityLinks() ==false  EntityLinkUtil 41 line throw NullPointerException  The reason is that only when it is true ，ProcessEngineConfigurationImpl#entityLinkServiceConfiguration Will be initialized

LoveMyOrange:
fix: When a node in the main process moves to the node that calls the activity && processEngineConfiguration.isEnableEntityLinks() ==false

EntityLinkUtil 41 line throw NullPointerException

The reason is that only when it is true ，ProcessEngineConfigurationImpl#entityLinkServiceConfiguration Will be initialized

The reason why it is normal in the test cases on GitHub is because ChangeStateForCallActiveTest # testSetCurrentActivityInSubProcessInstance  is running during the process   

ProcessEngineConfigurationImpl#entityLinkServiceConfiguration   is true， Resulting in the inability to reproduce official test cases


filiphr:
Thanks @LoveMyOrange

PR #4049, opened 2025-05-22
enhanced RequestUtil to support iso dates with optional seconds, mill…

WelschChristopher:
…iseconds and timezone/offset.

#### Check List:
* Unit tests: YES / NO / NA

### DIFF
--- a/modules/flowable-common-rest/src/test/java/org/flowable/common/rest/api/DateFormattingTest.java
+++ b/modules/flowable-common-rest/src/test/java/org/flowable/common/rest/api/DateFormattingTest.java
@@ -1,54 +1,63 @@
+/* Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


modules/flowable-common-rest/src/test/java/org/flowable/common/rest/api/DateFormattingTest.java
```diff
@@ -0,0 +1,54 @@
+package org.flowable.common.rest.api;
```
dbmalkovsky, line 1: Please add the missing the license header; thanks

modules/flowable-common-rest/src/test/java/org/flowable/common/rest/api/DateFormattingTest.java
```diff
@@ -0,0 +1,54 @@
...
+import java.time.Instant;
+import java.util.Date;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class DateFormattingTest {
+
+    @Test
+    public void testValidDates() {
+        assertEquals(
```
dbmalkovsky, line 14: I believe the project standards are to use `assertThat()` form instead of `assertEquals()`

PR #4058, opened 2025-06-06
Add property for configuring the max allowed length of some variable types

filiphr:
This PR adds support for configuring the maximum allowed size for some variable (e.g. String, Json, Bytes, etc.)

#### Check List:
* Unit tests: YES
* Documentation: NA

### DIFF
--- a/modules/flowable-app-engine/src/main/java/org/flowable/app/engine/AppEngineConfiguration.java
+++ b/modules/flowable-app-engine/src/main/java/org/flowable/app/engine/AppEngineConfiguration.java
@@ -393,7 +393,7 @@ public class AppEngineConfiguration extends AbstractBuildableEngineConfiguration
             }
             variableTypes.addType(new NullType());
             variableTypes.addType(new StringType(getMaxLengthString()));
-            variableTypes.addType(new LongStringType(getMaxLengthString() + 1));
+            variableTypes.addType(new LongStringType(getMaxLengthString() + 1, maxAllowedLengthVariableType));
             variableTypes.addType(new BooleanType());
             variableTypes.addType(new ShortType());
             variableTypes.addType(new IntegerType());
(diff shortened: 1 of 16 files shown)


modules/flowable-app-engine/src/main/java/org/flowable/app/engine/AppEngineConfiguration.java
```diff
@@ -393,7 +393,7 @@ public void initVariableTypes() {
             }
             variableTypes.addType(new NullType());
             variableTypes.addType(new StringType(getMaxLengthString()));
```
martin-grofcik, line 395: @filiphr What if (getMaxLengthString() > maxAllowedLengthVariableType) ?

modules/flowable-variable-service/src/main/java/org/flowable/variable/service/impl/types/ByteArrayType.java
```diff
@@ -41,6 +52,14 @@ public Object getValue(ValueFields valueFields) {
 
     @Override
     public void setValue(Object value, ValueFields valueFields) {
+        if (value == null) {
+            valueFields.setBytes(null);
+            return;
+        }
+
+        byte[] bytes = (byte[]) value;
+        VariableTypeUtils.validateMaxAllowedLength(maxAllowedLength, bytes.length, valueFields, this);
```
martin-grofcik, line 61: @filiphr Is it possible to make variableTypeUtils configurable?
The threshold exception is mandatory in this case.

use case: If we turn on maxAllowedLength the already running process instances (not on the latest definition) can fail.

modules/flowable-app-engine/src/main/java/org/flowable/app/engine/AppEngineConfiguration.java
filiphr: That's an edge case that people configuring should take into consideration. The main use case for this new property is for way larger variables, not the typical `maxLengthString` of 2K or 4K characters. 

I'll see if I can do something about that

martin-grofcik:
The process instance (v1) is on `userTask1` and will create variable `decision` with value `approved-longValue` when user decides to approve.
Meantime `maxAllowedLength` is set to 10, before the approval.
The process instance fails after the approval. Is it possible to replace `VariableTypeUtils` behavior easily without rewriting all types?
The same can happen with large variables not only user inputs.

**Question:** Is the length limit applied on content too? (I know EE).

modules/flowable-variable-service/src/main/java/org/flowable/variable/service/impl/types/ByteArrayType.java
martin-grofcik: The process instance (v1) is on userTask1 and will create variable decision with value approved-longValue when user decides to approve.
Meantime maxAllowedLength is set to 10, before the approval.
The process instance fails after the approval. Is it possible to replace VariableTypeUtils behavior easily without rewriting all types? (e.g. log a warning instead of exception....)
The same can happen with large variables not only user inputs.

Question: 
Is the length limit applied on the content and indexed entities too? (I know EE).

martin-grofcik:
PR -> I will try.

`JpaEntityListVariableType` serializes ids into the bytes. 
The bytes size can exceed `maxAllowedLength ` without the check.

PR #4060, opened 2025-06-11
Add an interface that can be customized for the variable length verifier

filiphr:
#### Check List:
* Unit tests: YES
* Documentation: NA


martin-grofcik:
Hi @filiphr ,

What I meant by https://github.com/flowable/flowable-engine/pull/4058#issuecomment-2958475192
> PR -> I will try.

Was little bit different, The cons of the proposal is focus on variable length only.
I would prefer an interface change may be on `VariableType` level which will allow to preprocess (for set) and postprocess (for set) value.

The default implementation can be an empty for most of the models. We can add length validator, or data encryption ++ in such a case.

Issue #4062, opened 2025-06-18
[JOB_RETRIES_DECREMENTED] CMMN support for JOB_RETRIES_DECREMENTED event


martin-grofcik:
**Problem description:**
Process engine fires `JOB_RETRIES_DECREMENTED` `FlowableEngineEvent`.
(!) CMMN engine does not fire the event.
If there is a listener which listens to the `JOB_RETRIES_DECREMENTED` event half of the events are lost. In fact it does not make sense to fire the event for process events in this case.

**Solution proposal:**

Add event firing code to cmmn `JobRetryCmd`.

**Describe alternatives you've considered**

JobRetryCmd can be replaced, but the engines should behave consistently.

**Additional context**

None.

PR #4063, opened 2025-06-18
#4062 cmmn job_retries_decremented event impl.

martin-grofcik:
#### Check List:
* Unit tests: YES
* Documentation:  NO


modules/flowable-cmmn-engine/src/test/java/org/flowable/cmmn/test/event/JobRetriesDecrementedEventTest.java
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.flowable.cmmn.test.event;
+
+import org.flowable.cmmn.api.CmmnRuntimeService;
```
dbmalkovsky, line 15: I think the order of the imports are wrong.  It should be statics, then java, then org.flowable.

Issue #4095, opened 2025-08-14
Fails to Create Tables in New Database When Multiple Databases Exist With Same Table Names


Taylor1221:
**Describe the bug**
When multiple MySQL databases exist in the same instance (e.g., common_activiti and common_workflow) and one database (common_activiti) already contains Flowable tables, a new Spring Boot application configured to use an empty database (common_workflow) with flowable.database-schema-update=true fails to create required tables. Flowable incorrectly detects tables as existing due to cross-database visibility, causing startup failure with Table doesn't exist errors.

**Expected behavior**
1. Flowable detects common_workflow is empty
2. Automatically creates all required tables
3. Application starts successfully
#

**Code**
`spring:`
`  application:`
`    name: WorkflowDemoApplication`
`  datasource:`
`    url: jdbc:mysql://localhost:3306/common_workflow?useUnicode=true&characterEncoding=UTF-8&serverTimezone=Asia/Shanghai`
`    username: root`
`    password: 123456`
`flowable:`
`  database-schema-update: true`
`  async-executor-activate: false`
`  check-process-definitions: false`
`  history-level: full`

**Additional context**
flowable-spring-boot-starter 6.8.1
Database: MySQL 8.2.0
Spring Boot: 2.7.18

Taylor1221:
## Proposed Fix Implemented
I've implemented the solution in my fork repository:
 **Commit**: [Fix flowable incorrectly detects tables as existing due to cross-database visibility, causing startup failure with Table doesn't exist errors.](https://github.com/Taylor1221/flowable-engine/commit/d877dd41052a3ec327131285be3580ffd38c7216)

filiphr:
@Taylor1221 have you tried using the `flowable.database-schema` property for this?

filiphr:
Perhaps for this we should expose `flowable.database-catalog` so it can be set through properties and / or we should do something in `AbstractSqlScriptBasedDbSchemaManager` and `TableDataManagerImpl` and get the catalog from the connection if not set.

Taylor1221:
@filiphr Thanks for the suggestion! I'll try implementing this approach tomorrow:
1. Expose `flowable.database-catalog` property in configuration
2. Enhance `AbstractEngineAutoConfiguration#configureEngine()` to:
   - First use the configured catalog if available
   - Fall back to dynamically obtaining catalog from DataSource connection
   
I'll submit an updated PR tomorrow with these changes. If you have any specific implementation preferences, please let me know. ^_^

PR #4096, opened 2025-08-15
feat(engine,spring-boot): support configurable databaseCatalog via Fl…

Taylor1221:
feat(engine,spring-boot): support configurable databaseCatalog via FlowableProperties (#4095)

- Added `databaseCatalog` property to `FlowableProperties` to allow explicit configuration in Spring Boot.
- Updated `AbstractEngineAutoConfiguration.configureEngine(...)` to set `databaseCatalog` from properties if not already configured in the engine.
- Added fallback in `AbstractEngineConfiguration` to initialize `databaseCatalog` from the DataSource's connection catalog when no value is set.

#### Check List:
* Unit tests: NA
* Documentation: NA


filiphr:
Thanks for the PR @Taylor1221.

I am not sure that we need the initialization of the database catalog in the `AbstractEngineConfiguration`. I would rather just update the 2 places where we use the database catalog. Namely the `AbstractSqlScriptBasedDbSchemaManager` and `TableDataManagerImpl`. In those 2 places, we are already in a transaction and we already have the `Connection`. Otherwise, it means that we need to open a DB connection on bootup to check this, and in my opinion I don't think that it makes sense to do that. What do you think?

PR #4103, opened 2025-08-23
Expose cmmn query by caseInstanceIds

roumn:
#### Check List:
* Unit tests: YES
* Documentation: NO


filiphr (commented):
Thanks @roumn. Can we perhaps also expose this in the `CaseInstanceCollectionResource`?

PR #4104, opened 2025-08-25
Add endUserId for hist process/case and state for hist process

BertMatthys:
#### Check List:
* Unit tests: YES
* Documentation: NA

Add endUserId field for HistoricProcessInstance and HistoricCaseInstance.
Add state field for HistoricProcessInstance.


modules/flowable-engine/src/test/java/org/flowable/engine/test/api/history/HistoricProcessInstanceQueryTest.java
```diff
@@ -651,5 +653,81 @@ public void testIncludeDefinedVariables() {
                     .isEmpty();
         }
 
+    public void testQueryByState() {
```
dbmalkovsky, line 656: is the `@Test` annotation missing on this method?

Issue #3921, opened 2024-07-11
Process start, potentially migration: Signal scope gets ignored starting with the 2nd process


robert-gurol-ebcont:
**Describe the bug**
If I start a process that has a Signal event definition and a Signal Start Event, starting it will work normally the first time. 
The second time, the resulting event subscription's scope definition will always be "null", because the event definition's signal ref gets overwritten the first time.

Potentially, this may also affect process instance migrations (there are 2 places in the code where the behavior is like this).

**Expected behavior**
I expect the same behavior each time - the signalref stays as-is, the event subscription always gets created with the right scope.

**Code**
I prepared a Pull Request here: https://github.com/flowable/flowable-engine/pull/3920

I would love to add regression tests as well, but I can't seem to find a good place.  
Please point me to a "good" place where I can add a test method as a regression test.

**Additional context**
All that should matter is the flowable version. 
I tried with 6.8.1 and 7.0.1, the bug is in both and seems to go back very far. 

**Workaround**
Without a fix, the bug does not occur if Signal ID and name must are the same, e.g.
`<bpmn:signal id="terminate_product" name="terminate_product" flowable:scope="processInstance" />`

PR #4109, opened 2025-08-26
fix issue 3921

OmarAlsawaf:
fix for [/issues/3921](https://github.com/flowable/flowable-engine/issues/3921) where the SignalRef was getting assigned the value of the Name by mistake resulting a null when querying the signal.


OmarAlsawaf:
fix issue 3921

PR #4116, opened 2025-09-03
added case and process end interceptor

WelschChristopher:
#### Check List:
* Unit tests: YES / NO / NA


GianniGiglio:
@tijsrademakers just out of curiosity: in which scenarios would you recommend using the new interceptors? How do they compare to something like AbstractProcessEngineListener? I currently use AbstractProcessEngineListener to react on process events — for example, to store arbitrary data transactionally or to add observability hooks on transaction commit.

tijsrademakers:
@GianniGiglio , can you share more details about what you mean with the AbstractProcessEngineListener, because I don't see that in our codebase. The new interceptors can help to plugin logic that should be executed when a case or process instance is ended, either with regular execution or via terminate / delete.

PR #4140, opened 2025-11-14
fix_noCatalog

redfish-zly:
#### Check List:
* Unit tests: YES / NO / NA


filiphr (commented):
What are you trying to fix here @redfish-zly? I am not sure that we should be using the connection on this level. The `connectionMetadataDefaultCatalog` is eventually used only in `AbstractSqlScriptBasedDbSchemaManager`, so if something needs changing we can maybe do it directly there.

Would also be good to have a bit more insight in what problem you are trying to solve

redfish-zly:
image

image

@filiphr 
Hello senior, I am not sure if my solution is correct. Here are the problems I have encountered.
Preconditions:
There are two database instances on the MySQL server, flowableTest1,
flowableTest2。
Among them, flowableTest1 has already initialized the process related tables (Act_ *), but flowableTest2 does not.
At this point, when my springboot project connects to flowableTest2 and says' flowable. database schema update=create ', the initialization of the flowableTest2 table fails due to the null catalog.

PR #4158, opened 2026-01-22
Fix boundary event iterator desynchronization in ContinueProcessOperation

uucoding:
**The `executeBoundaryEvents` method had a critical iterator desynchronization bug. The `while` loop assumed that the `boundaryEvents` and `boundaryEventExecutions` lists contained matching elements in the same order, but they did not due to inconsistent filtering.**

**For example:**
- `createBoundaryEvents` returns only: `[TimerExec, ErrorExec]` (compensation events are skipped)
- `executeBoundaryEvents` receives: `[Timer, Compensation, Error]` (all events)

**Loop pairing comparison:**
```
Expected pairing:              Actual incorrect pairing:
Timer ↔ TimerExec ✓            Timer ↔ TimerExec ✓
Error ↔ ErrorExec ✓            Compensation ↔ ErrorExec ✗ 
```

### DIFF
--- a/modules/flowable-engine/src/main/java/org/flowable/engine/impl/agenda/ContinueProcessOperation.java
+++ b/modules/flowable-engine/src/main/java/org/flowable/engine/impl/agenda/ContinueProcessOperation.java
@@ -13,9 +13,7 @@
 package org.flowable.engine.impl.agenda;
 
 import java.util.ArrayList;
-import java.util.Iterator;
 import java.util.List;
-import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


filiphr:
@uucoding can you please write a test case for this?

uucoding:
Hi @filiphr ,

Hope you're doing well!

Just circling back on the test case I submitted last month for the boundary event iterator desynchronization fix. When you have a moment, could you please review it and let me know if any changes are needed?

Thanks so much!

PR #4175, opened 2026-03-04
Disable lenient parsing in DateFormType

gilisd:
FastDateFormat.parseObject() silently accepted invalid dates by rolling over values (e.g. month 13 became month 1 of the next year). Use SimpleDateFormat with setLenient(false) to reject invalid dates.

#### Check List:
* Unit tests: YES 
* Documentation: NA

### DIFF
--- a/modules/flowable-bpmn-converter/src/main/java/org/flowable/bpmn/constants/BpmnXMLConstants.java
+++ b/modules/flowable-bpmn-converter/src/main/java/org/flowable/bpmn/constants/BpmnXMLConstants.java
@@ -366,6 +366,7 @@ public interface BpmnXMLConstants {
     public static final String ATTRIBUTE_FORM_REQUIRED = "required";
     public static final String ATTRIBUTE_FORM_DEFAULT = "default";
     public static final String ATTRIBUTE_FORM_DATEPATTERN = "datePattern";
+    public static final String ATTRIBUTE_FORM_LENIENT_DATE_PARSING = "lenientDateParsing";
     public static final String ELEMENT_VALUE = "value";
 
     public static final String ELEMENT_FIELD = "field";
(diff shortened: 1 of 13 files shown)


DavidGilis:
@jbarrez @filiphr Thanks for the review. I've reworked the PR based on your comments — a summary of what changed and why:

**FastDateFormat field restored — formatting unchanged.** `convertModelValueToFormValue` and the lenient parse path use the shared `FastDateFormat` field again, so there is no behaviour change in formatting. Side note on the `Y`-pattern concern: that javadoc passage describes the Java ≤6 vs 7+ difference, and on the JDK 17 baseline both implementations produce identical output for the patterns I tested. Keeping FastDateFormat for formatting still seems the safer choice, so I did.

**Why the strict parse path has no second field.** Strict parsing now uses `DateUtils.parseDateStrictly(value, datePattern)`, which besides disabling leniency also rejects trailing characters — a plain non-lenient `SimpleDateFormat` still accepted `"15/06/2024abc"`, since `parse()` only matches a prefix. On the two-fields suggestion: form types registered on the engine (the default `"date"` type and anything in `customFormTypes`) are engine-wide singletons used concurrently, and `SimpleDateFormat` is unsafe under concurrent *use* even when the field reference never changes — its internal `Calendar` mutates during parse. The per-datePattern instances are in any case rebuilt on every form command by `FormHandlerHelper`, so a parse-side field would save one pattern compilation per submit. A local parse seemed the better trade-off.

**Backwards compatibility: `lenientDateParsing` on two levels.** As suggested, there is now an opt-out — as an engine-wide default (`ProcessEngineConfiguration.setLenientDateParsing`) and as a per-property BPMN attribute (`<flowable:formProperty ... lenientDateParsing="true"/>`), with the property-level value taking precedence (`Boolean`: absent = follow engine default). That gives installations a single switch to restore the old behaviour, and per-definition granularity — which also covers multi-tenant setups, since deployments are tenant-scoped. I kept strict as the default because lenient rollover silently stores corrupted dates: the problem then surfaces far downstream as wrong data to troubleshoot, instead of a parse error at the boundary. I picked the name `lenientDateParsing` over `lenientDateFormatParse` since it names the behaviour rather than the implementing class — I can rename if you prefer your original suggestion.

**Known limitations, intentionally left as-is:** the property-level flag is only honoured when a `datePattern` is set (a date property without one resolves to the engine-registered shared type); strict mode also rejects trailing text and case-variant month names that the old `FastDateFormat` parser tolerated; and the flowable5 compatibility engine keeps its own (lenient) `DateFormType`.

filiphr (approved):
Thanks @DavidGilis. This looks OK to me. I'll let the CI run and then we can merge it

PR #4200, opened 2026-04-09
Filter all exceptions when sendFullErrorException is false

roumn:
Previously, only handleBadMessageConversion and handleOtherException in BaseExceptionHandlerAdvice respected the sendFullErrorException flag. All other exception handlers (handleNotSupported, handleConflict, handleNotFound, handleForbidden, handleIllegalArgument, handleIllegalState, handleTaskAlreadyClaimed) always returned the full exception message in the response, regardless of the flag.

This change applies the same pattern consistently to all handlers: when sendFullErrorException is false, a UUID error identifier is generated and logged at WARN level, and only the error identifier is returned to the client — preventing internal error details from leaking in the REST API response.

The test class has been updated accordingly, using ${json-unit.regex} matchers to verify the error ID pattern in each handler's response.


roumn:
Does not make sense, will close PR. This would break too many things.

PR #4203, opened 2026-04-16
适配kingbase数据库三种模式

198lma:
增加kingbase数据库三种模式适配


filiphr:
@198lma from what I can see in your changes:

```java
switch (mode.toLowerCase()) {
    case "mysql" -> databaseType = "kingbase-mysql";
    case "oracle" -> databaseType = "kingbase-oracle";
    case "sqlserver" -> databaseType = "kingbase-sqlserver";
    case "pg" -> databaseType = "postgres";
    default -> throw new FlowableException("Unknown Kingbase database_mode: " + mode);
}
```

it looks like Kingbase has support for different modes. I am not familiar with the Kingbase database. However, won't the DDL we have for MySQL, Oracle, SQL Server or Postgres work on Kingbase as well? I would expect that it does if it has those modes, so I would assume that you can manually set the `databaseType` to what you need.

Therefore, for now I'll close the PR. If we receive more interest from the community to support this DB we can explore it a bit more. However, if we do not have a way of testing against it, we can't officially support it.

PR #4212, opened 2026-05-20
fixed multiple typos and copy paste errors in the rest API docu

WelschChristopher:
also fixed missing withoutDueDate in task queries

#### Check List:
* Unit tests: YES / NO / NA


modules/flowable-cmmn-rest/src/main/java/org/flowable/cmmn/rest/service/api/repository/CaseDefinitionCollectionResource.java
filiphr, line 1: Should we maybe be consistent like we are in the other resourcs and only mention the encoding in the `ApiOperation#notes`?

filiphr (commented):
Only one comment / question @WelschChristopher

filiphr (approved):

PR #4219, opened 2026-06-29
Pass the source definition to migration callbacks

roumn:
## What

`CaseInstanceMigrationCallback` and `ProcessInstanceMigrationCallback` previously only received the definition an instance was migrated *to*. They now also receive the definition it was migrated *from*, so listeners can compare source and target (e.g. version-aware auditing or reacting to specific upgrade paths).

## How

Backwards compatible via a new `default` overload on each callback method that takes the source definition as an extra parameter and delegates to the existing method. Existing implementations keep working unchanged; the engines now invoke the new overloads.

- `CaseInstanceMigrationCallback#caseInstanceMigrated` and `#historicCaseInstanceMigrated` gain a `sourceCaseDefinition` parameter.
- `ProcessInstanceMigrationCallback#processInstanceMigrated` gains a `sourceProcessDefinition` parameter.

The source definition is resolved from the instance's pre-migration definition id, captured before the reference is overwritten, and only resolved when at least one callback is registered (no overhead otherwise).

## Tests

Adds `CaseInstanceMigrationCallbackTest` and `ProcessInstanceMigrationCallbackTest` covering the source definition on runtime and historic migrations, plus that callbacks implementing only the original methods are still invoked. Existing migration test suites (`CaseInstanceMigrationTest`, `HistoricCaseInstanceMigrationTest`, `ProcessInstanceMigrationTest`) remain green.


filiphr (approved):
Looks good. Thanks @roumn

Issue #4213, opened 2026-05-21
flowable-spring-boot-autoconfigure: register a DependsOnDatabaseInitializationDetector so the ProcessEngine waits for Flyway/Liquibase


galimru:
**Is your feature request related to a problem? Please describe.**

When `flowable-spring-boot-starter-process` is used alongside `spring-boot-starter-flyway` (or any other Spring Boot–managed DB initializer: Liquibase, `spring.sql.init`), there is no guaranteed bean-instantiation order between Flowable's `processEngine` and Flyway's `flywayInitializer`. Spring Boot's `DatabaseInitializationDependencyConfigurer` is the framework's mechanism for this — it walks beans flagged by registered `DependsOnDatabaseInitializationDetector` SPI implementations and adds `@DependsOn(flywayInitializer)` to each.

Spring Boot ships detectors for `EntityManagerFactory`, `JdbcOperations`, `JobRepository`, etc. `flowable-spring-boot-autoconfigure` does not register one for `processEngine`, and `ProcessEngineAutoConfiguration` does not annotate the bean with `@DependsOnDatabaseInitialization`. As a result, Flowable can be instantiated before Flyway, creates its `ACT_*` tables into an empty schema, and Flyway then refuses to start because the schema is non-empty but contains no `flyway_schema_history` row.

This often went unnoticed in Spring Boot 3 because `HibernateJpaAutoConfiguration` indirectly delayed Flowable through the JPA bean graph. With Spring Boot 4's modularised autoconfig — and projects increasingly dropping JPA in favour of plain JDBC, jOOQ, or Exposed — that incidental ordering vanishes and the race surfaces deterministically.

Steps to reproduce on Spring Boot 4.0.6 / Flowable 8.0.0:

1. `spring-boot-starter-jdbc` + `spring-boot-starter-flyway` + `flowable-spring-boot-starter-process` on the classpath.
2. **No** `spring-boot-starter-data-jpa`.
3. Empty database, one Flyway migration in `db/migration/`.
4. Boot the app. Flyway fails with `Schema "public" is not empty` because Flowable created `ACT_*` first.

**Describe the solution you'd like**

Register a `DependsOnDatabaseInitializationDetector` in `flowable-spring-boot-autoconfigure` that returns the names of the engine beans whose definitions are present (`processEngine`, and analogously `cmmnEngine`, `dmnEngine`, `appEngine`, etc.).

[...]

PR #4222, opened 2026-07-15
Add DependsOnDatabaseInitializationDetector so engines wait for database initializers

nikhiln64:
Implements #4213, following the approach @filiphr outlined there ("You can go ahead and create a PR for this").

**What:** a `FlowableDependsOnDatabaseInitializationDetector` extending Spring Boot's `AbstractBeansOfTypeDependsOnDatabaseInitializationDetector` for `org.flowable.common.engine.api.Engine` (covering process/cmmn/dmn/app/idm engines through the shared interface, no bean names), registered under the `DependsOnDatabaseInitializationDetector` SPI in `spring.factories`. Spring Boot's `DatabaseInitializationDependencyConfigurer` then orders every engine bean after Flyway/Liquibase/`spring.sql.init` initializers.

**Opt-out:** `flowable.depends-on-database-initialization-detection=false` (default `true`), mirroring the JPA detector's opt-out pattern per the issue discussion; documented in `additional-spring-configuration-metadata.json`.

**Testing:** `FlowableDependsOnDatabaseInitializationDetectorTest` uses `ApplicationContextRunner` with `spring.sql.init` and asserts the `processEngine` bean definition gains `dependsOn` on `dataSourceScriptDatabaseInitializer`, plus the opt-out case. Verified the assertion goes red when the `spring.factories` registration is removed. I can add a Flyway-based sample under `flowable-spring-boot-samples` as a follow-up if you'd prefer that shape of test as well.

- [x] Unit tests added
- [x] Documentation: property registered in configuration metadata

Developed with AI assistance (Claude Code), reviewed and driven by a human.

---
<a href="https://app.blacksmith.sh/flowable/codesmith/flowable-engine/pr/4222"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786673757&installation_id=145635894&pr_number=4222&repository=flowable%2Fflowable-engine&return_to=https%3A%2F%2Fgithub.com%2Fflowable%2Fflowable-engine%2Fpull%2F4222&signature=00aa8800cb270317cd4249d8088f92de6828d076fe62a803e2289b50aedf54db"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>

### DIFF
--- a/modules/flowable-spring-boot/flowable-spring-boot-starters/flowable-spring-boot-autoconfigure/src/test/java/org/flowable/test/spring/boot/FlowableDependsOnDatabaseInitializationDetectorTest.java
+++ b/modules/flowable-spring-boot/flowable-spring-boot-starters/flowable-spring-boot-autoconfigure/src/test/java/org/flowable/test/spring/boot/FlowableDependsOnDatabaseInitializationDetectorTest.java
@@ -1,0 +1,76 @@
+/* Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.flowable.test.spring.boot;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.flowable.spring.boot.ProcessEngineAutoConfiguration;
+import org.flowable.spring.boot.ProcessEngineServicesAutoConfiguration;
+import org.junit.jupiter.api.Test;
+import org.springframework.boot.autoconfigure.AutoConfigurations;
+import org.springframework.boot.jdbc.autoconfigure.DataSourceAutoConfiguration;
+import org.springframework.boot.jdbc.autoconfigure.DataSourceInitializationAutoConfiguration;
+import org.springframework.boot.jdbc.autoconfigure.DataSourceTransactionManagerAutoConfiguration;
+import org.springframework.boot.test.context.runner.ApplicationContextRunner;
+
+/**
+ * The Flowable engines create/validate their own schema on engine build, so their beans must be
+ * ordered after Spring Boot managed database initializers (Flyway, Liquibase, spring.sql.init).
+ * This is done through the {@code DependsOnDatabaseInitializationDetector} SPI (issue #4213).
+ *
+ * @author Nikhil Bharadwaj Ramashasthri
+ */
+class FlowableDependsOnDatabaseInitializationDetectorTest {
+
+    private final ApplicationContextRunner contextRunner = new ApplicationContextRunner()
+            .withConfiguration(AutoConfigurations.of(
+                    DataSourceAutoConfiguration.class,
+                    DataSourceTransactionManagerAutoConfiguration.class,
+                    DataSourceInitializationAutoConfiguration.class,
+                    ProcessEngineAutoConfiguration.class,
+                    ProcessEngineServicesAutoConfiguration.class));
+
+    @Test
+    void processEngineDependsOnDatabaseInitialization() {
+        contextRunner
+                .withPropertyValues("spring.sql.init.mode=always")
+                .run(context -> {
+                    assertThat(context).hasNotFailed();
+                    String[] dependsOn = context.getBeanFactory()
+                            .getBeanDefinition("processEngine")
+                            .getDependsOn();
+                    assertThat(dependsOn)
+                            .as("processEngine must wait for the SQL init database initializer")
+                            .isNotNull()
+                            .contains("dataSourceScriptDatabaseInitializer");
+                });
+    }
+
+    @Test
+    void detectionCanBeOptedOut() {
+        contextRunner
+                .withPropertyValues(
+                        "spring.sql.init.mode=always",
+                        "flowable.depends-on-database-initialization-detection=false")
+                .run(context -> {
+                    assertThat(context).hasNotFailed();
+                    String[] dependsOn = context.getBeanFactory()
+                            .getBeanDefinition("processEngine")
+                            .getDependsOn();
+                    assertThat(dependsOn == null || !java.util.Arrays.asList(dependsOn)
+                            .contains("dataSourceScriptDatabaseInitializer"))
+                            .as("opt-out must remove the database initializer dependency")
+                            .isTrue();
+                });
+    }
+}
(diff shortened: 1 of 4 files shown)


modules/flowable-spring-boot/flowable-spring-boot-starters/flowable-spring-boot-autoconfigure/src/test/java/org/flowable/test/spring/boot/FlowableDependsOnDatabaseInitializationDetectorTest.java
```diff
@@ -0,0 +1,76 @@
...
+        contextRunner
+                .withPropertyValues("spring.sql.init.mode=always")
+                .run(context -> {
+                    assertThat(context).hasNotFailed();
+                    String[] dependsOn = context.getBeanFactory()
+                            .getBeanDefinition("processEngine")
+                            .getDependsOn();
+                    assertThat(dependsOn)
+                            .as("processEngine must wait for the SQL init database initializer")
+                            .isNotNull()
```
filiphr: The `isNotNull` is obsolete

modules/flowable-spring-boot/flowable-spring-boot-starters/flowable-spring-boot-autoconfigure/src/test/java/org/flowable/test/spring/boot/FlowableDependsOnDatabaseInitializationDetectorTest.java
```diff
@@ -0,0 +1,76 @@
...
+                    DataSourceAutoConfiguration.class,
+                    DataSourceTransactionManagerAutoConfiguration.class,
+                    DataSourceInitializationAutoConfiguration.class,
+                    ProcessEngineAutoConfiguration.class,
+                    ProcessEngineServicesAutoConfiguration.class));
+
+    @Test
+    void processEngineDependsOnDatabaseInitialization() {
+        contextRunner
+                .withPropertyValues("spring.sql.init.mode=always")
```
filiphr: Can we also set `flowable.check-process-definitions=false` both here and the other test in order to avoid doing the auto deployment of the definitions.

modules/flowable-spring-boot/flowable-spring-boot-starters/flowable-spring-boot-autoconfigure/src/test/java/org/flowable/test/spring/boot/FlowableDependsOnDatabaseInitializationDetectorTest.java
```diff
@@ -0,0 +1,76 @@
...
+                        "flowable.depends-on-database-initialization-detection=false")
+                .run(context -> {
+                    assertThat(context).hasNotFailed();
+                    String[] dependsOn = context.getBeanFactory()
+                            .getBeanDefinition("processEngine")
+                            .getDependsOn();
+                    assertThat(dependsOn == null || !java.util.Arrays.asList(dependsOn)
+                            .contains("dataSourceScriptDatabaseInitializer"))
+                            .as("opt-out must remove the database initializer dependency")
+                            .isTrue();
```
filiphr: The failure message for this is not clear. Let's make it:


```suggestion
                    assertThat(dependsOn)
                            .as("opt-out must remove the database initializer dependency")
                            .doesNotContain("dataSourceScriptDatabaseInitializer");
```

filiphr (changes requested):
Thanks @nikhiln64. This looks good, I've only left some comments for the tests

modules/flowable-spring-boot/flowable-spring-boot-starters/flowable-spring-boot-autoconfigure/src/test/java/org/flowable/test/spring/boot/FlowableDependsOnDatabaseInitializationDetectorTest.java
nikhiln64: Removed.

modules/flowable-spring-boot/flowable-spring-boot-starters/flowable-spring-boot-autoconfigure/src/test/java/org/flowable/test/spring/boot/FlowableDependsOnDatabaseInitializationDetectorTest.java
nikhiln64: Added to both tests.

modules/flowable-spring-boot/flowable-spring-boot-starters/flowable-spring-boot-autoconfigure/src/test/java/org/flowable/test/spring/boot/FlowableDependsOnDatabaseInitializationDetectorTest.java
nikhiln64: Applied with one adjustment: with the opt-out active nothing sets `dependsOn` on the bean definition, so it comes back `null` and `doesNotContain` fails on the null actual. Used `isNullOrEmpty()` instead - same readable shape, and strictly stronger for this case.

filiphr: Good pint. Thanks @nikhiln64

nikhiln64:
@filiphr just a gentle nudge on this one. All the review comments were addressed back on July 15 and the branch is current with main, so whenever you get a moment for another look I believe it is ready.

filiphr (approved):
Thanks @nikhiln64. I've let the CI run one more time and I'll merge it afterwards

nikhiln64:
Quick note on the JDK 25 run since you mentioned merging after CI. The only failure is KafkaChannelDefinitionProcessorTest.fixedBackOffMultiTopicRetry in the event registry module, which errored with Topic(s) [fixed-backoff-multi-retry-topic-0] is/are not present and missingTopicsFatal is true, so the Kafka topic was not created in time on the CI broker. The JDK 17 and JDK 21 runs on the same commit are green and this PR only touches the spring boot autoconfigure module, so it looks like an infra timing flake rather than anything from the change.

filiphr:
Thanks @nikhiln64. That test is indeed a bit flaky, we'll look into it separately. I've merged the PR

PR #4229, opened 2026-07-27
fix cmmn faultCode/faultMessage in conditions in case multiple sentries trigger on fault

vzickner:
#### Check List:
* Unit tests: YES
* Documentation: NA


jbarrez:
Thx @vzickner , good catch

PR #4232, opened 2026-07-28
fix(flowable-engine-common):Resource Leakage Issue in Flowable-Engine…

panshaobin:
fix(flowable-engine-common):Resource Leakage Issue in Flowable-Engine-Common Module AbstractSqlScriptBasedDbSchemaManager.java


filiphr:

PR #4233, opened 2026-07-28
fix(flowable-job):timerEntity.getRepeat() in TimerJobEntityManagerImpl may return null, which causes a NullPointerException when calling split()

panshaobin:
fix(flowable-job):timerEntity.getRepeat() in TimerJobEntityManagerImpl may return null, which causes a NullPointerException when calling split()


filiphr:

PR #4234, opened 2026-07-28
fix(DeploymentCollectionResource):When using fileName.split("\\.")[0] to get the filename, a ArrayIndexOutOfBoundsException will be thrown if the filename has no extension.

panshaobin:
fix(DeploymentCollectionResource):When using fileName.split("\\.")[0] to get the filename, a ArrayIndexOutOfBoundsException will be thrown if the filename has no extension.


filiphr:

PR #4235, opened 2026-07-28
fix(DefaultCmmnHistoryConfigurationSettings):Two empty catch blocks silently suppress exceptions, resulting in untraceable errors

panshaobin:
fix(DefaultCmmnHistoryConfigurationSettings):Two empty catch blocks silently suppress exceptions, resulting in untraceable errors


filiphr:
@panshaobin this PR is against main, but I think that you used a wrong base to create the branch. Closing it for now.

PR #4238, opened 2026-07-28
fix timerEntity.getRepeat() in TimerJobEntityManagerImpl may return null, which causes a NullPointerException when calling split()

panshaobin:
fix(flowable-job):timerEntity.getRepeat() in TimerJobEntityManagerImpl may return null, which causes a NullPointerException when calling split()


filiphr (changes requested):
Thanks for the PR @panshaobin. Can you please add a test case for this? I do not think that this can ever happen since at this point the `timerEntity.getRepeat()` will not be null, because this gets called from:

https://github.com/flowable/flowable-engine/blob/36b4bc480f07982795c0aa4d9138afac183f07b0/modules/flowable-job-service/src/main/java/org/flowable/job/service/impl/asyncexecutor/TimerJobSchedulerImpl.java#L41-L42

filiphr:
Closing this until we get feedback on my last comment

PR #4242, opened 2026-07-30
Fix identityService binding name in VariableScopeResolver

kle-dev:
The identity service was registered in `SERVICE_RESOLVERS` under the key `identityServiceKey` instead of `identityService`. The map key is the name a script author types, so `identityService` did not resolve in BPMN scripts while every other engine service was exposed under its plain name.

The typo originates from the original `String` constants, where the field name was accidentally used as the value (all seven siblings followed `<name>ServiceKey = "<name>Service"`), and was carried over verbatim when the constants were replaced by the `SERVICE_RESOLVERS` map.

`identityService` is now the primary binding name. The former `identityServiceKey` is kept as an alias so that scripts relying on it keep working, so this change is not breaking.

#### Check List:
* Unit tests: YES
* Documentation: NA


filiphr (changes requested):
Thanks for the fix @kle-dev. Can we keep the old typo `identityServiceKey`, in case someone actually used it.

kle-dev:
Done. Thanks.

PR #4244, opened 2026-07-31
fix Calling get(0) without checking the event definitions of targetEvent

panshaobin:
fix Calling get(0) without checking the event definitions of targetEvent


filiphr:
@panshaobin if this can happen then please provide a test for this. We have

```
 if (sourceEvent.getEventDefinitions() != null && !sourceEvent.getEventDefinitions().isEmpty() && 
        (targetEvent.getEventDefinitions() == null || targetEvent.getEventDefinitions().isEmpty())) {

    return false;

}
```

in the beginning of the method. So if the source event definitions are null or empty and the target event definitions are null or empty then we are going to return earlier, which means at the place you added the target event definitions cannot be null or empty

PR #4245, opened 2026-07-31
fix In method calculateMaxIterationsValue, Integer.parseInt does not handle NumberFormatException. Invalid loop expressions will cause crashes

panshaobin:
fix In method calculateMaxIterationsValue, Integer.parseInt does not handle NumberFormatException. Invalid loop expressions will cause crashes


filiphr:
@panshaobin can you please provide a test case for this? Something like this shouldn't really happen, if for whatever reason it happens then it means that there is some misconfiguration somewhere and having a loud exception is more correct. Therefore, I am closing this for now.

PR #4247, opened 2026-07-31
fix Calling getValue() directly will throw an NPE if the record of next.dbid does not exist in the database (property is null)

panshaobin:
fix Calling getValue() directly will throw an NPE if the record of next.dbid does not exist in the database (property is null)


filiphr:
The only way that this can happen if if someone deletes the `next.dbid` property manually since we have

```sql
insert into ACT_GE_PROPERTY
values ('next.dbid', '1', 1);
```

in our DDL. Therefore, I am closing this PR

PR #4253, opened 2026-08-07
Fix self-deadlock in registerTenant() on databases with transactional…

### LINKED ISSUE #4215: `registerTenant` deadlocks on SQL Server when called post-boot
## Symptom

`MultiSchemaMultiTenantProcessEngineConfiguration.registerTenant(tenantId, ds)` called after `buildProcessEngine()` hangs indefinitely on SQL Server when both `databaseSchemaUpdate=true` and `useLockForDatabaseSchemaUpdate=true` are set. The latter is typically enabled in clustered deployments to coordinate concurrent schema updates across nodes. Reproduced on Flowable **8.0.0**; SQL Server 2019/2022.

Not reproducible on MariaDB/MySQL (DDL implicitly commits, so the cycle below never forms).

## Stack trace (abridged)

```
"register-tenant-0" WAITING — parked on socket read from SQL Server
  at org.flowable.common.engine.impl.cmd.LockCmd.execute(LockCmd.java:71)
  at org.flowable.common.engine.impl.lock.LockManagerImpl.executeCommand(LockManagerImpl.java:154)
  at org.flowable.common.engine.impl.lock.LockManagerImpl.waitForLock(LockManagerImpl.java:73)
  at org.flowable.engine.impl.db.ProcessDbSchemaManager.schemaUpdate(ProcessDbSchemaManager.java:166)
  at org.flowable.common.engine.impl.db.SchemaOperationsEngineBuild.execute(SchemaOperationsEngineBuild.java:68)
  at org.flowable.engine.impl.cfg.multitenant.ExecuteSchemaOperationCommand.execute(...)
  at ...createTenantSchema(MultiSchemaMultiTenantProcessEngineConfiguration.java:149)
  at ...registerTenant(MultiSchemaMultiTenantProcessEngineConfiguration.java:88)
```

## Root cause

`SchemaOperationsEngineBuild` opens transaction `T_outer` on connection `C_outer` and runs DDL on `ACT_GE_PROPERTY`, taking Sch-M (held until commit). It then calls `LockManagerImpl.waitForLock` from inside `T_outer`. `LockManagerImpl` deliberately opens a separate transaction `T_inner` on a separate connection `C_inner` so the lock can commit independently, and runs `SELECT * FROM ACT_GE_PROPERTY WHERE NAME_ = ?` — which needs Sch-S, blocked by `T_outer`'s Sch-M. `T_outer` cannot commit until the outer call returns, which cannot happen until the inner SELECT runs. No RDBMS deadlock detector fires because the blocking session is idle.

Confirmed via `sys.dm_exec_requests`: blocker `sleeping`, holds `LCK_M_SCH_S` wait on `OBJECT_NAME(...) = 'ACT_GE_PROPERTY'`, both sessions from the same application host.

## Minimal repro (sketch)

```java
MultiSchemaMultiTenantProcessEngineConfiguration config =
        new MultiSchemaMultiTenantProcessEngineConfiguration(tenantInfoHolder);
config.setDatabaseType(DATABASE_TYPE_MSSQL);
config.setDatabaseSchemaUpdate(DB_SCHEMA_UPDATE_TRUE);
config.setUseLockForDatabaseSchemaUpdate(true);  // required to trigger the bug

ProcessEngine engine = config.buildProcessEngine();

tenantInfoHolder.addTenant("tenant");
config.registerTenant("tenant", hikariFor("my_db"));   // hangs forever
```

## Full reproducible setup

Three files plus a single `docker compose up`. Tested on JDK 25 against SQL Server 2022.

### `docker-compose.yml`

```yaml
services:
  mssql:
    image: mcr.microsoft.com/mssql/server:2022-latest
    environment:
      ACCEPT_EULA: "Y"
      MSSQL_SA_PASSWORD: "yourStrong(!)Password"
    ports: ["1433:1433"]
```

Start it and create the tenant database:

```bash
docker compose up -d
docker exec -i $(docker compose ps -q mssql) \
  /opt/mssql-tools18/bin/sqlcmd -S localhost -U sa \
  -P 'yourStrong(!)Password' -No -Q "
    CREATE DATABASE my_db;
  "
```

### `pom.xml`

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0">
  <modelVersion>4.0.0</modelVersion>
  <groupId>repro</groupId>
  <artifactId>flowable-registertenant-deadlock</artifactId>
  <version>1.0-SNAPSHOT</version>

  <properties>
    <maven.compiler.release>17</maven.compiler.release>
    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
  </properties>

  <dependencies>
    <dependency>
      <groupId>org.flowable</groupId>
      <artifactId>flowable-engine</artifactId>
      <version>8.0.0</version>
    </dependency>
    <dependency>
      <groupId>com.microsoft.sqlserver</groupId>
      <artifactId>mssql-jdbc</artifactId>
      <version>12.8.1.jre11</version>
    </dependency>
    <dependency>
      <groupId>com.zaxxer</groupId>
      <artifactId>HikariCP</artifactId>
      <version>6.0.0</version>
    </dependency>
    <dependency>
      <groupId>org.slf4j</groupId>
      <artifactId>slf4j-simple</artifactId>
      <version>2.0.16</version>
    </dependency>
  </dependencies>

  <build>
    <plugins>
      <plugin>
        <groupId>org.codehaus.mojo</groupId>
        <artifactId>exec-maven-plugin</artifactId>
        <version>3.5.0</version>
        <configuration>
          <mainClass>repro.FlowableRegisterTenantDeadlockRepro</mainClass>
        </configuration>
      </plugin>
    </plugins>
  </build>
</project>
```

### `src/main/java/repro/InMemoryTenantInfoHolder.java`

```java
package repro;

import org.flowable.common.engine.impl.cfg.multitenant.TenantInfoHolder;

import java.util.Collection;
import java.util.LinkedHashSet;
import java.util.Set;

public class InMemoryTenantInfoHolder implements TenantInfoHolder {
    private final Set<String> tenants = new LinkedHashSet<>();
    private final ThreadLocal<String> current = new ThreadLocal<>();

    public void addTenant(String tenantId) { tenants.add(tenantId); }

    @Override public Collection<String> getAllTenants()        { return tenants; }
    @Override public String getCurrentTenantId()               { return current.get(); }
    @Override public void setCurrentTenantId(String tenantId)  { current.set(tenantId); }
    @Override public void clearCurrentTenantId()               { current.remove(); }
}
```

### `src/main/java/repro/FlowableRegisterTenantDeadlockRepro.java`

```java
package repro;

import com.zaxxer.hikari.HikariConfig;
import com.zaxxer.hikari.HikariDataSource;
import org.flowable.engine.ProcessEngine;
import org.flowable.engine.impl.cfg.multitenant.MultiSchemaMultiTenantProcessEngineConfiguration;

import javax.sql.DataSource;
import java.util.concurrent.TimeUnit;

public class FlowableRegisterTenantDeadlockRepro {

    public static void main(String[] args) throws Exception {
        InMemoryTenantInfoHolder tenantInfoHolder = new InMemoryTenantInfoHolder();
        MultiSchemaMultiTenantProcessEngineConfiguration config =
                new MultiSchemaMultiTenantProcessEngineConfiguration(tenantInfoHolder);
        config.setDatabaseType(
                MultiSchemaMultiTenantProcessEngineConfiguration.DATABASE_TYPE_MSSQL);
        config.setDatabaseSchemaUpdate(
                MultiSchemaMultiTenantProcessEngineConfiguration.DB_SCHEMA_UPDATE_TRUE);
        config.setDisableIdmEngine(true);
        config.setDisableEventRegistry(true);
        // comment out this line to make the problem disappear
        config.setUseLockForDatabaseSchemaUpdate(true);

        System.out.println("Booting engine ...");
        ProcessEngine engine = config.buildProcessEngine();
        System.out.println("Engine booted.");

        // Post-boot registration: invokes createTenantSchema(...)
        //   -> SchemaOperationsEngineBuild
        //   -> ProcessDbSchemaManager.schemaUpdate
        //   -> LockManagerImpl.waitForLock  ← deadlocks on MSSQL
        tenantInfoHolder.addTenant("tenant");

        Thread registerThread = new Thread(() -> {
            System.out.println("Registering tenant post-boot ...");
            config.registerTenant("tenant", buildHikari("my_db"));
            System.out.println("tenant registered (this line NEVER prints on MSSQL).");
        }, "register-tenant-0");
        registerThread.start();

        registerThread.join(TimeUnit.SECONDS.toMillis(10));
        if (registerThread.isAlive()) {
            System.err.println("\nDEADLOCK CONFIRMED: tenant registration did not complete in 10s.\n");
            for (StackTraceElement e : registerThread.getStackTrace()) {
                System.err.println("  at " + e);
            }
            System.exit(1);
        }
        engine.close();
    }

    private static DataSource buildHikari(String database) {
        HikariConfig cfg = new HikariConfig();
        cfg.setJdbcUrl(
                "jdbc:sqlserver://localhost:1433;databaseName=" + database
                        + ";encrypt=false;trustServerCertificate=true");
        cfg.setUsername("sa");
        cfg.setPassword("yourStrong(!)Password");
        cfg.setMaximumPoolSize(10);
        cfg.setPoolName("DB Connection Pool " + database);
        return new HikariDataSource(cfg);
    }
}
```

### Running it

```bash
mvn -q compile exec:java
```

Expected output:

```
Booting engine ...
Engine booted.
Registering tenant post-boot ...
DEADLOCK CONFIRMED: tenant registration did not complete in 10s.

  at sun.nio.ch.NioSocketImpl.implRead(NioSocketImpl.java:307)
  ...
  at com.zaxxer.hikari.pool.HikariProxyPreparedStatement.execute(...)
  ...
  at org.flowable.common.engine.impl.cmd.LockCmd.execute(LockCmd.java:71)
  ...
  at org.flowable.common.engine.impl.lock.LockManagerImpl.waitForLock(LockManagerImpl.java:73)
  at org.flowable.engine.impl.db.ProcessDbSchemaManager.schemaUpdate(ProcessDbSchemaManager.java:166)
  ...
  at MultiSchemaMultiTenantProcessEngineConfiguration.registerTenant(...)
```

Commenting out `config.setUseLockForDatabaseSchemaUpdate(true);` makes the problem disappear. Swapping the JDBC URL and `DATABASE_TYPE_MSSQL` for a MariaDB/MySQL instance also completes successfully even with the flag enabled.

## Environment

Flowable 8.0.0 · JDK 25 · HikariCP 6.0.0 · mssql-jdbc 12.8.1 · SQL Server 2022

honma89:
Description

MultiSchemaMultiTenantProcessEngineConfiguration.registerTenant(...) (modules/flowable-engine/src/main/java/org/flowable/engine/impl/cfg/multitenant/MultiSchemaMultiTenantProcessEngineConfiguration.java) hangs indefinitely on databases with transactional DDL (SQL Server, PostgreSQL) when both databaseSchemaUpdate=true and useLockForDatabaseSchemaUpdate=true are set. Not reproducible on MySQL/MariaDB, since DDL there implicitly commits.

Root Cause

The schema build (SchemaOperationsEngineBuild) runs as a single top-level command on one connection (C_outer) for its entire duration. CommonDbSchemaManager.schemaUpdate() runs first, unlocked, and issues DDL against ACT_GE_PROPERTY (e.g. CREATE TABLE on a fresh schema). On SQL Server/PostgreSQL, that DDL's schema-modification lock is held until C_outer's transaction commits.

ProcessDbSchemaManager.schemaUpdate() then acquires the schema lock via LockManagerImpl.waitForLock(), which is hard-coded to new CommandConfig(false, TransactionPropagation.REQUIRES_NEW) — forcing a brand-new connection (C_inner), since contextReusePossible=false alone triggers this in CommandContextInterceptor regardless of whether a transaction interceptor is configured. The lock command's SELECT against ACT_GE_PROPERTY on C_inner then blocks on the schema lock still held by the uncommitted DDL on C_outer. C_outer can't commit until the outer command returns, which can't happen until that inner SELECT completes — a self-deadlock invisible to the database's own deadlock detector, since C_outer is idle rather than genuinely blocked.

Confirmed via sys.dm_exec_requests on SQL Server: the blocking session is idle, holding LCK_M_SCH_S on ACT_GE_PROPERTY, and both sessions originate from the same application host/thread.

Fix

Added an opt-in reuseCurrentCommandContext mode to LockManagerImpl, exposed via AbstractEngineConfiguration#getLockManager(String, boolean) and threaded through SchemaManagerLockConfiguration / EngineSchemaManagerLockConfiguration. When enabled, the lock is acquired and released on the same connection/transaction as the surrounding schema build instead of a newly opened one, so it sees the outer transaction's own uncommitted DDL rather than blocking on it. Other nodes/processes attempting to acquire the same lock on their own connections are still correctly blocked until the schema-build transaction commits, so cross-process mutual exclusion is preserved.

Applied to ProcessDbSchemaManager (the class in the reported stack trace) and, for consistency, to EngineSqlScriptBasedDbSchemaManager, which is shared by the cmmn/dmn/app/event-registry/idm engines and has the identical structural exposure. Existing callers of getLockManager(String) are unaffected, since the new parameter defaults to false.

fixes #4215

### DIFF
--- a/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/AbstractEngineConfiguration.java
+++ b/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/AbstractEngineConfiguration.java
@@ -1121,7 +1121,16 @@ public abstract class AbstractEngineConfiguration {
     }
 
     public LockManager getLockManager(String lockName) {
-        return new LockManagerImpl(commandExecutor, lockName, getLockPollRate(), getEngineCfgKey());
+        return getLockManager(lockName, false);
+    }
+
+    /**
+     * @param reuseCurrentCommandContext see {@link LockManagerImpl#LockManagerImpl(CommandExecutor, String, Duration, Duration, String, boolean)}.
+     *         Pass {@code true} when the lock is (or may be) acquired from within a command that already holds uncommitted DDL
+     *         on the current connection, e.g. schema creation/update, to avoid a self-deadlock on databases with transactional DDL.
+     */
+    public LockManager getLockManager(String lockName, boolean reuseCurrentCommandContext) {
+        return new LockManagerImpl(commandExecutor, lockName, getLockPollRate(), null, getEngineCfgKey(), reuseCurrentCommandContext);
     }
 
     // getters and setters
--- a/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/db/EngineSchemaManagerLockConfiguration.java
+++ b/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/db/EngineSchemaManagerLockConfiguration.java
@@ -44,6 +44,11 @@ public class EngineSchemaManagerLockConfiguration implements SchemaManagerLockCo
     }
 
     @Override
+    public LockManager getLockManager(String lockName, boolean reuseCurrentCommandContext) {
+        return getEngineConfiguration().getLockManager(lockName, reuseCurrentCommandContext);
+    }
+
+    @Override
     public Duration getSchemaLockWaitTime() {
         return getEngineConfiguration().getSchemaLockWaitTime();
     }
--- a/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/db/EngineSqlScriptBasedDbSchemaManager.java
+++ b/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/db/EngineSqlScriptBasedDbSchemaManager.java
@@ -82,7 +82,11 @@ public abstract class EngineSqlScriptBasedDbSchemaManager extends AbstractSqlScr
     public void schemaCreate() {
 
         if (lockConfiguration.isUseLockForDatabaseSchemaUpdate()) {
-            LockManager lockManager = lockConfiguration.getLockManager(getDbSchemaLockName());
+            // Reuse the current command context/connection: schemaCreateInLock() may run DDL on the same connection.
+            // Acquiring the lock on a separate, newly opened connection/transaction would self-deadlock on databases
+            // with transactional DDL (e.g. SQL Server, PostgreSQL) once an earlier schema manager in this same build
+            // has already executed uncommitted DDL on this connection.
+            LockManager lockManager = lockConfiguration.getLockManager(getDbSchemaLockName(), true);
             lockManager.waitForLockRunAndRelease(lockConfiguration.getSchemaLockWaitTime(), () -> {
                 schemaCreateInLock();
                 return null;
@@ -123,7 +127,8 @@ public abstract class EngineSqlScriptBasedDbSchemaManager extends AbstractSqlScr
     @Override
     public String schemaUpdate() {
         if (lockConfiguration.isUseLockForDatabaseSchemaUpdate()) {
-            LockManager lockManager = lockConfiguration.getLockManager(getDbSchemaLockName());
+            // See the comment in schemaCreate() above for why the current command context/connection is reused here.
+            LockManager lockManager = lockConfiguration.getLockManager(getDbSchemaLockName(), true);
             return lockManager.waitForLockRunAndRelease(lockConfiguration.getSchemaLockWaitTime(), this::schemaUpdateInLock);
         } else {
             return schemaUpdateInLock();
--- a/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/db/SchemaManagerLockConfiguration.java
+++ b/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/db/SchemaManagerLockConfiguration.java
@@ -25,5 +25,15 @@ public interface SchemaManagerLockConfiguration {
 
     LockManager getLockManager(String lockName);
 
+    /**
+     * @param reuseCurrentCommandContext when {@code true}, the returned lock manager acquires/releases the lock using the
+     *         currently active command context (same connection/transaction), instead of a brand new one. This must be used
+     *         when locking around schema creation/update, to avoid a self-deadlock on databases with transactional DDL
+     *         (SQL Server, PostgreSQL, ...) - see {@code LockManagerImpl} for details.
+     */
+    default LockManager getLockManager(String lockName, boolean reuseCurrentCommandContext) {
+        return getLockManager(lockName);
+    }
+
     Duration getSchemaLockWaitTime();
 }
--- a/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/lock/LockManagerImpl.java
+++ b/modules/flowable-engine-common/src/main/java/org/flowable/common/engine/impl/lock/LockManagerImpl.java
@@ -51,6 +51,24 @@ public class LockManagerImpl implements LockManager {
     }
 
     public LockManagerImpl(CommandExecutor commandExecutor, String lockName, Duration lockPollRate, Duration forceAcquireAfter, String engineType) {
+        this(commandExecutor, lockName, lockPollRate, forceAcquireAfter, engineType, false);
+    }
+
+    /**
+     * @param reuseCurrentCommandContext when {@code true}, lock commands are executed on the currently active
+     *         {@link org.flowable.common.engine.impl.interceptor.CommandContext} (i.e. the same database connection/transaction),
+     *         instead of the default behaviour of always running them in a brand new, independently committed transaction.
+     *         <p>
+     *         This must be used whenever the lock is acquired from a command that may already hold uncommitted DDL on the
+     *         same connection (typically during schema creation/update). On databases with transactional DDL (SQL Server,
+     *         PostgreSQL, ...) a schema-modification lock taken by such DDL is held until the enclosing transaction commits.
+     *         If the lock command were to run on a <em>different</em>, newly opened connection/transaction (the default),
+     *         it would block forever waiting for that still-open outer transaction to release the schema lock - which it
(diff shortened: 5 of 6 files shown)


nikhiln64 (changes requested):
## Context

Thanks for chasing the self deadlock, moving the schema lock onto the current command context does remove the reentrant acquire. My worry is that `reuseCurrentCommandContext=true` changes more than the deadlock. With CommandConfig(true, REQUIRED) the acquire and release run inside the outer schema build's context, and a reused context is never closed, so flushSessions only runs when the whole build commits (`CommandContextInterceptor` at 80, 89 to 94, and 126 to 128). So the lock property write is buffered and not sent to the database until the build commits, and every poll iteration reads the lock property cache first from the shared EntityCache (`AbstractDataManager.findById` at 70 to 73). During the build there is no committed lock row for another node to see or block on.

▎ Example: Two nodes run schemaUpdate at startup against a schema that already exists, on an upgrade that does not alter ACT_GE_PROPERTY so no table level DDL lock serializes them. Each reads no committed lock value, acquires by buffering a write that only flushes at the end, and both run the upgrade DDL at the same time, which is what useLockForDatabaseSchemaUpdate is meant to prevent. On PostgreSQL the loser's insert also hits the unique constraint at the outer flush and aborts the whole schema build rather than rolling back harmlessly.

## Suggestion

The description says other nodes are still blocked until the schema build commits, but I could not find that in the code once the write is deferred. A concurrent test on SQL Server and PostgreSQL that runs two engine configs against the same schema, especially the upgrade with tables already present, asserting only one node runs the guarded DDL, would settle it. If it fails, the fix likely needs an uncached committed read of the lock property, or to scope reuse mode to the fresh create path where the create table lock already serializes nodes.

## NOTE

Happy to be wrong if there is a DB level guarantee I am missing on the upgrade path.

PR #4258, opened 2026-08-12
Record a delete reason when an error event sub-process interrupts. Fixes #4257

### LINKED ISSUE #4257: Interrupting error event sub-process leaves a null delete reason on the activities it terminates
# Interrupting error event sub-process leaves a null delete reason on the activities it terminates

**Version:** 8.0.0, and unchanged on `main` as of 2026-08-12.
**Database Vendor:** PostgreSQL
**SpringBoot:** 4.1.0

## Summary

When an interrupting **error** event sub-process catches a `BpmnError`, the executions it terminates are deleted with a `null` delete reason. `ACT_HI_ACTINST.DELETE_REASON_` is therefore null on every activity it killed, making a terminated activity indistinguishable in history from one that completed normally.

Every *other* interrupting event sub-process start type records a reason.

## Why this looks unintended

`org.flowable.engine.history.DeleteReason` declares `EVENT_SUBPROCESS_INTERRUPTING = "event subprocess"`, and five sibling behaviors use it:

- `EventSubProcessTimerStartEventActivityBehavior`
- `EventSubProcessMessageStartEventActivityBehavior`
- `EventSubProcessSignalStartEventActivityBehavior`
- `EventSubProcessEventRegistryStartEventActivityBehavior`
- `EventSubProcessVariableListenerlStartEventActivityBehavior`

each as `DeleteReason.EVENT_SUBPROCESS_INTERRUPTING + "(" + startEvent.getId() + ")"`.

`BoundaryEventActivityBehavior` likewise records `DeleteReason.BOUNDARY_EVENT_INTERRUPTING`.

The error path is the only one that does not. In `ErrorPropagation.executeEventHandler`:

```java
if (parentExecution.isProcessInstanceType()) {
    executionEntityManager.deleteChildExecutions(parentExecution, null, true);
```

`ErrorPropagation` never references `DeleteReason` at all.

This also contradicts the documented contract of the field. The javadoc for `HistoricActivityInstance#getDeleteReason()` reads:

> Returns the delete reason for this activity, if any was set (**if completed normally, no delete reason is set**)

An activity terminated by this path satisfies that description in representation while being its opposite in fact.

## Reproduction

[This gist](https://gist.github.com/chaserb/da044ff55eb52f91119f73d9993bc832) contains a test in the style of the existing `DeleteReasonTest`, which has `testInterruptingBoundaryEvent` but no event-sub-process equivalent. `ErrorEventSubProcessDeleteReasonTest` contains:

- `testInterruptingErrorEventSubProcess` — **fails**. A forked process parks one branch on a user task; the other raises a `BpmnError` from a start execution listener. The interrupting error event sub-process terminates the scope, and the parked branch's activity ends with `getDeleteReason() == null`.
- `testInterruptingBoundaryEventForComparison` — **passes**, showing the boundary-event path records its reason.

## Impact

Any consumer building an audit or monitoring view from `ACT_HI_ACTINST` reports these activities as completed. In our case a user task that was destroyed by the interruption showed as "Complete" to an approver-facing audit surface, with no signal anywhere in history that it had been terminated.

`ACTIVITY_CANCELLED` is dispatched and is a viable hook, but it requires listening and persisting alongside history rather than reading history, which is a very different integration for what the delete-reason column already expresses everywhere else.

## Suggested fix

Pass the reason the sibling behaviors pass. `fix.patch` is attached; both branches of `executeEventHandler` that delete executions currently pass `null` as the `deleteReason` argument:

```java
executionEntityManager.deleteChildExecutions(parentExecution,
        DeleteReason.EVENT_SUBPROCESS_INTERRUPTING + "(" + event.getId() + ")", true);
...
executionEntityManager.deleteExecutionAndRelatedData(currentExecution,
        DeleteReason.EVENT_SUBPROCESS_INTERRUPTING + "(" + event.getId() + ")", false);
```

I have deliberately left `cancelActivity` alone to keep the change minimal, even though the sibling behaviors pass the interrupting start event there and this path passes `null`. Happy to include that too if you would prefer the paths to match exactly.

I could not find an existing report for this on the issue tracker or the forum, but I may well have missed one.

chaserb:
Fixes [4257](https://github.com/flowable/flowable-engine/issues/4257)

`ErrorPropagation.executeEventHandler` deletes the executions an interrupting error event sub-process terminates without passing a delete reason, so `ACT_HI_ACTINST.DELETE_REASON_` is left null and a terminated activity is indistinguishable in history from one that completed normally — which contradicts `HistoricActivityInstance#getDeleteReason()`, whose javadoc states that no reason is set precisely when an activity completes normally. Every other interrupting event sub-process start type records `DeleteReason.EVENT_SUBPROCESS_INTERRUPTING`, and boundary events record `BOUNDARY_EVENT_INTERRUPTING`; the error start is the only one that does not. This passes that same reason at both call sites, and adds the `DeleteReasonTest` case that was missing for this start type.

#### Check List:
* Unit tests: YES
* Documentation: Causes behavior to conform to [existing documentation](https://www.flowable.com/open-source/docs/all-javadocs/org/flowable/engine/history/HistoricActivityInstance.html#:~:text=%C2%A0getDeleteReason()-,Returns%20the%20delete%20reason%20for%20this%20activity%2C%20if%20any%20was%20set%20(if%20completed%20normally%2C%20no%20delete%20reason%20is%20set),-getTenantId)

### DIFF
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/deletereason/ThrowBpmnErrorListener.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/deletereason/ThrowBpmnErrorListener.java
@@ -1,0 +1,26 @@
+package org.flowable.engine.test.api.deletereason;
+
+import org.flowable.engine.delegate.BpmnError;
+import org.flowable.engine.delegate.DelegateExecution;
+import org.flowable.engine.delegate.ExecutionListener;
+import org.flowable.engine.delegate.JavaDelegate;
+
+/** Raises a BpmnError from a start execution listener. */
+public class ThrowBpmnErrorListener implements ExecutionListener {
+
+    @Override
+    public void notify(DelegateExecution execution) {
+        throw new BpmnError("someError", "raised from a start execution listener");
+    }
+
+}
+
+/** Never reached; the listener throws before the behaviour runs. */
+class NoopDelegate implements JavaDelegate {
+
+    @Override
+    public void execute(DelegateExecution execution) {
+        // no-op
+    }
+
+}
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/deletereason/ThrowBpmnErrorListener.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/deletereason/ThrowBpmnErrorListener.java
@@ -1,3 +1,15 @@
+/* Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.flowable.engine.test.api.deletereason;
 
 import org.flowable.engine.delegate.BpmnError;
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/deletereason/ThrowBpmnErrorListener.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/deletereason/ThrowBpmnErrorListener.java
@@ -26,13 +26,3 @@ public class ThrowBpmnErrorListener implements ExecutionListener {
     }
 
 }
-
-/** Never reached; the listener throws before the behaviour runs. */
-class NoopDelegate implements JavaDelegate {
-
-    @Override
-    public void execute(DelegateExecution execution) {
-        // no-op
-    }
-
-}
--- a/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/deletereason/ThrowBpmnErrorListener.java
+++ b/modules/flowable-engine/src/test/java/org/flowable/engine/test/api/deletereason/ThrowBpmnErrorListener.java
@@ -15,7 +15,6 @@ package org.flowable.engine.test.api.deletereason;
 import org.flowable.engine.delegate.BpmnError;
 import org.flowable.engine.delegate.DelegateExecution;
 import org.flowable.engine.delegate.ExecutionListener;
-import org.flowable.engine.delegate.JavaDelegate;
 
 /** Raises a BpmnError from a start execution listener. */
 public class ThrowBpmnErrorListener implements ExecutionListener {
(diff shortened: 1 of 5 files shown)


modules/flowable-engine/src/test/java/org/flowable/engine/test/api/deletereason/ThrowBpmnErrorListener.java
```diff
@@ -0,0 +1,38 @@
...
+
+/** Never reached; the listener throws before the behaviour runs. */
+class NoopDelegate implements JavaDelegate {
+
+    @Override
+    public void execute(DelegateExecution execution) {
+        // no-op
+    }
+
+}
```
filiphr: This shouldn't be needed. See my comment int he BPMN XMLs

modules/flowable-engine/src/test/resources/org/flowable/engine/test/api/deletereason/ErrorEventSubProcessDeleteReasonTest.bpmn20.xml
```diff
@@ -0,0 +1,49 @@
...
+    <!-- Branch A: a call activity whose child parks on a user task. -->
+    <sequenceFlow id="toWaiting" sourceRef="fork" targetRef="waitingCall"/>
+    <callActivity id="waitingCall" name="Waiting Call" calledElement="waitingChild"/>
+    <sequenceFlow id="waitingToEnd" sourceRef="waitingCall" targetRef="endA"/>
+    <endEvent id="endA"/>
+
+    <!-- Branch B: a start execution listener raises a BpmnError. -->
+    <sequenceFlow id="toThrowing" sourceRef="fork" targetRef="throwingTask"/>
+    <serviceTask id="throwingTask" name="Throwing Task"
+                 flowable:class="org.flowable.engine.test.api.deletereason.NoopDelegate">
```
filiphr:
```suggestion
                 flowable:expression="${true}">
```

filiphr (changes requested):
Thanks @chaserb. Fix looks OK to me. I only have 2 comments for the testing so that we can simplify it a bit.

modules/flowable-engine/src/test/java/org/flowable/engine/test/api/deletereason/ThrowBpmnErrorListener.java
chaserb: Thanks @filiphr , removed.

modules/flowable-engine/src/test/resources/org/flowable/engine/test/api/deletereason/ErrorEventSubProcessDeleteReasonTest.bpmn20.xml
chaserb: Changed to a 'true' literal as suggested.

filiphr:
Thanks for the adjustments @chaserb

PR #4265, opened 2026-08-24
Add inheritVariables property to CMMN CaseTask and ProcessTask

BertMatthys:


modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/behavior/impl/ChildTaskActivityBehavior.java
```diff
@@ -83,6 +90,12 @@ protected void handleInParameters(PlanItemInstanceEntity planItemInstanceEntity,
                                       CmmnEngineConfiguration cmmnEngineConfiguration, Map<String, Object> inParametersMap,
                                       ExpressionManager expressionManager) {
 
+        if (inheritVariables) {
+            // Inherit all variables from the parent. Explicit in parameters are
+            // applied afterwards, so they take precedence over the inherited variables.
+            inParametersMap.putAll(planItemInstanceEntity.getVariables());
+        }
```
filiphr: We shouldn't do this like this. We should do something similar like we do in `IOParameterUtil.processInParameters` where we do

```java
if (value != null) {
    value = JsonUtil.deepCopyIfJson(value);
}
```

We should also look at not copying read only variable instances such as the `CmmnAggregatedVariableType`, `BpmnAggregatedVariableType`, `ParallelMultiInstanceLoopVariableType`

modules/flowable-cmmn-engine/src/main/java/org/flowable/cmmn/engine/impl/behavior/impl/ChildTaskActivityBehavior.java
```diff
@@ -83,6 +90,12 @@ protected void handleInParameters(PlanItemInstanceEntity planItemInstanceEntity,
                                       CmmnEngineConfiguration cmmnEngineConfiguration, Map<String, Object> inParametersMap,
                                       ExpressionManager expressionManager) {
 
+        if (inheritVariables) {
+            // Inherit all variables from the parent. Explicit in parameters are
+            // applied afterwards, so they take precedence over the inherited variables.
+            inParametersMap.putAll(planItemInstanceEntity.getVariables());
```
filiphr: Can we do this in a similar way we do it for BPMN where we distinguish between transient and non transient variables. With this we now also include transient variables in the ones that would be persisted.

filiphr (changes requested):
@BertMatthys should we also do this for the BPMN `CaseTaskActivityBehavior`? See also my inline comment

